PR #3, opened 2021-01-26
feat: Responsive Transitions, Modals, & Actionbars

pretzelhammer:
implements responsive transitions, modals, and actionbars so that they look differently for mobile and desktop resolutions

look at the updated docs: https://square.github.io/maker/styleguide/desktop-modals/#/Modal

**Other information:**
currently a draft, still need some feedback from design & eng


src/components/Modal/src/ModalLayer.vue
```diff
@@ -114,6 +128,36 @@ export default {
...
+	width: 100%;
+	height: 100%;
+	overflow: scroll;
+	box-shadow: 0 0 24px 8px rgba(0, 0, 0, 0.3);
+}
+
+@media screen and (min-width: 1200px) {
+	.ModalContainer {
+		display: inline-block;
+		width: auto;
```
landondurnan, line 155: We're likely going to need to set a specific width / min-width here. With fluid content everything can be real narrow.  Was testing and you could see the width change with after the image rendered.

src/components/Modal/src/ModalLayer.vue
```diff
@@ -114,6 +128,36 @@ export default {
...
+		max-width: 1000px;
+		height: auto;
```
landondurnan, line 157: Not sure yet how we're going to handle modal stacking and the height potentially changing, but I'm guessing CSS won't be enough and we may need to pass events about the height changing to keep "parent" modals in sync.

src/components/Modal/src/ModalLayer.vue
```diff
@@ -1,24 +1,37 @@
...
-		<div
-			v-if="modalApi.state.vnode"
-			:class="$s.ModalLayer"
-		>
-			<pseudo-window
-				body
-				:class="$s.disableScroll"
+	<div>
+		<m-transition-fade>
+			<div
```
landondurnan, line 4: I'm thinking we should start to get an overlay component wired up to accept click/tap events to close anything inside of it.

landondurnan (commented):
Since we're still waiting on some design decisions here, might be worth waiting to make too many changes. But there's a couple pieces we could start to consider sooner, noted in a few ideas.

src/components/Modal/src/ModalLayer.vue
pretzelhammer: This is one of those things where I'm not sure whether or not the library should have an opinion about. I think that modals which look "thin" (too narrow vertically or too narrow horizontally) don't look good, and also modals which dynamically resize after being opened also don't look good, and while adding a `min-width` and `min-height` to `.ModalContainer` can help _mitigate_ (not solve) these problems, they're trivially solvable by the end user as long as they simply pass in their own container with a fixed width & height or fixed minimum width & height, e.g.

```vue
<template>
    <m-modal>
        <div style="width: <perfect width for my modal>; height: <perfect height for my modal>;" />
    </m-modal>
</template>
```

Having the end user specify the ideal width & height for their modal seems like a more reliable way to consistently get good looking modals than arbitrarily enforcing some minimum width and height.

And yet just as I finished writing this I realized the above solution won't work for the mobile version of the modal. Hmmm, so either the end user has to be familiar with the breakpoints we use to switch from the mobile version to the desktop version or we add props to the modal like `desktop-width` and `desktop-height`. Actually, the props aren't that great of an idea either, as it would require us to watch the screen using JS. Well, maybe there's a pure CSS solution using custom properties, I'd have to experiment.

src/components/Modal/src/ModalLayer.vue
pretzelhammer: Why do we need to keep them in sync? Also would this not be the responsibility of the end user? They know the size requirements of their modals better than we do.

src/components/Modal/src/ModalLayer.vue
pretzelhammer: This is something else I debated about adding. If we add a listener for overlay / off-modal clicks that behavior would **only** be available on desktops and to desktop users, and since most devs develop on desktops they might add something important to the function of the modal to the overlay click handler without realizing they're leaving out all mobile users. Then again, this may be just another thing we leave as the responsibility of the dev.

src/components/Modal/src/ModalLayer.vue
landondurnan: I hadn't considered that level of flexibility on the width / height, but you make some valid points.  Then it's just the challenges between mobile / desktop. How much CSS styling is too much? (I'll look around at some other super flexible systems for some ideas)

src/components/Modal/src/ModalLayer.vue
landondurnan: I've been assuming based on our conversations with design that we'd want to constrain the proportions of the parent and child so that the child is contained inside the parent.  However, if a child needs to be bigger / smaller than the parent we'd communicate to the parent its size so we could transition smoothly when the child is closed to reset the parent back to it's original size.

Maybe that's too much to assume. Maybe that's too much to manage. But if that's the desired behavior we'd have to either do it internally or provide hooks / documentation for the dev to do it themselves.

src/components/Modal/src/ModalLayer.vue
landondurnan: Is the debate if the overlay component should be internal, external (so dev can choose to use it) or exist at all?  It seems like we should provide something, but worth discussing how.

src/components/Modal/README.md
```diff
@@ -5,102 +5,305 @@ Use the modal component to enter the user into a new _mode_.
...
 </script>
 ```
 
-_Demo.vue_
+_DemoModal.vue_
 
 ```vue
+<template>
+	<m-modal>
+		<img
```
landondurnan, line 49: There's a couple concerns with leveraging this placeholder image.

1. We might want to use our image component here in the demo so there's a nice loading state as the image is downloading
2. This behavior of the width of the modal being skinny and then popping to be wider when the image loads isn't ideal. This will happen on production unless we can either recommend how to style the modal to support the output in a way that it doesn't happen.

A CSS only solution that would get us closer to the current desired output would look like this:
```
.cover-photo {
	width: 100%;
	max-height: 30vh;
	aspect-ratio: 18 / 9;
	object-fit: cover;
	object-position: center;
}
```

Aspect ratio isn't fully supported yet, so that's why it might be nice to build it into the Image component with fallbacks -- that doesn't need to be done in this PR, but try the code out -- it provides a better experience.

src/components/Modal/src/Modal.vue
```diff
@@ -86,54 +92,35 @@ const Modal = {
...
+@media screen and (min-width: 1200px) {
+	.Modal {
+		display: inline-block;
+		width: auto;
+		min-width: 600px;
+		max-width: 50vw;
+		height: auto;
+		max-height: calc(100vh - 48px);
+		overflow: scroll;
+		border-radius: 16px;
```
landondurnan, line 122: This should be 8px

src/components/Modal/src/Modal.vue
```diff
@@ -86,54 +92,35 @@ const Modal = {
...
 	background: #f5f6f7;
-	box-shadow: 0 0 24px 8px rgba(0, 0, 0, 0.3);
+}
+
+@media screen and (min-width: 1200px) {
+	.Modal {
+		display: inline-block;
+		width: auto;
+		min-width: 600px;
+		max-width: 50vw;
```
landondurnan, line 118: I'm not sure if we need both min-width: 600 and max-width: 50vw.  50vw ends up being between 600-700px so there's some scaling that happens, but around 100px. This tiny bit of scaling doesn't seem to help anything and is more likely to cause some slight hiccups with rendering content at arbitrary desktop sizes.  If we just use max-width to a set px value at desktop things are more predictable.

src/components/Modal/src/Modal.vue
```diff
@@ -86,54 +92,35 @@ const Modal = {
...
 .Modal {
+	width: 100%;
 	height: 100%;
-	overflow-y: auto;
 	background: #f5f6f7;
-	box-shadow: 0 0 24px 8px rgba(0, 0, 0, 0.3);
+}
+
+@media screen and (min-width: 1200px) {
+	.Modal {
```
landondurnan, line 114: The modal is supposed to have 24px of padding. Right now if you try to add this padding manually to the modal, nothing visible happens on desktop. On mobile the entire modal is inset by 24px. Both aren't expected.

Do we need a ModalContent component or are we to instruct manually adding this padding selectively? The latter seems risky.

src/components/Modal/README.md
```diff
@@ -5,102 +5,305 @@ Use the modal component to enter the user into a new _mode_.
...
+	},
+};
+</script>
+```
+
+_StackedDemoFirstModal.vue_
+
+```vue
+<template>
+	<m-modal disable-pan-down-close>
```
landondurnan, line 257: I thought we were going to disable the behavior by default and enable through a prop? Does this need more discussion / alignment before doing so?

landondurnan (commented):
Some things we need to work out, added some comments / suggestions / questions.

src/components/Modal/src/ModalLayer.vue
landondurnan: Can we add a click to close behavior?  Some props I saw somewhere else:


closeOnEsc | boolean | If `true`, the modal will close when the `Esc` key is pressed | true
closeOnOverlayClick | boolean | If `true`, the modal will close when the overlay is clicked | true

landondurnan (commented):
Left a bunch of comments last night, but not sure if they went through.

src/components/Modal/src/Modal.vue
pretzelhammer: Shimaa told me it should only be 600px. I've updated it.

src/components/Modal/README.md
```diff
@@ -142,186 +365,496 @@ export default {
...
-```html
+```vue
 <template>
 	<m-modal>
-		Modal content
-
-		<button @click="modalApi.close()">
-			Close modal
-		</button>
+		<m-transition-resize>
```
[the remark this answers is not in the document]
pretzelhammer: Not possible.

src/components/Modal/src/ModalLayer.vue
pretzelhammer: solved with TransitionResize component

src/components/Modal/src/ModalLayer.vue
pretzelhammer: Shimaa said width is fixed to 600px and height is dynamic, and switching between different height modals is solved by TransitionResize component

src/components/Modal/src/Modal.vue
pretzelhammer: added ModalContent component

src/components/Modal/README.md
pretzelhammer: at this point it'd be better off as a docs example, we can do it in a follow-up PR

src/components/Modal/src/ModalLayer.vue
pretzelhammer: not clear if:
- these would go on ModalLayer or on Modal or both (not sure that last one even makes sense?)
- these should be multiple props or a single props that takes multiple arguments via an array
- should be default enabled and disabled by props or default disabled and enabled by props
- should be cancellable via callback

The good news: we don't have to answer any of this questions in this PR, regardless what the answer is to any of the above questions we can add it later in a follow up PR without introducing any breaking changes.

src/components/ActionBar/README.md
```diff
@@ -1,5 +1,174 @@
...
+
+<style scoped>
+.icon {
+	width: 24px;
+	height: 24px;
+}
+</style>
+```
+
+<!--
```
landondurnan, line 107: Is this needed?

src/components/ActionBar/README.md
```diff
@@ -1,5 +1,174 @@
 # ActionBar
```
landondurnan, line 1: The biggest nuance of ActionBar is how it works between views on mobile. I think we should document putting the action bar layer at the root of the app and demonstrating the transitions between action bars.  Perhaps it's sufficient for now to link to the modal demo or lab?  But others should be able to construct the same code looking at this documentation / demo.

src/components/ActionBar/README.md
```diff
@@ -8,8 +177,44 @@
...
+
+Supports attributes from [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).
+
+| Prop       | Type      | Default     | Possible values                    | Description                           |
+| ---------- | --------- | ----------- | ---------------------------------- | ------------------------------------- |
+| type       | `string`  | `'button'`  | —                                  | Type of the button                    |
+| full-width | `boolean` | `false`     | —                                  | Whether to make the button full-width |
+| color      | `string`  | `'#000'`    | —                                  | Background color of button            |
+| text-color | `string`  | —           | —                                  | Text color of button                  |
+| variant    | `string`  | `'primary'` | `primary`, `secondary`, `tertiary` | Semantic variant                      |
```
landondurnan, line 190: I think we need to reduce this to just primary / secondary.  There's only two states in actionbar buttons.

src/components/Modal/README.md
```diff
@@ -258,70 +212,425 @@ export default {
...
+				Second view content
+			</m-text>
+		</m-modal-content>
+
+		<m-responsive-action-bar>
+			<m-action-bar-button
+				key="close"
+				shape="pill"
+				color="#f6f6f6"
+				@click="modalApi.close()"
```
landondurnan, line 552: This ends up closing the second view and first view. This should only close the current view, in this case the SecondView and reveal the previous view, FirstView

landondurnan (changes requested):
Going to keep testing and reviewing, but had enough to submit some feedback.

src/components/ActionBar/README.md
pretzelhammer: removed

src/components/ActionBar/README.md
pretzelhammer: added **keyed transitions** demo to https://square.github.io/maker/styleguide/desktop-modals/#/ActionBar which shows the transitions between a root actionbar and modal actionbar

src/components/ActionBar/README.md
pretzelhammer: removed

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -0,0 +1,341 @@
...
+			<!-- @slot Information label -->
+			<slot
+				name="information"
+			/>
+		</span>
+	</button>
+</template>
+
+<script>
+import chroma from 'chroma-js';
```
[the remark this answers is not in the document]
pretzelhammer: added TODO comment

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -0,0 +1,341 @@
...
+		 */
+		textColor: {
+			type: String,
+			default: undefined,
+			validator: (color) => chroma.valid(color),
+		},
+		/**
+		 * Semantic variant
+		 */
+		variant: {
```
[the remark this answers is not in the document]
pretzelhammer: Do you want me to remove the variants prop entirely from actionbarbuttons?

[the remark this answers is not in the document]
landondurnan: I think a primary and secondary make sense, assuming the primary is the brand filled and the secondary is white with the brand color for the text color.
CleanShot 2021-02-22 at 10 29 16

src/components/Modal/README.md
pretzelhammer: This is not a bug, this works as intended, but since it sounds like you want it to behave more like a "flow" I've updated the example.

src/components/ActionBar/src/ActionBarButton.vue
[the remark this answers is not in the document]
pretzelhammer: I don't think we should be making design decisions in PRs. I've removed the variants, nobody is using them anyway and all the designs you show in your screenshot are already easy to achieve with the existing `color` and `text-color` props on ActionBarButtons.

landondurnan (approved):
Didn't test in every browser, but the primary demos and documentation are in a good enough state to share. (Discussed doing a pre-release to get additional feedback in our current use cases)

PR #15, opened 2021-02-25
fix(action-bar): vnode listeners patch

privatenumber:
When given two vNodes with the same key that are swapped for each other, the second vNode's listener extends the previous listener's via `.fns` property. This didn't handle a vNode reverse scenario, which broke when swapping the Action Bar vNodes back to the previous page's.

- Added a custom patch that backs up and restores the vNode's original listeners
- Enhanced "ActionBarTest" lab to have multiple pages where each page can open a modal to demonstrate that this solution works.

**Other information:**

### DIFF
--- a/lab/experiments/ActionBarTest/CartModal.vue
+++ b/lab/experiments/ActionBarTest/CartModal.vue
@@ -60,7 +60,10 @@ export default {
 	},
 
 	props: {
-		openedBy: String,
+		openedBy: {
+			type: String,
+			required: true,
+		},
 	},
 };
 </script>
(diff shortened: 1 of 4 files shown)


pretzelhammer:
Thanks for doing this. This is awesome. However, like you, I'm worried about the maintainability of this solution. I pushed back against Design and they agreed to compromise and loosen some requirements so I took a different approach in this PR https://github.com/square/maker/pull/16 where I refactor an InlineActionBar component out of ActionBar to render inline so it can be placed inside Modals and Blades. I'm going to close this for now but we may have to revisit it in the future.

PR #16, opened 2021-02-25
feat: Restacking modals

pretzelhammer:
engs still want to stack modals using the `modalApi`, this PR adds modal stacking back in after previously being removed

scroll down to **Stacking Modals** example: https://square.github.io/maker/styleguide/restacking-modals/#/Modal

**Other information:**
Migration steps from Maker 1.1.x to Maker 1.2.x:
1. Move `m-modal-layer` outside of `m-action-bar-layer` at the root of your App, e.g. update this:
```html
<template>
    <m-action-bar-layer>
        <!-- your app -->
        <m-modal-layer />
    </m-action-bar-layer>
</template>
```
To this:
```html
<template>
    <div>
        <m-action-bar-layer>
            <!-- your app -->
        </m-action-bar-layer>
        <m-modal-layer />
    </div>
</template>
```
2. Change `m-action-bar` to `m-inline-action-bar` inside all modals, e.g. update this:
```html
<template>
    <m-modal>
        <m-action-bar>
            <!-- buttons & stuff -->
        </m-action-bar>
    </m-modal>
</template>
```
To this:
```html
<template>
    <m-modal>
        <m-inline-action-bar>
            <!-- buttons & stuff -->
        </m-inline-action-bar>
    </m-modal>
</template>
```
3. Change `m-button` inside of `m-action-bar` and `m-inline-action-bar` to `m-action-bar-button`, e.g. update this:
```html
<template>
    <m-action-bar>
        <m-button>
            <!-- content -->
        </m-button>
    </m-action-bar>
    <m-inline-action-bar>
        <m-button>
            <!-- content -->
        </m-button>
    </m-inline-action-bar>
</template>
```
To this:
```html
<template>
    <m-action-bar>
        <m-action-bar-button>
            <!-- content -->
        </m-action-bar-button>
    </m-action-bar>
    <m-inline-action-bar>
        <m-action-bar-button>
            <!-- content -->
        </m-action-bar-button>
    </m-inline-action-bar>
</template>
```


src/components/Modal/src/ModalLayer.vue
```diff
@@ -33,16 +40,27 @@ import {
...
+	inject: {
+		currentLayer: {
+			default: undefined,
+			from: modalApi,
+		},
+	},
+
 	provide() {
 		const vm = this;
+		// const id = Math.random().toString().slice(2, 5);
```
landondurnan, line 52: Want to clean out these inline comments?

src/components/ActionBar/index.js
```diff
@@ -2,3 +2,4 @@ export { default as MActionBar } from './src/ActionBar.vue';
 export { default as MActionBarLayer } from './src/ActionBarLayer.vue';
 export { default as MActionBarButton } from './src/ActionBarButton.vue';
 export { default as MResponsiveActionBar } from './src/ResponsiveActionBar.vue';
+export { default as MInlineActionBar } from './src/InlineActionBar.vue';
```
landondurnan, line 5: I'm kind of confused about the differences between ResponsiveActionBar and InlineActionBar or why you would use one over the other. I can see that InlineActionBar has more functionality, but is there a scenario to simplify both into a single MActionBar?

src/components/ActionBar/src/InlineActionBar.vue
```diff
@@ -26,14 +44,18 @@ export default {
...
-			default: 'static',
+			default: 'absolute',
 			validator: (position) => ['static', 'relative', 'absolute', 'fixed'].includes(position),
 		},
 		hideOn: {
 			type: String,
 			default: 'none',
 			validator: (hideOn) => ['none', 'mobile', 'desktop'].includes(hideOn),
 		},
+		flush: {
```
landondurnan, line 55: When would you use flush? Typically we'd used that to adjust CSS padding, but this is doing some conditional rendering?

src/components/ActionBar/src/InlineActionBar.vue
```diff
@@ -47,6 +69,16 @@ export default {
 	background-image: linear-gradient(rgba(255, 255, 255, 0), rgba(255, 255, 255, 1));
 }
 
+.ActionBarWrapper {
```
landondurnan, line 72: Eventually we'll need to account for this margin spacing and how it's calculated in case there's some scenarios in which it might change, but more of a mental note -- this is fine for now.

src/components/Modal/src/Modal.vue
```diff
@@ -32,7 +32,7 @@
 
 	.Modal {
 		width: 600px;
-		max-height: calc(100vh - 48px);
+		max-height: calc(100vh - 64px);
```
landondurnan, line 35: Add a comment here about the calc, I assume it's to make space for the actionbar? Or is this simply to provide vertical padding above / below modal on desktop?

src/components/Modal/src/ModalLayer.vue
```diff
@@ -126,6 +148,15 @@ export default {
 	background-color: rgba(0, 0, 0, 0.3);
 }
 
+.Transparent {
+	background-color: transparent;
+}
+
+.Hidden {
+	opacity: 0 !important;
```
landondurnan, line 156: Is there a conflict with opacity somewhere that the important is necessary?

landondurnan (commented):
Some comments / questions, overall looks good.

src/components/Modal/src/Modal.vue
pretzelhammer: Nope, this has nothing to do for actionbar, this just so that the Modal has some 32px of margin above and below it if the modal content overflows the modal. I got these numbers from Shimaa.

src/components/Modal/src/ModalLayer.vue
pretzelhammer: Yes, the fade transition puts a `style: opacity` directly on the element so the `!important` is necessary. I can add an inline comment to explain this.

src/components/ActionBar/src/InlineActionBar.vue
pretzelhammer: It's necessary in the scenario where `InlineActionBar` is used by `ActionBarLayer`. Well, I guess I could try refactoring `InlineActionBar` into an `InlineActionBar` and an `AtomicActionBar`. That might be better.

src/components/ActionBar/index.js
pretzelhammer: After the refactor the only place `ResponsiveActionBar` is used is from local doc dev so I can test that both `ActionBar` and `InlineActionBar` work how I expect, lol. I'll remove the `ResponsiveActionBar` export and remove it from the docs, and try to add more explanation on when to use `ActionBar` vs `InlineActionBar`, although if it's being used in any kind of responsive scenario the answer almost always will be `InlineActionBar`. Merging the 2 would be a breaking change so I'd like to hold off on that before we get some feedback from users, but yes, the end goal would eventually be to have a single `ActionBar` component.

src/components/Modal/src/ModalLayer.vue
pretzelhammer: Every time I've cleaned them out I've put them back in a day later, they serve as really good debugging info but sure, I can remove them.

src/components/ActionBar/index.js
pretzelhammer: I removed `ResponsiveActionBar` and updated the `ActionBar` docs.

src/components/ActionBar/src/InlineActionBar.vue
pretzelhammer: I did the refactor.

landondurnan (approved):

PR #21, opened 2021-03-03
feat: Update button state styles

pretzelhammer:
part of Maker design audit

https://square.github.io/maker/styleguide/update-button-state-styles/#/Button
updates: hover, active, and focus styles, and padding on tertiary/ghost variant

**Other information:**
I asked Design whether they wanted these changes propagated to ActionBar and SegmentedControl and they said:
> _"i would say leave these states on the buttons them selfs and don’t apply those changes to segmented control and actionbar"_

So not adding these updates to buttons in ActionBar and SegmentedControl is not an oversight but an intention omission.


src/components/Button/src/Button.vue
```diff
@@ -51,38 +51,77 @@ function getContrast(chromaBg, targetChromaFg) {
...
+	// mix color with 10% black
+	return chroma.mix(chromaColor, '#000', 0.1);
 }
 
 function fill(tokens) {
 	const color = chroma(tokens.color);
+	const colorHover = getHover(color);
+	const colorActive = getActive(color);
 	const textColor = tokens.textColor ? chroma(tokens.textColor) : undefined;
 	const contrastColor = getContrast(color, textColor);
```
lokesh-squareup, line 72: Didn't know we were doing basic contrast checking in Maker. This is better than nothing, thought it doesn't line up with how the contrast works in website-springboard as there is more nuance to the color flipping.

lokesh-squareup (approved):

src/components/Button/src/Button.vue
```diff
@@ -227,13 +266,12 @@ export default {
 	cursor: pointer;
 	transition:
 		color 0.2s ease-in,
-		background-color 0.2s ease-in;
+		background-color 0.2s ease-in,
```
lokesh-squareup, line 269: Curious, do we share any variables between components in Maker? for example, would we add a `hover-transition` in a shared spot and use between component?

lokesh-squareup (approved):

src/components/Button/src/Button.vue
pretzelhammer: yep, we will probably import the contrast checks from website to maker or from maker to website at some point.

src/components/Button/src/Button.vue
pretzelhammer: nope, we don't have any CSS sharing yet, that's a TODO that's available for anyone to tackle

PR #28, opened 2021-03-09
feat: update stepper to include a size slot

Ryguy11:
Added size variations to stepper component
https://www.figma.com/file/2cgnI0Cb5L7DqdCNiIRrot/%F0%9F%8D%81-Maker-Design-System?node-id=2379%3A19590

- Small stepper size uses small buttons, and primary button type
- Medium and large buttons use medium and large buttons respectively, and use secondary button type

Screen Shot 2021-03-09 at 2 02 31 PM

**Other information:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@square/maker",
-  "version": "1.2.3",
+  "version": "1.3.0",
   "description": "📚 Maker Design System by Square",
   "license": "Apache-2.0",
   "files": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@square/maker",
-  "version": "1.2.2",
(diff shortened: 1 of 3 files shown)


Ryguy11 (commented):
I think I need to add logic to adjust icon sizes within the buttons depending on variation sizes

EDIT: actually I think it looks worse when I do that - going to leave as is

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@square/maker",
-  "version": "1.2.2",
```
pretzelhammer, line 3: Since this is a new feature / backwards-compatible additive change you should increment the minor version instead of the patch version, e.g. `1.3.0` instead of `1.2.3`.

pretzelhammer (approved):

PR #29, opened 2021-03-10
docs: Contributing docs

pretzelhammer:
contributing docs

- [Rendered CONTRIBUTING.md](https://github.com/square/maker/blob/contributing-docs/.github/CONTRIBUTING.md)
- [Rendered WHAT_IS_MAKER.md](https://github.com/square/maker/blob/contributing-docs/.github/WHAT_IS_MAKER.md)
- [Rendered Styleguide README.md](https://github.com/square/maker/blob/contributing-docs/styleguide/README.md)
- [Rendered Lab README.md](https://github.com/square/maker/blob/contributing-docs/lab/README.md)

Also updates the Maker index to split Styleguide Deploys into Versioned Releases and WIP Branches: https://square.github.io/maker/


landondurnan (approved):

.github/WHAT_IS_MAKER.md
```diff
@@ -0,0 +1,63 @@
+# What is Maker?
+
+Most people contributing to and using Maker will be familiar with [Orbit](https://github.com/square/orbit) and will likely have a lot of wrong ideas about Maker's goals by assuming they're the same as Orbit's. This is incorrect. Maker is not Orbit. Their key high-level differences in a table:
```
lokesh-squareup, line 3:
```suggestion
Most people contributing to and using Maker will be familiar with [Orbit](https://github.com/square/orbit), but might have the wrong idea about Maker's goals, by assuming they're the same as Orbit's. This is incorrect. Maker is not Orbit. Their key high-level differences in a table:
```

.github/WHAT_IS_MAKER.md
```diff
@@ -0,0 +1,63 @@
...
+
+| | Orbit | Maker |
```
lokesh-squareup, line 5: This is awesome. Can you give examples of high-level and mid-levels components in parens on the API Design row.

.github/WHAT_IS_MAKER.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+
+### When should I update a Maker component vs implementing a one-off in my project?
+
+This question assumes you are a product engineer working on the product based on product designs you received from a product designer.
+
+First of all, Maker is both a Vue.js component library and a design system. The Maker Design System (MDS) has its own Figma file with designs for all of the components within the design system. To make an engineering-friendly analogy, this Figma file is like the `master` branch of MDS. The `master` branch of the Maker repo, which contains the Maker Vue.js Components (MVC), should always faithfully represent the components in the `master` branch of MDS, i.e. its Figma file.
+
+Second, at any given point in time, a product designer can be in one of two buckets: they can be an Ideal Designer or they can be a Cowboy Designer.
```
lokesh-squareup, line 36: I think this section bucketing designers into Ideal Designers and Cowboy Designers is amazing, but problematic.

Setting up the scene:
1. Maker's feature set is limited and development moves slowly and deliberately. There are limited resources dedicated to it.
2. Maker doesn't always have the best UI solutions for a given scenario.

When these two factors are in play, it often makes sense for a designer to explore a solution outside of Maker. The Maker Designers should look to these solutions and see if they are worth rolling back up in to Maker. 

At a certain point in the future when Maker is further along and more polished, I think this will be an exception. Till then, I think Maker should continue to grow but should be adopting patterns and designs from the product designers who are on the front lines vs the other way around.

.github/WHAT_IS_MAKER.md
```diff
@@ -0,0 +1,63 @@
...
+| **Maintanence** | <ul><li>Maintained by a single team</li><li>Only one team had the authority to approve & publish changes</li></ul> | <ul><li>Collectively maintained by all teams</li><li>All teams have authority to approve & publish changes (which includes taking responsiblity for breaking other teams' code if that becomes the case)</li></ul> |
+| **Versioning** | <ul><li>Follows semver conventions</li><li>Infrequent but large major releases with many breaking changes across many components</li><li>Rarely publishs pre-releases</li></ul> | <ul><li>Follows semver conventions</li><li>Frequent but small major releases with breaking changes typically scoped to a single component</li><li>Regularly publishs pre-releases to experiment with and stabilize eventual breaking API changes before making official releases</li></ul> |
+
+
+
+## FAQs
+
+
```
lokesh-squareup, line 16: Can you address the question: 

> Should I use maker?

If I'm a product engineer, should I use it? Or should I hold off till I'm explicitly told to do so from my designer?

lokesh-squareup (approved):
@pretzelhammer Great job on this! 

In particular, thanks for noting the current principles. Getting that down in writing and having it be clear can be challenging.

.github/WHAT_IS_MAKER.md
pretzelhammer: Kinda hard to clearly explain it in some inline parens within the table, I may try to go into an more in-depth explanation later somewhere else in this doc.

PR #30, opened 2021-03-10
style: update modal breakpoint

Ryguy11:
Modal breakpoint of 1200px was too large.. it was decided to reduce this to 840px
Reduced breakpoints from 1200px to 840px for modals, the inline-action-bar, and the transition variable.

**Other information:**
I left the atomic action bar and action bar layer at 1200px since those breakpoints were unrelated to modals specifically.

### DIFF
--- a/src/utils/transitions.js
+++ b/src/utils/transitions.js
@@ -1,5 +1,6 @@
 export const mobileMinWidth = 0;
-export const desktopMinWidth = 840;
+export const tabletMinWidth = 840;
+export const desktopMinWidth = 1200;
 
 export const stiffness = 600;
 export const damping = 60;
@@ -56,5 +57,6 @@ export default {
 	stiffness,
 	damping,
 	mobileMinWidth,
+	tabletMinWidth,
 	desktopMinWidth,
 };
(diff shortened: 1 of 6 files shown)


src/utils/transitions.js
```diff
@@ -1,5 +1,5 @@
 export const mobileMinWidth = 0;
-export const desktopMinWidth = 1200;
```
pretzelhammer, line 2: desktopMinWidth is still `1200`, this PR introduces a new breakpoint for tablets, you should add a new const called `tabletMinWidth = 840` and use that within the responsive transition in the modal component.

pretzelhammer (approved):

pretzelhammer:
@mattmewton github releases are a totally independent thing from npm releases, so they each have to be maintained separately. I haven't bothered with github releases yet because nobody uses them or needs them. Most projects just use github releases as a place to publish a changelog between versioned releases, which we'll probably start doing eventually with Maker, but I'm thinking about ways we could automate it.

Also, the project is built and released by:
```bash
$ git checkout master
$ git pull
$ npm run build
$ npm publish (only works if you have an npm account that's under the square org on npmjs.com)
```
Although that's another thing I'm looking into ways of automating.

PR #41, opened 2021-03-22
feat(calendar)!: Calendar component

sohnj877:
Created localized calendar that allows to disable dates by setting min/max date or specific dates.

Default
Screen Shot 2021-03-22 at 12 55 27 PM
Japanese
Screen Shot 2021-03-22 at 1 31 49 PM
https://square.github.io/maker/styleguide/calendar-component/#/Calendar

**Other information:**


pretzelhammer:
Overall it looks good, just a couple minor questions regarding the API:
- I noticed `minDate`, `maxDate`, and `selectedDate` are all `String`s and not `Date`s, is there a reason you chose to use `String`s and parse them into `Date`s vs just using `Date`s directly?
- I noticed `selectedDate` is not required, which is fine, but if no `selectedDate` is passed to the component it immediately selects and emits the current date.
    - If it's a requirement that some date is always selected in the Calendar I would prefer the API to be more explicit and for the dev to pass in a `selectedDate` themselves, which should be made a required prop.
    - If it's not a requirement that the Calendar always has a selected date then `selectedDate` can stay optional but the component shouldn't auto-select a date in that case and wait for the end-user to interact/explicitly-set-a-date in the component until it emits that a date has been selected.
    
If anyone other than yourself is planning to use this component once it's released please also cc them in this PR so they can take a look.

privatenumber:
This is from my Datepicker in Website. The reason I chose to take in a date string rather than a Date object is to align with the [native input date picker](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date). I also felt allowing passing in Date objects was risky as they're hard to grasp and introduces complexities like time zones.

I recommended this over Slack: highly advise Date manipulations are refactored to use date-fns. Date objects are very tricky to use and has a steep learning curve. With the amount of "utils" and edge-cases I ended up writing (and polluting the components with), it's more than justified to introduce a light weight Date util that will make code cleaner and easier to maintain.

pretzelhammer:
@privatenumber has more experience than me with this particular component and if he recommends we use [date-fns](https://github.com/date-fns/date-fns) then let's use that lib.

Btw, would using that lib allow us to use Date objects as the prop values for this component or would you still recommend sticking with strings?

src/components/Calendar/src/Calendar.vue
```diff
@@ -0,0 +1,448 @@
...
+		disabledDates: {
+			type: Array,
+			default: () => ([]),
+			validator: (v) => v.every((date) => isDateFormat(date)),
+		},
+	},
+
+	data() {
+		return {
+			headerButtonColor: '#f2f2f2',
```
mattdeclaire, line 149: How come this one is in `data` while color values are used directly in other cases?

sohnj877: I can't use CSS var here b/c I'm passing a color to the button component

package.json
```diff
@@ -116,5 +116,8 @@
     "webpack": "^4.44.2",
     "webpack-cli": "^3.3.12",
     "webpack-distsize": "^0.1.0"
+  },
+  "dependencies": {
+    "date-fns": "^2.19.0"
```
privatenumber, line 121: All dependencies should either be peerDependencies (so it can be de-duped by the parent app) or devDependencies (so it can be bundled in).

Can you try making it a devDependency to see the size impact? If it's big, we can make it a peer dep.

src/components/Calendar/src/Calendar.vue
```diff
@@ -0,0 +1,455 @@
...
+	addDays,
+	format,
+	isToday,
+	isAfter,
+	isSameDay,
+	isSameMonth,
+	isSameYear,
+	addMonths,
+	parse,
+} from 'date-fns';
```
privatenumber, line 96: This seems to be bundling in all of date-fns, which explains the `657.9 kB` size.

(The total size is much larger due to source maps)
Screen Shot 2021-03-24 at 3 33 03 PM


We should consider upgrading to Webpack 5 for better tree-shaking. In the mean time, you can either:
1. Import each method directly by path. eg. `import format from 'date-fns/format'`;
2. Use [`babel-plugin-date-fns`](https://github.com/date-fns/babel-plugin-date-fns) to automatically convert to the above syntax

privatenumber:
> Btw, would using that lib allow us to use Date objects as the prop values for this component or would you still recommend sticking with strings?

I personally would. Its just safer and much more explicit that way. Even with using date-fns, we're still moving dates around in a "broken date container". I believe we also just get the dates from the backend as a string, so it's a convenience to keep it that way.

Also note, website-springboard doesn't use date-fns.

src/components/Calendar/src/Calendar.vue
```diff
@@ -0,0 +1,455 @@
...
+import * as dateFnsLocale from 'date-fns/locale';
```
privatenumber, line 97: Can we get away with not bundling this in? This adds a lot and we previously didn't need it.

If not, you should exclude locales we don't support: https://date-fns.org/v2.19.0/docs/webpack

privatenumber:
Just noticed this is just the Calendar, meant to be manually stitched together with an Input component.

I think this is a bad decision because every Calendar usage will use the Input, so it's just shifting the library code to the application code, but because of that the usage code gets duplicated each time.

pretzelhammer:
@privatenumber yup, this component was just suppose to be the MVP Calendar component, and like you mentioned there are use-cases without an input, so if an input were to be added that should go in a follow-up PR.

src/components/Calendar/src/Calendar.vue
mattdeclaire: I meant to ask why use `data()` at all instead of just `<m-button color="#f2f2f2"...`?

src/components/Calendar/src/Calendar.vue
```diff
@@ -0,0 +1,460 @@
...
+	addMonths,
+	parse,
+} from 'date-fns';
+import { MButton } from '@square/maker/components/Button';
+import ChevronLeft from '@square/maker-icons/ChevronLeft';
+import ChevronRight from '@square/maker-icons/ChevronRight';
+
+const DATE_STRING_FORMAT = 'yyyy-MM-dd';
+const isDateFormat = (string) => !string || /^\d{4}-\d{2}-\d{2}$/.test(string); // YYYY-MM-DD
+const getLocale = (locale) => import(`date-fns/locale/${locale}/index.js`);
```
privatenumber, line 103: JS libraries can't have asynchronous imports because that's an application concern.

Let's remove the date-fns localization dependency and use the native browser localization like it was doing before.

Use date-fns for basic date hadnling though.

privatenumber:
A little concerning this is 62 kB minified (almost as big as Vue.js minified). Will review this soon and make a PR against it.

privatenumber (changes requested):
Sorry for the wait.

I made my review here: https://github.com/square/maker/pull/48

privatenumber (approved):
Calendars looking great!
Thanks for your patience @sohnj877

Merge whenever you're ready and it will trigger an auto-release action.

PR #43, opened 2021-03-24
docs(input): remove double props section

privatenumber:
Props section is now generated but Input docs had old props section left over from before

Removed old props section heading

**Other information:**


privatenumber:
Make some of these CIs can be updated to only run when it's a feat/fix, but can be triggered manually via comment.

pretzelhammer (approved):

PR #46, opened 2021-03-29
feat: Add choice component

sohnj877:
Created choice component that allows to select a single option or multiple options

Screen Shot 2021-03-29 at 10 53 28 AM
https://square.github.io/maker/styleguide/choice-component/#/Choice

**Other information:**

### DIFF
--- a/src/components/Choice/README.md
+++ b/src/components/Choice/README.md
@@ -1,0 +1,186 @@
+
+# Choice
+
+## Examples
+
+### Default
+
+```vue
+<template>
+	<div>
+		<m-choice v-model="selected">
+			<m-choice-option value="choice-1">
+				Choice
+			</m-choice-option>
+			<m-choice-option value="choice-2">
+				Choice
+			</m-choice-option>
+			<m-choice-option value="choice-3">
+				Choice
+			</m-choice-option>
+			<m-choice-option
+				:disabled="true"
+				value="choice-4"
+			>
+				Choice
+			</m-choice-option>
+		</m-choice>
+		<br>
+		<p>Selected value: {{ selected }}</p>
+	</div>
+</template>
+
+<script>
+import { MChoice, MChoiceOption } from '@square/maker/components/Choice';
+
+export default {
+	components: {
+		MChoice,
+		MChoiceOption,
+	},
+	data() {
+		return {
+			selected: 'choice-1',
+		};
+	},
+};
+</script>
+```
+
+### Multi select mode
+
+```vue
+<template>
+	<div>
+		<m-choice
+			v-model="selected"
+			mode="multi-select"
+		>
+			<m-choice-option value="choice-1">
+				Choice
+			</m-choice-option>
+			<m-choice-option value="choice-2">
+				Choice<br>second line
+			</m-choice-option>
+			<m-choice-option value="choice-3">
+				Choice longer text
+			</m-choice-option>
+			<m-choice-option
+				:disabled="true"
+				value="choice-4"
+			>
+				Choice disabled
+			</m-choice-option>
+		</m-choice>
+		<br>
+		<p>Selected value: {{ selected }}</p>
+	</div>
+</template>
+
+<script>
+import { MChoice, MChoiceOption } from '@square/maker/components/Choice';
+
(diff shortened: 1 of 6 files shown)


src/components/Choice/README.md
```diff
@@ -0,0 +1,186 @@
...
+		return {
+			selected: 'choice-1',
+		};
+	},
+};
+</script>
+
+<style scoped>
+.choice-option {
+	width: 172px;
```
mattmewton, line 145: Shouldn't the text inside the component be responsible for setting the width instead of a static width via CSS?

src/components/Choice/src/Choice.vue
```diff
@@ -0,0 +1,96 @@
...
+		prop: 'selected',
+		event: 'choice:update',
+	},
+
+	props: {
+		/**
+		 * Selected choice option
+		 */
+		selected: {
+			type: undefined,
```
mattmewton, line 32: Is there a reason we want to leave this `undefined`?

src/components/Choice/src/ChoiceOption.vue
```diff
@@ -0,0 +1,110 @@
...
+
+export default {
+	inject: {
+		controlState: key,
+	},
+
+	props: {
+		value: {
+			type: undefined,
+			required: true,
```
mattmewton, line 24: How can a type that is undefined also be required?

src/components/Choice/src/ChoiceOption.vue
```diff
@@ -0,0 +1,110 @@
...
+				if (currentValueArray.includes(currentValue)) {
+					currentValueArray.splice(currentValueArray.indexOf(currentValue), 1);
+				} else {
+					currentValueArray.push(currentValue);
+				}
+
+				currentValue = currentValueArray;
+			}
+
+			this.controlState.currentValue = currentValue;
```
mattmewton, line 53: I'm not entirely sure that any props watching `controlState` will respond to this value being set if it's deeply nested. Probably not an issue unless you intend for `controlState` to be watched in this way.

src/components/Choice/src/Choice.vue
pretzelhammer: I believe this is because `selected` is either a `String` in `single-select` mode and an `Array` in `multi-select` mode. Although `undefined` works here I would try `[String, Array]` because I believe Vue also allows passing an array of constructors in the `type` field of the prop validation objects.

mattmewton: @pretzelhammer agreed, I think `[String, Array]` makes more sense here

src/components/Choice/src/ChoiceOption.vue
pretzelhammer: All of the docs examples show using `String`s as values, but @sohnj877 is your intent here to allow any possible value, e.g. `Number`s, `Object`s, `String`s, `Array`s?

src/components/Choice/src/Choice.vue
pretzelhammer: Oh, @sohnj877 intent here might be to allow _any_ value in `single-select` mode, not just `String`s but also `Number`s, `Object`s, etc hence why the type is `undefined`.

src/components/Choice/README.md
pretzelhammer: It does. But text can have uneven lengths and if the dev is going for a more uniform appearance they would have to set a fixed width themselves. `flexbox` doesn't work here because we'd have to set an explicit width on the parent container which we can't do here because `MChoice` is suppose to be allowed to have arbitrary dynamic widths while being thrown into a carousel-like component so that it's horizontally scrollable if it overflows.

src/components/Choice/src/Choice.vue
sohnj877: My intention was to allow any types.

src/components/Choice/src/ChoiceOption.vue
sohnj877: @pretzelhammer Correct

pretzelhammer (approved):

PR #48, opened 2021-03-30
review(calendar): review changes

privatenumber:
PR review for https://github.com/square/maker/pull/41

Specific changes commented inline.

Size change: `62.3 kB` -> [`15.4 kB`](https://github.com/square/maker/pull/49)

**Other information:**

@pretzelhammer :
- Highly recommend adding tests for stuff like this
- I corrected JSDoc pretty seriously here. It's pretty helpful when using VueDX. What do you think about linting them so they're correctly used?

Side note:
- The button styles on active state behaves a little strange to me. When holding the mouse down for a second too long, the border shows up and disappears. Is this expected?

### DIFF
--- a/src/components/Calendar/README.md
+++ b/src/components/Calendar/README.md
@@ -66,7 +66,7 @@ export default {
 			minDate: '',
 			maxDate: '',
 			disabledDates: [],
-			locale: 'en-US',
+			locale: undefined,
 			localeSelection: [
 				'en-US',
 				'da',
@@ -98,7 +98,6 @@ export default {
 		const minDay = `0${today.getDate() - 1}`.slice(-2);
 		const disabledDay = `0${today.getDate() + 5}`.slice(-2);
 
-		this.selectedDate = `${year}-${month}-${selectedDay}`;
 		this.minDate = `${year}-${month}-${minDay}`;
 		this.maxDate = `${year}-${maxMonth}-${selectedDay}`;
 		this.disabledDates.push(`${year}-${month}-${disabledDay}`);
@@ -110,22 +109,18 @@ export default {
 <!-- api-tables:start -->
 ## Props
 
-Supports attributes from [`<div>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div).
-
-| Prop           | Type     | Default   | Possible values | Description                           |
-| -------------- | -------- | --------- | --------------- | ------------------------------------- |
-| locale         | `string` | `'en-US'` | —               | Language specific format to use       |
-| v-model        | `string` | `''`      | —               | Selected date value (YYYY-MM-DD)      |
-| min-date       | `string` | `''`      | —               | Disable the dates before the min-date |
-| max-date       | `string` | `''`      | —               | Disable the dates after the max-date  |
-| disabled-dates | `array`  | `[]`      | —               | List of disabled dates                |
+| Prop           | Type     | Default | Possible values | Description                                          |
+| -------------- | -------- | ------- | --------------- | ---------------------------------------------------- |
+| locale         | `string` | —       | —               | Calendar locale. Defaults to browser locale.         |
+| v-model        | `string` | —       | —               | Selected date value in ISO format.                   |
+| min-date       | `string` | —       | —               | Disable the dates before the min-date in ISO format. |
+| max-date       | `string` | —       | —               | Disable the dates after the max-date in ISO format.  |
+| disabled-dates | `array`  | `[]`    | —               | List of disabled dates in ISO format.                |
 
 
 ## Events
 
-Supports events from [`<div>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div).
-
-| Event           | Type | Description |
-| --------------- | ---- | ----------- |
-| calendar:update | -    | —           |
+| Event           | Type     | Description  |
+| --------------- | -------- | ------------ |
+| calendar:update | `string` | New ISO date |
 <!-- api-tables:end -->
--- a/src/components/Calendar/README.md
+++ b/src/components/Calendar/README.md
@@ -111,11 +111,11 @@ export default {
 
 | Prop           | Type     | Default | Possible values | Description                                          |
 | -------------- | -------- | ------- | --------------- | ---------------------------------------------------- |
-| locale         | `string` | —       | —               | Calendar locale. Defaults to browser locale.         |
 | v-model        | `string` | —       | —               | Selected date value in ISO format.                   |
 | min-date       | `string` | —       | —               | Disable the dates before the min-date in ISO format. |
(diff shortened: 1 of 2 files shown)


src/components/Calendar/README.md
```diff
@@ -66,7 +66,7 @@ export default {
 			minDate: '',
 			maxDate: '',
 			disabledDates: [],
-			locale: 'en-US',
+			locale: undefined,
```
privatenumber, line 69: To demonstrate the default browser locale detection

src/components/Calendar/src/Calendar.vue
```diff
@@ -121,214 +109,180 @@ export default {
...
 			default: () => ([]),
-			validator: (v) => v.every((date) => isDateFormat(date)),
+			validator: (disabledDates) => disabledDates.every((date) => isIsoFormat(date)),
+		},
+		/**
+		 * Calendar locale. Defaults to browser locale.
+		 */
+		locale: {
+			type: String,
+			default: undefined,
```
privatenumber, line 148: `toLocale` methods detect the browser's locale when set to `undefined`. This should be the default behavior instead of `en-US`.

src/components/Calendar/src/Calendar.vue
```diff
@@ -121,214 +109,180 @@ export default {
...
-		shouldDisableHeaderButton(n) {
-			if (n > 0 && this.maxDate) {
-				const maxDate = this.getDateObj(this.maxDate);
-				return isSameMonth(maxDate, this.viewingDate);
-			} if (n < 0 && this.minDate) {
-				const minDate = this.getDateObj(this.minDate);
-				return isSameMonth(minDate, this.viewingDate);
-			}
 			return false;
 		},
```
privatenumber, line 232: Broken behavior.

This logic only checks by month and doesn't factor in year.

src/components/Calendar/src/Calendar.vue
```diff
@@ -121,214 +109,180 @@ export default {
...
 
-			[...emptyDays, ...eachDay].forEach((date, index) => {
-				const key = Math.floor(index / 7);
-				if (!weeks[key]) {
-					weeks[key] = [];
+			const weeks = [];
+			let week = [];
+			while (dateOfMonth.getMonth() === month) {
+				const weekday = dateOfMonth.getDay();
+				week[weekday] = new Date(dateOfMonth);
```
privatenumber, line 177: I noticed a lot of the methods below were re-computing the Date object from a date string, so I decided to just use Date objects here.

This resulted in being able to remove methods `getDeriveViewingDateObj` and `deriveViewingDate`.

src/components/Calendar/src/Calendar.vue
```diff
@@ -81,30 +79,20 @@
...
 	startOfWeek,
 	startOfMonth,
-	eachDayOfInterval,
-	getDay,
-	lastDayOfMonth,
+	endOfMonth,
 	addDays,
-	format,
 	isToday,
-	isAfter,
```
privatenumber, line 90: Removed in favor of using larger/smaller than operators for date comparison.

src/components/Calendar/src/Calendar.vue
```diff
@@ -81,30 +79,20 @@
 import {
 	startOfWeek,
 	startOfMonth,
-	eachDayOfInterval,
-	getDay,
-	lastDayOfMonth,
+	endOfMonth,
 	addDays,
-	format,
```
privatenumber, line 88: Removed in favor of `formatISO`

src/components/Calendar/src/Calendar.vue
```diff
@@ -81,30 +79,20 @@
...
 	isToday,
-	isAfter,
-	isSameDay,
-	isSameMonth,
-	isSameYear,
```
privatenumber, line 93: Removed in favor of comparing ISO date strings.

src/components/Calendar/src/Calendar.vue
```diff
@@ -81,30 +79,20 @@
...
 	addMonths,
-	parse,
```
privatenumber, line 95: Removed in favor of `parseISO`.

src/components/Calendar/src/Calendar.vue
```diff
@@ -121,214 +109,180 @@ export default {
...
 		},
 
 		weekdays() {
-			const date = new Date();
-			const firstDOW = startOfWeek(date);
-			const weekdays = [...new Array(7)].map((_, i) => addDays(firstDOW, i));
-			return weekdays.map((d) => d.toLocaleDateString(this.locale, { weekday: 'short' }));
+			const firstDayOfWeek = startOfWeek(new Date());
+			const weekdays = Array.from({ length: 7 }, (_, i) => addDays(firstDayOfWeek, i));
+			return weekdays.map((date) => date.toLocaleDateString(this.locale, { weekday: 'short' }));
```
privatenumber, line 166: In the Website version, designer Hannah Atkin wanted the English version of the calendar to display single-character headings (`S M T W T F S`) via:

```js
if (localizedDays.toString() === 'Sun,Mon,Tue,Wed,Thu,Fri,Sat') {
	return localizedDays.map(d => d.slice(0, 2));
}
```

Do we not want that anymore?

src/components/Calendar/src/Calendar.vue
```diff
@@ -402,7 +356,6 @@ export default {
 	overflow: hidden;
 	color: var(--color-black);
 	text-align: center;
-	cursor: pointer;
```
privatenumber, line 405: Since the disabled state was moved to the button-element inside of this table-cell, the `cursor: pointer` should also be moved down, otherwise disabled button _table-cells_ will still have pointer cursors.

src/components/Calendar/src/Calendar.vue
```diff
@@ -6,12 +6,11 @@
 	>
 		<div :class="$s.CalendarHeader">
 			<m-button
-				:disabled="shouldDisableHeaderButton(-1)"
+				:disabled="isCalendarNavDisabled(-1)"
 				color="#f2f2f2"
 				size="small"
 				variant="primary"
-				tabindex="-1"
```
privatenumber, line 13: Removed `tabindex` for accessibility.

The `tabindex` was previously disabled when it was in a popover because the calendar was only open while focused on the input.

pretzelhammer (approved):

src/components/Calendar/src/Calendar.vue
sohnj877: Nope

sohnj877 (approved):
Thanks!

src/components/Calendar/src/Calendar.vue
mattdeclaire: Is removing a lot of these methods where the majority of the file savings came from?

privatenumber: Yep!

Issue #39, opened 2021-03-17
Add ability to share CSS between components


pretzelhammer:
**Describe the bug**
Currently the tablet & desktop breakpoints are being copied & pasted between components, this is very bug-prone. Other variables can be shared too.

**Expected behavior**
There should be a single source of truth for what the tablet & desktop breakpoints are and it should be imported into every component that has responsive CSS. Again, same is true for other shared CSS variables.

**When does this need to be addressed by?**
The sooner the better.

Note: there's no requirement to de-duplicate the shared CSS. If some CSS gets duplicated it's okay, we can handle de-duplication as a follow-up optimization task.

This issue is available for anyone to take.

PR #50, opened 2021-03-30
feat: success and error button variants

ashjtan:
_THIS PR IS NOT READY TO MERGE_

https://jira.sqprod.co/browse/WBX-6004

Added the success and button variants, which auto selects the color / text color and adds in the respective icon.
Screen Shot 2021-03-29 at 7 05 08 PM

**Other information:**
I made this PR to receive feedback before continuing on with this feature.

### DIFF
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -483,6 +483,16 @@
 		>
 			One-Off Reversed Button
 		</m-button>
+			<m-button
+				variant="success"
+			>
+				Success button
+			</m-button>
+			<m-button
+				variant="error"
+			>
+				Error button
+			</m-button>
 	</div>
 </template>
 
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -534,18 +534,18 @@ body {
 
 Supports attributes from [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).
 
-| Prop       | Type      | Default     | Possible values                    | Description                           |
-| ---------- | --------- | ----------- | ---------------------------------- | ------------------------------------- |
-| type       | `string`  | `'button'`  | —                                  | Type of the button                    |
-| size       | `string`  | `'medium'`  | `small`, `medium`, `large`         | Size of the button                    |
-| full-width | `boolean` | `false`     | —                                  | Whether to make the button full-width |
-| color      | `string`  | `'#000'`    | —                                  | Background color of button            |
-| text-color | `string`  | —           | —                                  | Text color of button                  |
-| variant    | `string`  | `'primary'` | `primary`, `secondary`, `tertiary` | Semantic variant                      |
-| shape      | `string`  | `'rounded'` | `squared`, `rounded`, `pill`       | Shape of button                       |
-| disabled   | `boolean` | `false`     | —                                  | Toggles button disabled state         |
-| align      | `string`  | `'center'`  | `center`, `stack`, `space-between` | How to align button's contents        |
-| loading    | `boolean` | `false`     | —                                  | Toggles button loading state          |
+| Prop       | Type      | Default     | Possible values                                        | Description                           |
+| ---------- | --------- | ----------- | ------------------------------------------------------ | ------------------------------------- |
+| type       | `string`  | `'button'`  | —                                                      | Type of the button                    |
+| size       | `string`  | `'medium'`  | `small`, `medium`, `large`                             | Size of the button                    |
+| full-width | `boolean` | `false`     | —                                                      | Whether to make the button full-width |
+| color      | `string`  | `'#000'`    | —                                                      | Background color of button            |
+| text-color | `string`  | —           | —                                                      | Text color of button                  |
+| variant    | `string`  | `'primary'` | `primary`, `secondary`, `tertiary`, `success`, `error` | Semantic variant                      |
+| shape      | `string`  | `'rounded'` | `squared`, `rounded`, `pill`                           | Shape of button                       |
+| disabled   | `boolean` | `false`     | —                                                      | Toggles button disabled state         |
+| align      | `string`  | `'center'`  | `center`, `stack`, `space-between`                     | How to align button's contents        |
+| loading    | `boolean` | `false`     | —                                                      | Toggles button loading state          |
 
 
 ## Slots
--- a/src/components/Button/src/Button.vue
+++ b/src/components/Button/src/Button.vue
@@ -126,10 +126,57 @@ function ghost(tokens) {
 	};
 }
 
+function success(tokens) {
+	const color = chroma('#1fad1f');
+	const colorHover = getHover(color);
+	const colorActive = getActive(color);
+	const textColor = tokens.textColor ? 'rgba(255, 255, 255, 0.95)' : undefined;
+	const contrastColor = getContrast(color, textColor);
+	const contrastColorHover = getHover(contrastColor);
+	const contrastColorActive = getActive(contrastColor);
+	const focusColor = getFocus(color);
+	return {
+		'--small-padding': '8px 16px',
+		'--medium-padding': '12px 24px',
+		'--large-padding': '20px 32px',
+		'--color-main': color.hex(),
+		'--color-main-hover': colorHover.hex(),
+		'--color-main-active': colorActive.hex(),
+		'--color-contrast': contrastColor.hex(),
+		'--color-contrast-hover': contrastColorHover.hex(),
+		'--color-contrast-active': contrastColorActive.hex(),
+		'--color-focus': focusColor.hex(),
+	};
+}
+function error(tokens) {
+	const color = chroma('#ce3217');
+	const colorHover = getHover(color);
+	const colorActive = getActive(color);
+	const textColor = tokens.textColor ? 'rgba(255, 255, 255, 0.95)' : undefined;
+	const contrastColor = getContrast(color, textColor);
+	const contrastColorHover = getHover(contrastColor);
+	const contrastColorActive = getActive(contrastColor);
+	const focusColor = getFocus(color);
+	return {
+		'--small-padding': '8px 16px',
+		'--medium-padding': '12px 24px',
+		'--large-padding': '20px 32px',
+		'--color-main': color.hex(),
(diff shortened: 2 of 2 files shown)


src/components/Button/src/Button.vue
```diff
@@ -17,6 +17,19 @@
...
+		>
+			<component
+				:is="iconComponent"
+				:class="[
+					$s.Icon,
+					$s[`type_${type}`],
+				]"
+				inline
+			/>
+		</div>
```
ashjtan, line 32: i would prefer to add this into the MainText section somehow so that it mimics
```
<m-button>
  <check-circle class="icon" />
</m-button>
```
but i couldn't figure it out.  when i placed it inside the MainText section, the svg colors weren't passed in correctly.

src/components/Button/src/Button.vue
```diff
@@ -126,10 +141,57 @@ function ghost(tokens) {
 	};
 }
 
+function success(tokens) {
```
ashjtan, line 144: ok there's gotta be a less copy-pasty way to do this right? this is just `fill()` with hardcoded `color` and `textColor`.

src/components/Button/src/Button.vue
```diff
@@ -416,4 +489,28 @@ export default {
 .Button.align_space-between .InformationText {
 	margin-right: 8px;
 }
+
```
ashjtan, line 492: if i can figure out how to put the icons in the `$s.MainText` section, i won't need these classes

src/components/Button/src/Button.vue
mattmewton: A couple things here:

1. I think this outer wrapper should be a `<span>` so it's natively an inline element inside the button, unless you need it to be a `<div>` for some reason.
2. We should be able to use the existing icon prefix slot perhaps to do this, so we don't need to create another condition where an icon is added. Unless that messes up the preferred alignment of this type of button, as the prefix icons are always aligned left? cc @pretzelhammer for your thoughts on this

src/components/Button/src/Button.vue
```diff
@@ -17,6 +17,19 @@
 		v-bind="$attrs"
 		v-on="$listeners"
 	>
+		<div
+			v-if="variant === 'success' || 'error'"
```
mattmewton, line 21: Let's make this a method, `isSuccessOrErrorVariant(variant)`

src/components/Button/src/Button.vue
```diff
@@ -126,10 +141,57 @@ function ghost(tokens) {
...
+	const color = chroma('#1fad1f');
+	const colorHover = getHover(color);
+	const colorActive = getActive(color);
+	const textColor = tokens.textColor ? 'rgba(255, 255, 255, 0.95)' : undefined;
+	const contrastColor = getContrast(color, textColor);
+	const contrastColorHover = getHover(contrastColor);
+	const contrastColorActive = getActive(contrastColor);
+	const focusColor = getFocus(color);
+	return {
+		'--small-padding': '8px 16px',
```
mattmewton, line 154: We have variables for all these, I think

src/components/Button/src/Button.vue
mattmewton: Probably worth creating a util function that spits out these values holistically. `getFillColor`, or something to that effect

src/components/Button/src/Button.vue
```diff
@@ -189,7 +253,7 @@ export default {
 		variant: {
 			type: String,
 			default: 'primary',
-			validator: (variant) => ['primary', 'secondary', 'tertiary'].includes(variant),
+			validator: (variant) => ['primary', 'secondary', 'tertiary', 'success', 'error'].includes(variant),
```
mattmewton, line 256: We might want to consider making these constants at some point if we're going to repeat the strings everywhere

src/components/Button/src/Button.vue
```diff
@@ -416,4 +489,28 @@ export default {
...
+
+.IconAligner {
+	display: flex;
+	align-items: center;
+	height: 24px;
+	margin-right: 8px;
+}
+
+.Icon {
+	width: 16px;
```
mattmewton, line 501: Again let's try to use the spacing variables if Maker has them available. If not, we should add them

pretzelhammer: there's an open issue free for anyone to take if they want to tackle CSS reuse within Maker: https://github.com/square/maker/issues/39

pretzelhammer:
@ashjtan given the designs I think you may have meant to make these changes on [`ActionBarButton`](https://github.com/square/maker/blob/master/src/components/ActionBar/src/ActionBarButton.vue) which is the button component used inside ActionBars.

src/components/Button/src/Button.vue
pretzelhammer: The string literals are already their own constants. Turning them into JS constants makes the code more verbose with no benefit:

```js
// some/constants/file.js

export const SUCCESS = 'success';
export const ERROR = 'error';

// some/vue/component.vue

import { SUCCESS, ERROR } from 'some/constants/file.js';
// ...
validator(variant) => [SUCCESS, ERROR].includes(variant),
```

vs just keeping it:

```js
// some/vue/component.vue

validator(variant) => ['success', 'error'].includes(variant),
```

It's "easier to refactor" the previous version _but not really_. For example, if we had:
```js
export const ERROR = 'error';
```

And we wanted to rename it to `'failure'` we wouldn't just do this:
```js
export const ERROR = 'failure';
```
And leave it at that, because now it's confusing that `ERROR = 'failure'` and not `ERROR = 'error'` in the rest of the codebase, so we'd much more likely do this instead:
```js
export const FAILURE = 'failure';
```
Which ironically doesn't actually save us any work from just using string literals in the first place because now we have to go over the entire codebase and change `ERROR` into `FAILURE` anyway.

Furthermore, if the constants are used in the public API of the components then changing them is a breaking change and requires a major version release so people very rarely, if ever, change them because it's often not worth the hassle.

I think it's okay for us to stick to using string literals as their own constants for now. I'm willing to revisit this idea in the future if it can be shown it provides a benefit that's actually greater than its maintenance cost.

src/components/Button/src/Button.vue
pretzelhammer: or `variantHasPrefixIcon(variant)`

src/components/Button/src/Button.vue
pretzelhammer: @ashjtan 

When you tried

```vue
<m-button>
    <check-circle class="icon" /> Some text here
</m-button>
```

Did you have these styles on the `.icon` class?:

```css
.icon {
    width: 16px;
    height: 16px;
    fill: currentColor;
    // and maybe even
    stroke: currentColor;
}
```

It should be inheriting the text color if it has those styles.

@mattmewton 

MInputs and MSelects have icon slots because it's otherwise not possible to set icons inside them. With buttons is possible to pass icons straight through using the default slot so I'd rather not add unnecessary icon slots to button.

src/components/Button/src/Button.vue
pretzelhammer: Ideally we wouldn't need any of these classes and the icons can be passed straight through the default slot. Please let us know what you have tried so far and why it's not working.

src/components/Button/src/Button.vue
```diff
@@ -230,6 +294,15 @@ export default {
 				textColor: this.textColor,
 			});
 		},
+		iconComponent() {
```
pretzelhammer, line 297: I don't think it's a good idea to hardcode these icons as part of the variants themselves. If we cannot pass the icons through the default slot for some reason I'd suggest we push back on design here and ask if it's alright to just change the color of the button without also prepending the prefix icons.

src/components/Button/README.md
```diff
@@ -483,6 +483,16 @@
 		>
 			One-Off Reversed Button
 		</m-button>
+			<m-button
```
pretzelhammer, line 486: I would put these variants into their own standalone documentation example rather than adding them to this already massive one.

src/components/Button/src/Button.vue
mattmewton: Works for me!

mattdeclaire: I've always been on the fence about constants. I agree it doesn't help with refactoring. But, I also thought constants help catch typos.
```js
if (variant === SUCCES) // unknown variable error
if (variant === 'succes') // harder to spot bug
```
That being said, it's a lot of overhead.

ashjtan:
> @ashjtan given the designs I think you may have meant to make these changes on [`ActionBarButton`](https://github.com/square/maker/blob/master/src/components/ActionBar/src/ActionBarButton.vue) which is the button component used inside ActionBars.

@pretzelhammer true, i just saw that the `ActionBarButton` uses the `Button` and wasn't sure if i should add `ActionBarButton` variants or `Button` variants and then pass them in. but i can just work on the `ActionBarButton`

ashjtan:
thanks for the feedback everyone! gonna take it and work on `ActionBarButton` instead and see what happens 😃

src/components/Button/src/Button.vue
ashjtan: @pretzelhammer to clarify, ur example
```
<m-button>
    <check-circle class="icon" /> Some text here
</m-button>
```
works fine. but the idea was to have the icons included as part of the variant automatically so that i could get the same result by just doing `<m-button variant="success">`. i wanted to achieve this by adding in the icon in the MainText section like
```
<span :class="$s.MainText">
    <span v-if="isSuccessOrErrorVariant>
        <check-circle or alert-circle class="icon" />
    </span>
    <!-- @slot Button label -->
    <slot />
</span>
```
but when i did it this way, even with the `.icon` styles you put above, the svg renders w/ `fill=none` so the icon is basically invisible. that's why i ended up adding all the complexity i did, with the icons [outside of the MainText section](https://github.com/square/maker/pull/50/files#diff-a63a381a9ca0c90e9b1158a50b1c4bfbb0cb75b95488098dba716e13324c2342R20-R32).

given this, do you think that the icon should not be included in the maker `ActionBarButton` (i.e. what you advocated for in a [below comment](https://github.com/square/maker/pull/50#discussion_r604170952))?

PR #53, opened 2021-04-01
ci: refactor and use nvm

privatenumber:
- Workflows weren't using nvm to detect correct Node.js version
- Some workflows didn't have names
- Some workflows could be consolidated

- Added names to every workflow
- Use nvm to detect and use correct Node.js version
- Consolidated workflows that have the same trigger.

**Other information:**


pretzelhammer (approved):

privatenumber:
I renamed the workflows, so the old ones will appear hanging and wont ever run.

Issue #55, opened 2021-04-02
bug: missing utils/transitions


privatenumber:
**Describe the bug**

`utils/transitions.js` is missing from package versions [v3.0.0](https://unpkg.com/browse/@square/maker@3.0.0/utils/) and [v4.0.0](https://unpkg.com/browse/@square/maker@4.0.0/utils/). It's present in [v2.2.4 (the version before)](https://unpkg.com/browse/@square/maker@2.2.4/utils/).

When building:
```
ERROR in ./node_modules/@square/maker/components/Modal/script.js
Module not found: Error: Can't resolve '../../utils/transitions' in '/Users/osame/Documents/ecom-github/weebly/website-springboard/node_modules/@square/maker/components/Modal'

ERROR in ./node_modules/@square/maker/components/TransitionFadeIn/script.js
Module not found: Error: Can't resolve '../../utils/transitions' in '/Users/osame/Documents/ecom-github/weebly/website-springboard/node_modules/@square/maker/components/TransitionFadeIn'

ERROR in ./node_modules/@square/maker/components/TransitionSpringUp/script.js
Module not found: Error: Can't resolve '../../utils/transitions' in '/Users/osame/Documents/ecom-github/weebly/website-springboard/node_modules/@square/maker/components/TransitionSpringUp'

ERROR in ./node_modules/@square/maker/utils/Transition/script.js
Module not found: Error: Can't resolve '../../utils/transitions' in '/Users/osame/Documents/ecom-github/weebly/website-springboard/node_modules/@square/maker/utils/Transition'
```

**Expected behavior**

For `utils/transitions.js` to exist.

**Reproduction steps (incl. Browser and OS)**

- Can't reproduce locally
- Seems to be happening on GH actions

**When does this need to be addressed by?**

**Which version are you using? Did this work in previous versions?**

**Which version of Vue.js are you using? Did this work in previous versions?**

privatenumber:
### Relevant links
#### v3.0.0
PR: https://github.com/square/maker/pull/41
Release logs: https://github.com/square/maker/runs/2249349049?check_suite_focus=true

#### v4.0.0
PR: https://github.com/square/maker/pull/47
Release logs: https://github.com/square/maker/runs/2249395456?check_suite_focus=true


### Observations

Both release logs show that `utils/transitions.js` was built:

```
Child
    Hash: cb93081400e15ea9e468
    Time: 3266ms
    Built at: 04/01/2021 8:26:16 PM
[...]

PR #56, opened 2021-04-02
fix: remove clean-webpack-plugin

privatenumber:

- Race condition that deletes `transitions.js`

- Remove `clean-webpack-plugin`
- Closes https://github.com/square/maker/issues/55

### DIFF
--- a/.github/workflows/lint-pr-title.yml
+++ b/.github/workflows/lint-pr-title.yml
@@ -5,6 +5,7 @@ on:
     types:
       - opened
       - edited
+      - synchronize
 
 jobs:
   lint-pr-title:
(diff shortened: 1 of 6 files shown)


pretzelhammer:
The only downside of this is, if we ever remove/rename a util or component we'd have to know to manually nuke the dist folders ourselves, right? Because if we're expecting every newer version to just overwrite/add to the previous version then just running `npm run build` would work fine but... OH WAIT, this is all on CI which always has a clean starting environment anyway. Nevermind, the gotcha above only applies if we're manually doing releases but that's not the case here.

pretzelhammer (approved):

privatenumber:
Might happen on lab/styleguide releases that have previous deploys. If there's issue there, I recommend doing `rm -rf` over adding a plugin. Alternatively, upgrading to Webpack 5 has a "clean" feature built in.

PR #59, opened 2021-04-06
feat(actionbar): add window.esc to add ability to close modal on ESC

sohnj877:

We need the ability to close the modal on ESC key press

Added `closeOnEsc` prop

## Other information

`closeOnEsc` isn't supported on the stacked modal

### DIFF
--- a/src/components/ActionBar/src/ActionBarButton.vue
+++ b/src/components/ActionBar/src/ActionBarButton.vue
@@ -35,7 +35,7 @@
 		</span>
 		<pseudo-window
 			document
-			@keyup.esc="$emit('window-esc')"
+			@keyup.esc="handleEscKey"
 		/>
 	</button>
 </template>
@@ -157,6 +157,14 @@ export default {
 				(vnode) => vnode.tag || vnode.text.trim().length > 0,
 			);
 			return children.length === 1 && children[0].tag;
+		},
+
+		handleEscKey() {
+			/**
+			 * ESC keyup event
+			 * @property {string}
+			 */
+			this.$emit('window-esc');
 		},
 	},
 };
--- a/src/components/ActionBar/src/ActionBarButton.vue
+++ b/src/components/ActionBar/src/ActionBarButton.vue
@@ -33,11 +33,16 @@
 				name="information"
 			/>
 		</span>
+		<pseudo-window
+			document
+			@keyup.esc="$emit('window-esc')"
+		/>
 	</button>
 </template>
 
 <script>
 import chroma from 'chroma-js';
+import PseudoWindow from 'vue-pseudo-window';
 import { MLoading } from '@square/maker/components/Loading';
 
 // TODO: refactor the code below so it's shared with Button component
@@ -74,6 +79,7 @@ function fill(tokens) {
 export default {
 	components: {
 		MLoading,
+		PseudoWindow,
 	},
 
 	inheritAttrs: false,
--- a/src/components/ActionBar/src/ActionBarButton.vue
+++ b/src/components/ActionBar/src/ActionBarButton.vue
@@ -161,7 +161,7 @@ export default {
 
 		handleEscKey() {
 			/**
-			 * ESC keyup event
+			 * ESC keyup event on window
 			 * @property {string}
 			 */
 			this.$emit('window-esc');
--- a/src/components/Modal/src/Modal.vue
+++ b/src/components/Modal/src/Modal.vue
@@ -4,13 +4,22 @@
 			<!-- @slot Modal content -->
 			<slot />
 		</div>
+		<pseudo-window
+			document
+			@keyup.esc="handleEscKey"
+		/>
 	</div>
 </template>
 
 <script>
+import PseudoWindow from 'vue-pseudo-window';
 import modalApi from './modal-api';
 
 export default {
+	components: {
+		PseudoWindow,
+	},
+
 	inject: {
 		modalApi,
 	},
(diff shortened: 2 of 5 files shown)


src/components/Modal/src/Modal.vue
```diff
@@ -7,6 +7,55 @@
...
+			if (this.closeOnEsc) {
+				document.addEventListener('keyup', this.handleEscKey);
+			}
+		},
+
+		removeEvents() {
+			if (this.closeOnEsc) {
+				document.removeEventListener('keyup', this.handleEscKey);
+			}
+		},
```
pretzelhammer, line 47: You can replace the ~20 lines above using [vue-pseudo-window](https://github.com/privatenumber/vue-pseudo-window) which is a dep we already include in Maker. After removing these lines you'd just add this to the template above:

```vue
<pseudo-window
	document
	@keyup="handleEscKey"
/>
```

src/components/Modal/src/Modal.vue
```diff
@@ -7,6 +7,55 @@
...
+
+		handleEscKey(event) {
+			const isClosingStackedModal = !!this.modalApi.state.vnode;
```
pretzelhammer, line 50: I'm assuming since this is here you've tested that this prop works and behaves correctly with stacked nested modals, right?

src/components/Modal/src/Modal.vue
privatenumber: `keyup.esc="handleEscKey"`

privatenumber:
Thoughts:
- Wondering if this shouldn't be behind a prop. I think it might be a good default. If need be, we can add an opt-out flag later.
- Making it easy to close modals introduces risk of losing data (eg. populating a form in a modal and accidentally hitting esc). We should consider a confirmation alert as a follow up.

src/components/Modal/src/Modal.vue
sohnj877: Yup

pretzelhammer (approved):

src/components/Modal/src/Modal.vue
```diff
@@ -4,9 +4,47 @@
...
+		closeOnEsc: {
+			type: Boolean,
+			default: false,
+		},
+	},
+
+	methods: {
+		handleEscKey(event) {
+			const isClosingStackedModal = !!this.modalApi.state.vnode;
+			if (this.closeOnEsc && !isClosingStackedModal && event.key === 'Escape') {
```
privatenumber, line 40: Remove `event.key` check

src/components/Modal/src/Modal.vue
```diff
@@ -4,9 +4,47 @@
...
+	},
+
+	props: {
+		/**
+		 * Close the modal on ESC
+		 */
+		closeOnEsc: {
+			type: Boolean,
+			default: false,
+		},
```
privatenumber, line 34: Remove prop. Should be default behavior.

privatenumber (changes requested):

src/components/Modal/src/Modal.vue
sohnj877: We this the prop b/c we don't allow users to close the location modal until location is selected

privatenumber: Wondering if this check can be improved. eg. if the Modal has a Cancel/Close button, it automatically supports esc to close, but doesn't if not.

privatenumber:
Just checked out the branch. Doesn't seem to be working with stacked modals.

src/components/Modal/src/Modal.vue
privatenumber: Doesn't seem to be the case in the Stacked Modal demo: https://square.github.io/maker/styleguide/close-modal-on-esc/#/Modal

src/components/Modal/src/Modal.vue
privatenumber: Since whether this should be enabled is bound to when the Cancel/Close button passed in, I think it makes more sense to couple the logic too.

Currently, `close-on-esc` needs to be passed into the user-end when the Close button is passed in. It might not be that much more work to instead pass in `<pseudo-window @keyup.esc="modalApi.close()" />` , and move the `hasModalOnTop` check to `close()`.

Taking this thought further, if the esc key is actually triggering the close button, why not add a `trigger-on-esc` prop or support a `@window.esc` event on the button?
```vue
<m-action-bar-button
	...
	@click="modalApi.close()"
	@window-esc="modalApi.close()"
>
	...
</m-action-bar-button>
```

cc @pretzelhammer

sohnj877: I decided to add `window-esc` event to the modal and let the developer to close the modal or show alert before closing.

privatenumber: The problem I'm surfacing is that the `esc` and clicking Cancel are equivalent, but the logic is spread apart (eg. cancel is handled by button, esc is handled by modal), and therefore you have to be aware of two API surfaces to enable it.

Regarding the before-close confirmation alert, we'll need something more robust to handle cases like the user accidentally closing the window (prompt native alert to prevent), or in the future if we support "clicking out of modal to close" they might accidentally click out of the modal. In Orbit we had a promise based hook `before-close`.

pretzelhammer: @privatenumber I like the `window-esc` prop on the button idea, however how would you expect the case where there are 2 modals stacked on top of each other, both with cancel/close buttons, to work?

privatenumber: The `hasModalOnTop` check should move inside modal `close()` to prevent closing modals that aren't active.

This should actually already be there because the modals below are only hidden via `opacity: 0`, so there is currently a bug where you can actually "tab" back and "click" on the close button of a hidden modal. As a side note, have you considered doing `display: none` on it after the fade-out? This will be an accessibility issue for those navigating with a keyboard.

On Slack, @sohnj877 also wondered if there's a case where we want the "esc to close" but no Cancel button. I don't think that should ever happen either because it's inaccessible to those that can't use a keyboard.

privatenumber (changes requested):
> [The hasModalOnTop check should move inside modal close() to prevent closing modals that aren't active.](https://github.com/square/maker/pull/59#discussion_r610908621)

Looks like this isn't implemented yet ^

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -33,11 +33,16 @@
 				name="information"
 			/>
 		</span>
+		<pseudo-window
+			document
+			@keyup.esc="$emit('window-esc')"
```
privatenumber, line 38: Add event description here for the Event table in the docs

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -158,6 +158,14 @@ export default {
 			);
 			return children.length === 1 && children[0].tag;
 		},
+
+		handleEscKey() {
+			/**
+			 * ESC keyup event
```
privatenumber, line 164:
```suggestion
			 * ESC keyup event on window
```

privatenumber (approved):
Great work @sohnj877

PR #60, opened 2021-04-07
feat(modal): multi-pane modals

sohnj877:

We need the ability to navigate through the multiple modal content with left/right arrows

## Other information

Arrows are visible on desktop only


landondurnan:
@sohnj877 The arrows are only visible on desktop. Is there a design and output for mobile? I'm wondering if we should actually leverage the actionbar instead.

sohnj877:
@landondurnan The arrows were designed for desktop only and there is no design for mobile.

landondurnan (changes requested):
We shouldn't build solutions for desktop only and I'm concerned we're getting ahead of ourselves with a solution when there's only one example of this currently.

src/components/Modal/src/ModalArrow.vue
```diff
@@ -0,0 +1,70 @@
...
+</script>
+
+<style module="$s">
+.ModalArrow {
+	--arrow-color: #fff;
+	--arrow-size: 24px;
+	--arrow-size-half: calc(var(--arrow-size) / 2);
+
+	position: absolute;
+	top: calc(50% - var(--arrow-size-half));
```
pretzelhammer, line 54: if you have `top: calc(50% - var(--arrow-size-half))` here then I'm pretty sure you don't need the `transform: translateY(-50%)` below because all those do is move the arrow up half its height...

src/components/Modal/src/ModalArrow.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+	<span
```
pretzelhammer, line 2: is this extra `span` necessary? why not just put these styles on the inner component?

src/components/Modal/src/ModalArrow.vue
```diff
@@ -0,0 +1,70 @@
...
+		:class="[
+			$s.ModalArrow,
+			$s[`position_${position}`],
+		]"
+		@click="$emit('click')"
```
pretzelhammer, line 7: `v-bind="$listeners` so all listeners are proxied, and not just clicks

src/components/Modal/src/ModalArrow.vue
```diff
@@ -0,0 +1,70 @@
...
+			:is="icon"
+			:class="$s.ModalArrowIcon"
+		/>
+	</span>
+</template>
+
+<script>
+import ChevronLeft from '@square/maker-icons/ChevronLeft';
+import ChevronRight from '@square/maker-icons/ChevronRight';
+import modalApi from './modal-api';
```
pretzelhammer, line 19: please remove this import if it's not being used

src/components/Modal/src/Modal.vue
```diff
@@ -1,12 +1,47 @@
 <template>
-	<div :class="$s.Container">
-		<div :class="$s.Modal">
-			<!-- @slot Modal content -->
-			<slot />
+	<div :class="$s.ArrowsContainer">
+		<m-modal-arrow
+			v-if="hasLeftArrow"
+			:class="$s.modalArrow"
```
pretzelhammer, line 5: what is this class? `modalArrow` is not declared anywhere in this component

src/components/Modal/src/Modal.vue
```diff
@@ -1,12 +1,47 @@
...
+		/>
+		<div :class="$s.Container">
+			<div :class="$s.Modal">
+				<!-- @slot Modal content -->
+				<slot />
+			</div>
 		</div>
+		<m-modal-arrow
+			v-if="hasRightArrow"
+			:class="$s.modalArrow"
```
pretzelhammer, line 17: please remove class if it is not declared anywhere

src/components/Modal/src/Modal.vue
```diff
@@ -1,12 +1,47 @@
 <template>
-	<div :class="$s.Container">
-		<div :class="$s.Modal">
-			<!-- @slot Modal content -->
-			<slot />
+	<div :class="$s.ArrowsContainer">
```
pretzelhammer, line 2: where is the `ArrowsContainer` styles? they're not declared in the style tag below, this class is not doing anything

src/components/Modal/README.md
```diff
@@ -547,14 +547,213 @@ export default {
...
 | ------- | ------------- |
 | default | Modal content |
 
 
+## Modal Events
+
+| Event    | Type | Description |
+| -------- | ---- | ----------- |
+| go-left  | -    | —           |
+| go-right | -    | —           |
```
pretzelhammer, line 754: we can probably think of better names for these events, maybe just simply `left` and `right` or `navigate-left` or `navigate-right` to be more explicit/descriptive

src/components/Modal/README.md
```diff
@@ -547,14 +547,213 @@ export default {
...
+
+<style scoped>
+.cover-photo {
+	width: 100%;
+	height: 600px;
+	object-fit: cover;
+	object-position: center;
+}
+
+.icon {
```
pretzelhammer, line 726: this class doesn't seem to be used anywhere above

sohnj877:
We won't add the arrows until we have a good solution for both mobile and desktop

PR #64, opened 2021-04-15
fix: hides actionbarlayer padding if actionbar not present

pretzelhammer:

people complained about empty padding at the bottom of their sites when no actionbar is present

removes padding from actionbarlayer when actionbar is not present

## Other information

nope


privatenumber:
Does this create a sudden scroll bump whenever the action bar appears?

pretzelhammer:
@privatenumber this is usually an issue if the height/padding is being added _above_ content, but in this case all of the padding/height is being added to the _very bottom of the page below_ all content (since we instruct people to put ActionBarLayer at the root of their apps) and so it's harmless, no content is shifted vertically on the page, and yeah, technically the scroll bar gets a tiny bit smaller but it's pretty much unnoticeable

PR #66, opened 2021-04-21
fix(actionbar): actionbar buttons flickering

sohnj877:

Action bar buttons were flickering and had border/outline around the buttons while loading the modal on mobile devices
actionbar-before

Added `-webkit-transform: translate3D(0, 0, 0)`. See: https://stackoverflow.com/questions/10134935/weird-css3-transition-flickering
actionbar-after

### DIFF
--- a/src/components/ActionBar/src/ActionBarLayer.vue
+++ b/src/components/ActionBar/src/ActionBarLayer.vue
@@ -113,7 +113,7 @@ export default {
 
 .Action {
 	margin-right: 8px;
-	-webkit-transform: translate3d(0, 0, 0);
+	-webkit-transform: translate3d(0, 0, 0); /* Fixes buttons flickering on mobile devices */
 	filter: drop-shadow(0 15px 10px rgb(0 0 0 / 20%));
 
 	&:last-child {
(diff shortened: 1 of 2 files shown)


src/components/ActionBar/src/ActionBarLayer.vue
```diff
@@ -113,6 +113,7 @@ export default {
 
 .Action {
 	margin-right: 8px;
+	-webkit-transform: translate3d(0, 0, 0);
```
pretzelhammer, line 116: Can you please add a comment above this style that explains its purpose? As-is it looks like it does nothing and is very likely to get removed by another engineer if there's no comment.

src/components/ActionBar/src/AtomicActionBar.vue
```diff
@@ -99,6 +99,7 @@ export default {
 
 .Action {
 	margin-right: 8px;
+	-webkit-transform: translate3d(0, 0, 0);
```
pretzelhammer, line 102: same feedback as above

pretzelhammer (approved):

PR #70, opened 2021-04-28
ci: perform styleguide deploy after semantic release

pretzelhammer:
runs styleguide deploys after semantic releases in CI

no screenshots

## Other information

maybe it works or maybe it doesn't, there's no way of knowing until we merge into master

note to future self: if it doesn't work it's probably because the github checkout action does not check out tags along with the head commit, so look into that if this fails

### DIFF
--- a/.github/workflows/deploy-styleguide.yml
+++ b/.github/workflows/deploy-styleguide.yml
@@ -28,6 +28,7 @@ jobs:
           cd .dist
           git config user.name github-actions
           git config user.email github-actions@github.com
+          cd ..
           nvm i
           npm ci
           npm run styleguide-deploy
--- a/.github/workflows/deploy-styleguide.yml
+++ b/.github/workflows/deploy-styleguide.yml
@@ -23,8 +23,6 @@ jobs:
       - name: Build & deploy
         shell: bash -l {0}
         run: |
-          git config user.name github-actions
-          git config user.email github-actions@github.com
           cd .dist
           git config user.name github-actions
           git config user.email github-actions@github.com
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,3 +21,19 @@ jobs:
         nvm i
         npm ci
         npx semantic-release
+    - name: Checkout nested Maker Repo deploys branch in .dist directory
+      uses: actions/checkout@v2
+      with:
+        repository: square/maker
+        ref: deploys
+        path: .dist
+    - name: Build & deploy styleguide
+      shell: bash -l {0}
+      run: |
+        git config user.name github-actions
+        git config user.email github-actions@github.com
+        cd .dist
+        git config user.name github-actions
+        git config user.email github-actions@github.com
+        cd ..
+        npm run styleguide-deploy
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,8 +30,6 @@ jobs:
     - name: Build & deploy styleguide
       shell: bash -l {0}
       run: |
-        git config user.name github-actions
-        git config user.email github-actions@github.com
         cd .dist
         git config user.name github-actions
         git config user.email github-actions@github.com
--- a/styleguide/webpack-styleguide-config.js
+++ b/styleguide/webpack-styleguide-config.js
@@ -4,18 +4,13 @@ const webpack = require('webpack');
 const { JustSsrPlugin } = require('vue-just-ssr');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const ent = require('ent');
-const branch = require('git-branch');
 const { merge } = require('../build/utils');
 const webpackBaseConfig = require('../build/webpack-base-config');
 const { version } = require('../package.json');
+const { getCurrentBranch, isSemanticReleaseBranch, getLibVersion } = require('../build/utils');
 
-// branch.sync() doesn't work in the Github Actions env
-let branchName = branch.sync();
-if (!branchName) {
-	// but that's okay because the Github Actions env provides this var
-	branchName = process.env.GITHUB_HEAD_REF;
-}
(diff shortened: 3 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -21,3 +21,19 @@ jobs:
...
+      uses: actions/checkout@v2
+      with:
+        repository: square/maker
+        ref: deploys
+        path: .dist
+    - name: Build & deploy styleguide
+      shell: bash -l {0}
+      run: |
+        git config user.name github-actions
+        git config user.email github-actions@github.com
```
privatenumber, line 34: Are we actually committing something in the top-level repo? Wondering if this can be removed.

styleguide/webpack-styleguide-config.js
```diff
@@ -4,18 +4,13 @@ const webpack = require('webpack');
...
 
-// branch.sync() doesn't work in the Github Actions env
-let branchName = branch.sync();
-if (!branchName) {
-	// but that's okay because the Github Actions env provides this var
-	branchName = process.env.GITHUB_HEAD_REF;
-}
-const deploy = branchName === 'master' ? version : branchName;
+let branchName = getCurrentBranch();
+const deploy = isSemanticReleaseBranch(branchName) ? getLibVersion() : branchName;
```
privatenumber, line 13: My understanding of `getLibVersion()` is that it's specifically for post semantic-release CI.

I'm curious if doing that is necessary--does `semantic-release` actually revert the version bump in `package.json`? I'm thinking not since it's designed for CI and it expects changes to be scrapped.

privatenumber:
You can test this using a custom local npm registry via [verdaccio](https://verdaccio.org/).

Might be able to test it out for you later today or tomorrow.

.github/workflows/release.yml
pretzelhammer: good point, removed

styleguide/webpack-styleguide-config.js
pretzelhammer: `getLibVersion()` is generally useful. I used it to checkout old commits and build & deploy the styleguides for versions: v3.0.0, v4.0.0, v4.0.1, v4.1.0, v4.1.1, v4.1.2, v4.1.3, and v4.1.4 which are all now live on the [public deploy index](https://square.github.io/maker/). It's nice to have a script that works when run locally in case CI ever fails, or in case we ever want do some one-off styleguide deploy independent of CI for whatever reason. I'd like to keep `getLibVersion()` as-is because it works locally.

privatenumber: FYI `npx semantic-release` will work locally too if you want to release yourself

pretzelhammer:
@privatenumber Huh, how would verdaccio help me in this situation? It says it's a private npm registry proxy but... can it also run github actions locally? Well, either way, I don't want to go through the trouble of setting it up, I'm down to merge to master and debug it live if it fails, but if you wanna test it for me then that works for me too :)

privatenumber:
Merging this won't run the action because `ci:` doesn't trigger a release. You'll have to wait till the next `fix`/`feat`/`revert`. At that point, you probably don't want to risk blocking any real releases.

You can't run GitHub Actions locally, but you can run `npx semantic-release` and do a fake test release to the local npm registry.

I can try to find time to do it later today or tomorrow. It's instantaneous to setup.

pretzelhammer:
@privatenumber oh, if this fails it's non-blocking, because it currently doesn't work anyway, so not a big deal, as long as the CI succeeds up to the semantic-release then it's okay if it fails afterward for the styleguide deploy, as I fixed the script so we can deploy the styleguide locally now anyway

.github/workflows/deploy-styleguide.yml
```diff
@@ -23,11 +23,10 @@ jobs:
       - name: Build & deploy
         shell: bash -l {0}
         run: |
-          git config user.name github-actions
-          git config user.email github-actions@github.com
```
privatenumber, line 27: Since you removed this, I think you gotta set the cwd on these commands to the dist directory:

https://github.com/square/maker/blob/master/build/deploys/push.js#L15

privatenumber:
Confirmed that this doesn't revert the `version` field in `package.json` post-publish, so you can read from `package.json`.

In the case that CI becomes unavailable, and you need to publish locally, run from `master`:
```sh
$ npx prerelease-checks && CI=true GH_TOKEN=<GITHUB_TOKEN> npx semantic-release
```

.github/workflows/deploy-styleguide.yml
pretzelhammer: That's a good point.... how in the world did it work before though 🤔

pretzelhammer: Ohhh, it worked before because both `pull.js` and `push.js` call `ensure.js` which changes the directory to the deploy directory here: https://github.com/square/maker/blob/master/build/deploys/ensure.js#L13

Anyway, that's a bit too easy to miss, so I made it more explicit in this commit: https://github.com/square/maker/pull/70/commits/fcde22398068ce924af4ed20a7b3806a5be3c50b

pretzelhammer:
@privatenumber the utility function already reads the package.json version first and only falls back to checking the current git tag if the current package.json version is `0.0.0-semantic-release`: https://github.com/square/maker/pull/70/files#diff-216385f62eec0147a05da4c867a26e4dfc3352e9d822d20ff7ff85e3d8d39516R39-R40

The above code works in all environments, including checking out old commits locally and re-publishing the just the styleguide for that Maker version (e.g. we may change the styleguide build process and want to backport the newer styleguide to older Maker versions, which doesn't include re-publishing the NPM packages, so no calls to semantic-release).

privatenumber:
Gotcha, good to know!

privatenumber (approved):

PR #71, opened 2021-04-29
feat(dialog): initial Dialog component, based on Modal

pretzelhammer:

adds Dialog component

## Other information

nope


landondurnan (approved):
Extremely straightforward to review as it's so similar to modals. Nothing surprising here. Good callout in informing how to setup the layering.

We should in a next iteration try to start sharing more styles, but I know that's a larger system improvement we want to make all over. Nothing blocking this from getting in now.

landondurnan:
Feature request: Tap overlap to close? Has that been discussed yet?

PR #72, opened 2021-04-30
feat: Notice refresh

pretzelhammer:

Notice refresh

- updates Notice status colors (including the Error status color which is updated across all components)
- adds NoticeButton child component to be used with Notice

https://square.github.io/maker/styleguide/notice-refresh/#/Notice

image
image

## Other information

none

### DIFF
--- a/src/components/Select/src/SelectControl.vue
+++ b/src/components/Select/src/SelectControl.vue
@@ -159,7 +159,7 @@ export default {
 	--color-accent: #222;
 	--color-border: transparent;
 	--color-border-active: #222;
-	--color-error: #ff3b30;
+	--color-error: rgba(206, 50, 23, 1);
 	--focus-shadow: 0 0 0 2px rgba(34, 34, 34, 0.3);
 	--border-radius: 8px;
 }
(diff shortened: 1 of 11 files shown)


src/components/Select/src/SelectControl.vue
```diff
@@ -159,7 +159,7 @@ export default {
 	--color-accent: #222;
 	--color-border: transparent;
 	--color-border-active: #222;
-	--color-error: #ff3b30;
+	--color-error: rgba(206, 50, 23, 1);
```
mattmewton, line 162: Why do we need to copy these values over and over again? It would be good to have a centralized variables file for colors that are universal throughout the framework.

pretzelhammer: open issue: [Add ability to share CSS between components](https://github.com/square/maker/issues/39)

> This issue is available for anyone to take.

if you have the bandwidth then go for it

mattmewton: Gotcha, I do not unfortunately

mattmewton (approved):

PR #73, opened 2021-05-19
feat(modal): close modal on click outside and before close hook

Ryguy11:
Modals cannot be closed when you click outside of them.
Add the ability to pass in an options object on modal open, with a parameter to close modals when you click outside of them

### DIFF
--- a/src/components/Modal/README.md
+++ b/src/components/Modal/README.md
@@ -227,7 +227,7 @@ export default {
 </script>
 ```
 ### Configurable options
-The `modalApi.open()` function has a second optional object parameter that you allows for configurable options. Current available options are:
+The `modalApi.open()` function has a second optional object parameter that offers configurable options. Current available options are:
 
 ```html
 {
--- a/src/components/Modal/README.md
+++ b/src/components/Modal/README.md
@@ -34,7 +34,12 @@ export default {
 
 	methods: {
 		openModal() {
-			this.modalApi.open(() => <DemoModal />);
+			this.modalApi.open(
+				() => <DemoModal />,
+				{
+					closeOnOutsideClick: true,
+				},
+			);
 		},
 	},
 };
@@ -405,7 +410,10 @@ export default {
 
 	methods: {
 		openModal() {
-			this.modalApi.open(() => <StackingDemoFirstModal />);
+			this.modalApi.open(
+				() => <StackingDemoFirstModal />,
+				{ closeOnOutsideClick: true },
+			);
 		},
 	},
 };
@@ -476,7 +484,7 @@ export default {
 
 	methods: {
 		openSecondModal() {
-			this.modalApi.open(() => <StackingDemoSecondModal />);
+			this.modalApi.open(() => <StackingDemoSecondModal />, { closeOnOutsideClick: true });
 		},
 		closeFirst() {
 			this.modalApi.close();
--- a/src/components/Modal/README.md
+++ b/src/components/Modal/README.md
@@ -37,7 +37,7 @@ export default {
 			this.modalApi.open(
 				() => <DemoModal />,
 				{
-					closeOnOutsideClick: true,
+					closeOnClickOutside: true,
 				},
 			);
 		},
@@ -136,7 +136,16 @@ export default {
 	methods: {
 		openMyModal() {
 			// this.modalApi is provided by MModalLayer.apiMixin
-			this.modalApi.open(() => <MyModal />);
+			this.modalApi.open(
+				() => <MyModal />,
+				{
+					closeOnClickOutside: true,
+					beforeCloseHook: () => {
+						// Intercept logic here
+						return true; // or false if you want to block modal from closing
+					},
+				},
+			);
 		}
 	}
 };
@@ -216,6 +225,17 @@ export default {
 	},
 };
 </script>
+```
+### Configurable options
+The `modalApi.open()` function has a second optional object parameter that you allows for configurable options. Current available options are:
+
+```html
+{
+	// Modal will close when clicked outside of it - default false
+	closeOnClickOutside: boolean,
+	// function to run before modal is closed. If return value is false modal will not close.
+	beforeCloseHook: () => { return boolean },
+}
 ```
 
 To close a modal on ESC, use the `@window-esc` from the `MActionBarButton` component.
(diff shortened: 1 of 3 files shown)


src/components/Modal/README.md
```diff
@@ -405,7 +410,10 @@ export default {
 
 	methods: {
 		openModal() {
-			this.modalApi.open(() => <StackingDemoFirstModal />);
+			this.modalApi.open(
+				() => <StackingDemoFirstModal />,
+				{ closeOnOutsideClick: true },
```
privatenumber, line 415:
```suggestion
				{
					closeOnOutsideClick: true,
				},
```

src/components/Modal/README.md
```diff
@@ -476,7 +484,7 @@ export default {
 
 	methods: {
 		openSecondModal() {
-			this.modalApi.open(() => <StackingDemoSecondModal />);
+			this.modalApi.open(() => <StackingDemoSecondModal />, { closeOnOutsideClick: true });
```
privatenumber, line 487:
```suggestion
			this.modalApi.open(
				() => <StackingDemoSecondModal />,
				{
					closeOnOutsideClick: true,
				},
			);
```

privatenumber:
Looks good!

I wonder if `closeOnClickOutside` is a better name than  `closeOnOutsideClick` though.

src/components/Modal/src/ModalLayer.vue
```diff
@@ -155,6 +162,16 @@ export default {
 	destroyed() {
 		this.unwatchStackedModal();
 	},
+
+	methods: {
+		closeOnOutsideClick(event) {
+			const { closeOnOutsideClick } = this.currentLayer.state.options;
+			const { modal } = this.$refs;
+			if (modal && closeOnOutsideClick && !modal.contains(event.target)) {
+				this.modalApi.close();
```
privatenumber, line 171: @pretzelhammer expressed concern that this implementation should support close interception (eg. confirmation dialog before close).

IMO, this implementation will allow for that because it closes the modal via `close()`, and interception should intercept all calls to `close()`.

So, if we were to implement something like a `before-close` hook, the specific modal instance can pass in a hook:
```html
<o-modal :before-close="beforeCloseHook">
```

where
```js
async beforeCloseHook() {
   // anything async
  return true; // or false to block
}
```

the `close()` method will be updated to something like:
```js
async close() {
  if (typeof this.options.beforeCloseHook === 'function') {
    if (!(await this.options.beforeCloseHook())) {
      return; // cancel
    }
  }

  // close modal
}
```

pretzelhammer: @Ryguy11 could you add @privatenumber 's suggestion to this PR? I think it would be nice if we had this functionality and knew it worked before settling on an implementation.

src/components/Modal/README.md
```diff
@@ -212,6 +226,17 @@ export default {
...
 ```
+### Configurable options
+The `modalApi.open()` function has a second optional object parameter that you allows for configurable options. Current available options are:
+
+```html
+{
+	// Modal will close when clicked outside of it - default false
+	closeOnClickOutside: boolean,
+	// function to run before modal is closed. If return value is false modal will not close.
+	beforeCloseHook: () => { return boolean },
```
privatenumber, line 237: This hook should be on Modal since that's the file that would be concerned about the sensitivity of its contents. 

API should look like this
```vue
<o-modal :before-close="beforeCloseHook">
```

src/components/Modal/README.md
```diff
@@ -212,12 +231,26 @@ export default {
 };
 </script>
 ```
+### Configurable options
+The `modalApi.open()` function has a second optional object parameter that offers configurable options. Current available options are:
+
+```html
```
privatenumber, line 237:
```suggestion
```ts
```

Or `json5`

src/components/Modal/README.md
```diff
@@ -212,12 +231,26 @@ export default {
...
+{
+	// Modal will close when clicked outside of it - default false
+	closeOnClickOutside: boolean,
```
privatenumber, line 240:
```suggestion
	closeOnClickOutside: boolean;
```

src/components/Modal/README.md
```diff
@@ -243,6 +276,13 @@ export default {
 	inject: {
 		modalApi,
 	},
+
+	methods: {
+		beforeCloseHook() {
```
privatenumber, line 281: Might be good to set the default recommendation to async since the most common use-case would either be to dispatch an async save action or confirmation dialog.

```suggestion
		async beforeCloseHook() {
```

src/components/Modal/src/Modal.vue
```diff
@@ -7,6 +7,29 @@
...
+	name: 'Modal',
+
+	inject: {
+		modalApi,
+	},
+
+	props: {
+		beforeClose: {
+			type: Function,
+			default: () => true,
```
privatenumber, line 23: Remove default. Shouldn't be necessary since you do a check before calling it.

src/components/Modal/src/Modal.vue
```diff
@@ -7,6 +7,29 @@
...
+		},
+	},
+
+	mounted() {
+		this.modalApi.state.options.beforeCloseHook = this.beforeClose;
```
privatenumber, line 28: Since `beforeClose` is a prop, it can be reactive. Instead of putting this in `mounted`, put it in a `watch` with `immediate: true`.

src/components/Modal/src/Modal.vue
```diff
@@ -7,6 +7,37 @@
...
+
+	inject: {
+		modalApi,
+	},
+
+	props: {
+		/**
+		 * Before close hook, can block closing
+		 */
+		beforeClose: { // eslint-disable-line vue/require-default-prop
```
Ryguy11, line 24: @privatenumber I removed the default, but eslint was upset - I disabled this line but not sure if you would prefer me to do something else here

src/components/Modal/src/Modal.vue
Ryguy11: Very good point, I should have done it that way to start

src/components/Modal/src/Modal.vue
privatenumber: Didn't think a default was necessary if the prop was not required.

You can explicitly set default to `undefined` instead.

privatenumber (approved):
Looks good to me. Thanks @Ryguy11

PR #75, opened 2021-05-27
fix(modal): use better logic to decide whether a click is inside of a modal or not

Ryguy11:
Certain child elements of modals were incorrectly triggering modals to close.

`Node.contains` relies on the fact that the element that was clicked is contained within the DOM of the modal. If the following example of code were to be inside of a modal:
```
<span
	v-if="display"
	@click="display = !display"
>
	testing
</span>
```
Whenever the span is clicked, it toggles itself off of the DOM. The click event propagates up to the ModalLayer and the `.contains` logic checks if the span is inside or outside of the modal. Given that the element is no longer in the DOM the modal layer closes the current modal.

While this kind of issue can be stopped at the source using `click.stop` or `v-show` instead of a v-if, I believe that it is easy to imagine more complex implementations of this logic that could cause the modal to close frustrate Maker developers. We should be able to handle this type of situation from the maker side, and developers should not need to worry about their event propagation.

Instead of using `Node.contains` I used `Node.documentPositionContains` which consistently returns the value of 10 or `Node.DOCUMENT_POSITION_PRECEDING + Node.DOCUMENT_POSITION_CONTAINS` whenever a click is registered outside of the modal. By explicitly checking for a click outside of the modal, instead of checking for a click inside and then using a `! NOT` operator, we also have a more redundant solution.

 I have tested this change across multiple contexts, including the style-guide docs, stacked modals, and in our website repo using an `npm link`.

### DIFF
--- a/src/components/Modal/src/ModalLayer.vue
+++ b/src/components/Modal/src/ModalLayer.vue
@@ -187,6 +187,8 @@ export default {
 
 		isClickInsideModal(modal, event) {
 			const pos = modal.getBoundingClientRect();
+			console.log(pos); // eslint-disable-line no-console
+			console.log(`Event x: ${event.clientX} and Event y: ${event.clientY}`); // eslint-disable-line no-console
 			const containsX = event.clientX > pos.left && event.clientX < pos.right;
 			const containsY = event.clientY > pos.top && event.clientY < pos.bottom;
 			return containsX && containsY;
--- a/src/components/Modal/src/ModalLayer.vue
+++ b/src/components/Modal/src/ModalLayer.vue
@@ -178,7 +178,8 @@ export default {
 			if (
 				modal
 				&& closeOnClickOutside
-				&& modal.compareDocumentPosition(event.target) === 10
+				&& modal.compareDocumentPosition(event.target)
+					=== (Node.DOCUMENT_POSITION_PRECEDING + Node.DOCUMENT_POSITION_CONTAINS)
 			) {
 				this.modalApi.close();
 			}
--- a/src/components/Modal/src/ModalLayer.vue
+++ b/src/components/Modal/src/ModalLayer.vue
@@ -175,7 +175,11 @@ export default {
 		closeOnClickOutside(event) {
 			const { closeOnClickOutside } = this.currentLayer.state.options;
 			const { modal } = this.$refs;
-			if (modal && closeOnClickOutside && !modal.contains(event.target)) {
+			if (
+				modal
+				&& closeOnClickOutside
+				&& modal.compareDocumentPosition(event.target) !== 20
+			) {
 				this.modalApi.close();
 			}
 		},
--- a/src/components/Modal/src/ModalLayer.vue
+++ b/src/components/Modal/src/ModalLayer.vue
@@ -14,7 +14,7 @@
 				v-if="currentLayer.state.vnode"
 				ref="baseModalLayer"
 				:class="$s.ModalLayer"
-				@click="closeOnClickOutside"
+				@click.capture="closeOnClickOutside"
 			>
 				<pseudo-window
 					body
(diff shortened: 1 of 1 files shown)


privatenumber:
Can you provide a minimal reproduction of the problem?

Currently, this issue is vaguely worded IMO. I kind of understand the solution but don't really understand the problem.

It will also be more helpful to someone were to revisit this in the future.

src/components/Modal/src/ModalLayer.vue
```diff
@@ -175,7 +175,11 @@ export default {
 		closeOnClickOutside(event) {
 			const { closeOnClickOutside } = this.currentLayer.state.options;
 			const { modal } = this.$refs;
-			if (modal && closeOnClickOutside && !modal.contains(event.target)) {
+			if (
+				modal
+				&& closeOnClickOutside
+				&& modal.compareDocumentPosition(event.target) === 10
```
privatenumber, line 181:
```suggestion
				&& modal.compareDocumentPosition(event.target) === (Node.DOCUMENT_POSITION_PRECEDING + Node. DOCUMENT_POSITION_CONTAINS)
```

Hmm, I thought we had [`no-magic-numbers`](https://eslint.org/docs/rules/no-magic-numbers) enabled.

pretzelhammer (approved):

privatenumber:
Thanks for the updates. Problem is much clearer now.

Given your example, I'm not sure if this is the best solution. In the example you gave, clicking on the toggle has a document position of 37, which is `DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC`, `DOCUMENT_POSITION_FOLLOWING`, and `DOCUMENT_POSITION_DISCONNECTED`. 

`DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC` is still not easy to understand for me.

Wondering if this is a better solution than something simpler like using the capture phase to catch the `click` event (`click.capture`). eg. is the robustness of `compareDocumentPosition` excessive or warranted?

Ryguy11:
> Thanks for the updates. Problem is much clearer now.
> 
> I'm not sure if this is the best solution though. In the example you gave, clicking on the toggle has a document position of 37, which is `DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC`, `DOCUMENT_POSITION_FOLLOWING`, and `DOCUMENT_POSITION_DISCONNECTED`.
> 
> `DOCUMENT_POSITION_IMPLEMENTATION_SPECIFIC` is still not easy to understand for me. And if the element is no longer in the DOM, what does `DOCUMENT_POSITION_FOLLOWING` mean?
> 
> Wondering if this is a better solution than something simpler like:
> 
> * using the capture phase to catch the `click` event (`click.capture`)
> * checking if the target element is still in the DOM via [`.isConnected`](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected)
> 
> Is the robustness of `compareDocumentPosition` excessive or warranted?

@privatenumber I have never heard of the `.capture` event modifier, I think that would be perfect for this situation! Thanks for pointing that out

Ryguy11:
Using @privatenumber's suggestion the only change that needs to be made is adding the `.capture` modifier to the click event handler. I don't believe there is any reason to use the `.isConnected` instead of `.contains`

privatenumber:
Thanks for the update.

Some unanswered Qs:
- **Does this prevent `click` events from getting to the child elements?** No, because `.stopPropagation()` isn't called. 

- **Should `.stopPropagation()` be called then?** We would if there are any click targets outside the modal, above the layer. But in this case, there isn't any so it doesn't matter.

privatenumber (approved):

Ryguy11:
I can add in the `.isConnected` check but wouldn't it just be redundant? Given that we are capturing the click event prior to the event reaching its target destination any action that would stem from the click event, such as removing the element from the DOM, should not have occurred yet. Or am I understanding that wrong?

privatenumber:
I was just explaining the alternative option to using `.capture` just to clarify. No action needed.

PR #79, opened 2021-06-04
feat(action-bar): allow clicking through action bar container

Ryguy11:
The action bar container does not let pointer events through to the layer behind. This is particularly noticeable in situations where the primary action is not visible. This PR allows for pointer events to go through the container but not the buttons.


landondurnan (commented):
Tested and working as expected. I don't think we need to include the test in the readme demos - ideally we can get some unit testing in place and move it there.

landondurnan (approved):

PR #80, opened 2021-06-07
fix(action-bar): Improve device / browser behavior

landondurnan:
The original design for the ActionBar intended the padding to be the same all the way around the buttons and anchored to the bottom. Over time we adjusted the bottom padding to be much higher to address the invisible container that Mobile Safari adds to toggle the browser toolbar (details: https://bugs.webkit.org/show_bug.cgi?id=194235) 

I've updated the behavior so that rather than degrading the design intent for all browsers that we target Mobile Safari and then add the repositioning for that browser.

mobile-action-bar

- Design: Update padding to be consistent 24px.
- Design: Remove responsive button sizing. While this was the original design, we'd like to move to a simpler and more consistent default. Now using default medium button sizes regardless of screen size.
- Utility added to determine if browser is Mobile Safari (iphone / ipad)
- Target actionbar positioning to be higher only for Mobile Safari.
- Fix some linting issues

## Other information

We've already discussed these changes with product / design, so no additional need to align further.

### DIFF
--- a/src/components/ActionBar/src/ActionBarButton.vue
+++ b/src/components/ActionBar/src/ActionBarButton.vue
@@ -47,15 +47,18 @@ import { MLoading } from '@square/maker/components/Loading';
 // TODO: refactor the code below so it's shared with Button component
 
 function getContrast(chromaBg, targetChromaFg) {
-	if (!targetChromaFg || chroma.contrast(chromaBg, targetChromaFg) < 4.5) {
-		const isLight = chromaBg.luminance() > 0.32;
+	const accessibleContrast = 4.5;
+	const lightLuminance = 0.32;
+	if (!targetChromaFg || chroma.contrast(chromaBg, targetChromaFg) < accessibleContrast) {
+		const isLight = chromaBg.luminance() > lightLuminance;
 		return chroma(isLight ? '#000' : '#fff');
 	}
 	return targetChromaFg;
 }
 
 function getFocus(chromaColor) {
-	return chromaColor.alpha(0.8);
+	const colorAdjust = 0.8;
+	return chromaColor.alpha(colorAdjust);
 }
 
 function fill(tokens) {
@@ -153,8 +156,10 @@ export default {
 				return false;
 			}
 			const children = (this.$slots.default || []).filter(
+				// eslint-disable-next-line no-magic-numbers
 				(vnode) => vnode.tag || vnode.text.trim().length > 0,
 			);
+			// eslint-disable-next-line no-magic-numbers
 			return children.length === 1 && children[0].tag;
 		},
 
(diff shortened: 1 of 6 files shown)


src/utils/browser.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-env browser */
+export function isMobileSafari() {
```
pretzelhammer, line 2: can you please write a comment linking the source where you pulled this function from?

src/components/ActionBar/src/AtomicActionBar.vue
```diff
@@ -99,7 +108,7 @@ export default {
 
 .Action {
 	margin-right: 8px;
-	-webkit-transform: translate3d(0, 0, 0);  /* Fixes buttons flickering on mobile devices */
+	transform: translate3d(0, 0, 0);  /* Fixes buttons flickering on mobile devices */
```
pretzelhammer, line 111: Did you confirm that this change doesn't break the intended fix?

src/components/ActionBar/src/AtomicActionBar.vue
```diff
@@ -36,17 +38,28 @@ export default {
...
+			return isMobileSafari();
+		},
+	},
 };
 </script>
 
 <style module="$s">
 .ActionBar {
-	--action-bar-bottom-padding: 64px;
+	--action-bar-padding: 24px;
```
pretzelhammer, line 52: Shouldn't this be 32px? [The line you deleted has it as 32px](https://github.com/square/maker/pull/80/files#diff-6137173e29b51786287282c92fedffb9a92f90d5605a9203d1079fe83711ddf2L64).

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -47,15 +47,18 @@ import { MLoading } from '@square/maker/components/Loading';
 // TODO: refactor the code below so it's shared with Button component
 
 function getContrast(chromaBg, targetChromaFg) {
-	if (!targetChromaFg || chroma.contrast(chromaBg, targetChromaFg) < 4.5) {
-		const isLight = chromaBg.luminance() > 0.32;
+	const accessibleContrast = 4.5;
```
pretzelhammer, line 50: Can you please move these ActionBarButton refactors to a separate PR? They seem to have nothing to do with the detecting Safari.

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -171,20 +176,17 @@ export default {
 
 <style module="$s">
 .Button {
-	--button-large: 64px;
```
pretzelhammer, line 174: Did Design say to remove the large button style on mobile? If not, then why did you do this? Please get alignment with Design before you make any design changes.

src/components/ActionBar/src/AtomicActionBar.vue
landondurnan: Design would prefer to change it to 24px.

src/components/ActionBar/src/ActionBarButton.vue
landondurnan: Yes

src/components/ActionBar/src/ActionBarButton.vue
landondurnan: These aren't refactors, it's resolving lint issues regarding no magic numbers: https://eslint.org/docs/rules/no-magic-numbers

PR #84, opened 2021-06-21
docs: update styleguide styles

landondurnan:
Light style pass on output.

CleanShot 2021-07-20 at 15 57 40

Wasn't able to fix image so removed it.


styleguide/SideNav.vue
```diff
@@ -1,15 +1,22 @@
 <template>
 	<div>
-		<ul>
-			<li
+		<img
+			src="/static/maker-logo.png"
```
privatenumber, line 4: Needs to be a Webpack resolvable path (eg. relative or aliased) according to docs: https://vue-loader.vuejs.org/guide/asset-url.html#transform-rules

landondurnan:
This a low risk change and has been in review for some time now. Going to merge the changes.

Issue #90, opened 2021-06-30
make actionbarbuttons size medium for mobile and tablet+ resolutions


pretzelhammer:
actionbarbutton size is responsive depending on window resolution. on mobile resolutions they are size large and on tablet+ resolutions they are size medium. this task is to make them size medium for all resolutions, i.e. remove size large on mobile resolutions and default to size medium for everything.

pretzelhammer:
https://github.com/square/maker/pull/92

PR #92, opened 2021-07-07
fix(actionbar): make actionbar buttons medium size on all resolutions

pretzelhammer:


pretzelhammer:
https://github.com/square/maker/issues/90

PR #93, opened 2021-07-09
feat(theming)!: adds theming support to buttons

pretzelhammer:
theming mvp beta release, adds theming support to buttons and actionbar buttons

https://square.github.io/maker/styleguide/theming-2/#/Theme

## v4.x to v5.x Migration Guide

- `MTheme` has completely changed, which probably won't require you to refactor or migrate any of your existing code since you probably weren't using it in the first place
- Aside from that there are no breaking changes to any other components, but now you can put a `MTheme` at the root of your app and select a default set of props for MButtons and MActionBarButtons which will be inherited by all of those components within your app


landondurnan (commented):
Why remove the default values on the props? Wouldn't it be possible to leverage the components without the Theme. The Theme just overrides those default properties?

landondurnan (changes requested):
I've been giving this a lot of thought and I don't think there should be a default theme and it'd be preferable to have default props on the components.

Consumers of Maker do not need to use a Theme and they should be able to use a single component and understand those defaults in the documentation.

The Theme is an optional parent component that should provide overrides to the default behavior of the components. Sellers will have their own theme data and even in scenarios where they've customized nothing in their brand/design data, that data would be a better representation of that default. There's no need to duplicate it in Maker.

When we start considering the type of unit and visual regression tests we may do at the component level, we're going to be forced to couple data from another source file further away from the component we're testing. I think we should be very cautious about coupling these components when they're optional.

The schema for our themes will need to do validation, but it may not always be one-to-one with our components. We may want to map things differently.

Also having so many duplicative and verbose computed properties to determine whether to leverage the local component prop or the theme data, is pretty overwhelming. Considering we will continue to have multiple authors of components, this pattern established leaves a lot of room for human error and is pretty confusing. I think this is an opportunity to consider how we can share more code CSS / JS (mixins or directives) to simplify this logic in two extremely similar components: Button and Actionbar Button.

Taking a step back and thinking about the Theme schema and what I would suspect to see.  The theme has a primary color. The default button is primary and uses the primary color. Seeing that there's actually a `theme.button.color` I suppose is alright, but now as a developer I need to know everything about the Theme schema and this could go a lot of different ways. Why not `theme.colors.primary`?  What about how the variants may work, `theme.button.variant.primary.color` It's not really clear what the rules are, they seem like they can be anything. As a component developer how do I know what to use? How much do I need to know about the Theme properties? I think a lot can go wrong here.

src/components/Button/src/Button.vue
```diff
@@ -230,9 +238,45 @@ export default {
 	},
 
 	computed: {
+		resolvedSize() {
```
lokesh-squareup, line 241: Do we want to add validation for resolved computer props to prevent invalid values being passed in via `<m-theme>`?

lokesh-squareup:
@landondurnan
Trying to breakdown your comments:

> it'd be preferable to have default props on the components.

If we add default prop values, will we be able to distinguish when the default is utilized _vs_ when a value is passed in that happens to match the default? In the former, if a theme provider exists, would want to use that value, and in the latter we would want to use the component prop value.

> The Theme is an optional parent component that should provide overrides to the default behavior of the components. Sellers will have their own theme data and even in scenarios where they've customized nothing in their brand/design data, that data would be a better representation of that default. **There's no need to duplicate it in Maker.**

What do you mean by duplicate it in Maker?

> the Theme schema and what I would suspect to see. The theme has a primary color. The default button is primary and uses the primary color.

If you have a section that is using a different color profile, buttons might not use this _primary color_, so the granularity is needed. Also, by outlining all the values in the theme we can see the scope of its impact and the relationships between different values.

I do think there are some good questions in your last paragraph about how developers building apps will use theme data and what understanding they will need. I think that challenge will exist in most scenarios, but I can imagine that developers in apps will reference different tokens as they see fit.


@pretzelhammer 
If no `<m-theme>` is available as a parent, where will the default theme come from? I know you mentioned it could be defined in the component. Is this something you want to change in this PR?

The other option is requiring one `<m-theme>` to use Maker components. You could say that by adding that component you are creating a _maker space_. :-)

lokesh-squareup:
Summary of experience from consuming devs perspective:
image

lokesh-squareup (approved):

laurenhu:
To @landondurnan 's concern about not having default props on the components and requiring the use of a theme to style a component, could we split the difference and include default values in the resolved props rather than as the default prop? I understand that having a default value for the local shape overrides can cause it to always take precedence over a theme prop, but would it satisfy our concerns if within the resolved props, there is a fallback value if no local value or theme value is passed in?

I agree with @lokesh-squareup that the impact on our current usage in website-springboard is minimal and we can always adjust it later on but just a thought.

pretzelhammer:
@laurenhu 

> To @landondurnan 's concern about not having default props on the components and requiring the use of a theme to style a component, could we split the difference and include default values in the resolved props rather than as the default prop?

This is how it works currently.

PR #103, opened 2021-07-22
fix(select): change fill variant color

ashjtan:
The fill color was previously `rgba(0, 0, 0, 0.05)`, which was not visible on dark backgrounds. 

Changed the fill color to a specific grey color, rather than a low-opacity black.

### Before (Select not visible on black background):
Screen Shot 2021-07-21 at 1 51 13 PM

### After (Select visible on black background):
Screen Shot 2021-07-22 at 1 44 33 PM

## Other information
In a similar fashion, we may want to change the outline variant's colors; it currently is white with a black outline. However, it may make more sense to use something like
- background color: inherit or transparent;
- outline color: inherit text color (for contrast)

### DIFF
--- a/src/components/Select/src/SelectControl.vue
+++ b/src/components/Select/src/SelectControl.vue
@@ -151,7 +151,7 @@ export default {
 	until we get a Theme Context component
 */
 .variant_fill {
-	--color-background: rgba(0, 0, 0, 0.05);
+	--color-background: #f6f7f9;
 	--color-background-focus: rgb(255, 255, 255, 0.95);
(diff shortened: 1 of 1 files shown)


landondurnan (approved):
We're likely going to have to take a different strategy with the borders and when we do that revisit how we're setting default colors on forms more holistically with theming and surface aware components.

For now this fixes the primary issue we're seeing and worth getting in to move forward.

landondurnan:
@ashjtan Could you include a "before" screenshot in the description to provide some more context?

PR #108, opened 2021-08-04
feat(modal): add closeAll()

sohnj877:

Unable to close stacking modals all at the same time

Added modalApi.closeAll()

### DIFF
--- a/src/components/Modal/src/ModalLayer.vue
+++ b/src/components/Modal/src/ModalLayer.vue
@@ -94,6 +94,10 @@ const apiMixin = {
 					vm.currentLayer.state.vnode = undefined; // close modal
 				}
 			},
+
+			async closeAll() {
+				vm.currentLayer.state.vnode = undefined; // close all modals
+			},
 		};
 
 		if (!this.modalApi) {


privatenumber:
Implementation correctness aside, can you describe the problem you're tackling in a new issue?

Looking at this PR, this change seems unwarranted.

mattmewton:
@privatenumber there is a bug with modals where we cannot close a modal that exists on top of another modal. This change allows us to close all modals at the same time, otherwise there is a flicker of the first modal that appears after the second one is closed.

landondurnan:
@sohnj877 Ideally we have a record with some more details and this template can be used to provide additional information about what's being added and why: https://github.com/square/maker/issues/new?assignees=&labels=feature+request%2Cpending+triage&template=FEATURE_REQUEST.yml

src/components/Modal/src/ModalLayer.vue
```diff
@@ -94,6 +94,10 @@ const apiMixin = {
 					vm.currentLayer.state.vnode = undefined; // close modal
 				}
 			},
+
+			async closeAll() {
+				vm.currentLayer.state.vnode = undefined; // close all modals
```
landondurnan, line 99: Wouldn't this only target the modal on the currentLayer? How are we getting into a state where multiple modals are opened on the same layer?

I'm wondering if we should have a check in place so that new modals cannot be opened on the same layer -- the behavior of continuously opening modals on the same layer is likely contributing towards the need to close all of them.

landondurnan: I think we need more context around how multiple modals are being opened at the same time to suggest a better behavior. If one modal is open and another one can open in place of it then I would assume a lot of things can go wrong there. If that's the scenario and this code is being added to fix that problem -- I think we need to change the cause of that problem to begin with, otherwise each engineer is just going to be pulling the rip code to close all modals when that could have a negative experience on customers that are currently using a modal to enter information.

I'm very hesitant to add more code to propagate the idea that multiple modals can even exist. This is a slippery slope of introducing more code, more test, more behaviors that we have to account for rather than pushing for simplification of using modals.

PR #110, opened 2021-08-09
refactor(lab): removed redundant nested root app

privatenumber:

The lab app root was being nested (eg, another app instance was being inserted into the root app). This isn't a big deal, just created an unnecessary double-`div`.

- Removed App.vue nesting
- Made `router-view` a root component
- Made all routes root-routes (since it's no longer nested)

### DIFF
--- a/lab/App.vue
+++ b/lab/App.vue
@@ -1,7 +1,5 @@
 <template>
-	<div>
-		<router-view />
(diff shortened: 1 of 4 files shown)


landondurnan (approved):
I noticed this recently, good clean up.

Issue #112, opened 2021-08-17
MContainer requirement label is too light


ashjtan:
### Bug description

The requirement label in the MContainer is difficult to see when the color chosen is not very dark.
![image](https://user-images.githubusercontent.com/20190589/129774316-fba1275e-9f14-46fc-8b5a-603b906aae6a.png)

See [slack discussion](https://square.slack.com/archives/C01HXVA3G4E/p1629219221002400).

### Reproduction

See any square site that has an item with a required option, e.g. https://www.plantsandfriendstogo.com/product/6AnthuriumClarinervium/1853?cp=true&sa=true&sbp=false&q=false

### Environment

```shell
n/a
```

### Can you contribute a fix?

- [X] I’m interested in opening a pull request for this issue.

ashjtan:
temporarily fixed in #113

PR #113, opened 2021-08-17
fix(container): increase sub/requirement label opacity

ashjtan:
The requirement label was too light. See #112
Before:
image

After:
Screen Shot 2021-08-17 at 12 54 54 PM

bumped the opacity from 0.55 -> 0.7 as a temporary fix before theming is available.

### DIFF
--- a/src/components/Container/src/Container.vue
+++ b/src/components/Container/src/Container.vue
@@ -98,7 +98,7 @@ export default {
 
 <style module="$s">
 .Container {
-	--opacity-sublabel: 0.55;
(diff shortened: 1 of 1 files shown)


mattmewton (approved):

landondurnan (approved):

landondurnan:
Hmm, I missed that the commit was a feat and the PR title was a fix. This triggered a minor release rather than patch. @privatenumber any advice on avoiding the discrepancy in the future (rather than having a keen eye?)

PR #114, opened 2021-08-18
fix(container): make requirement label no-wrap

ashjtan:
On website, the `requirement label` was wrapping on overflow. However, it is typically pretty short, so we want it not to wrap, and to have the the `label` accommodate to that size instead.

image-2021-08-18-13-38-20-317

- ~~Added a `no-wrap` prop that would~~ make the `requirement label` not wrap ~~for its usage in website~~.
- Added `--space` between the `label` and `requirement-label` so that they are not touching

Before:
Screen Shot 2021-08-18 at 1 08 56 PM

After:
Screen Shot 2021-08-18 at 1 08 20 PM

### DIFF
--- a/src/components/Container/README.md
+++ b/src/components/Container/README.md
@@ -73,14 +73,13 @@ export default {
 
 Supports attributes from [`<section>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section).
 
-| Prop     | Type      | Default    | Possible values            | Description                           |
-| -------- | --------- | ---------- | -------------------------- | ------------------------------------- |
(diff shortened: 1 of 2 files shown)


landondurnan (changes requested):
Rather than adding an option prop, I think we can align on making it the default behavior with CSS and avoid adding another option.

mattmewton:
@landondurnan it was my idea to do the prop since I wasn't sure if we wanted this to be default behavior, but it makes sense to me to be the default

landondurnan:
I ran the default behavior past design too and they're cool with it.

landondurnan (approved):

PR #117, opened 2021-08-26
feat(theme): Theme lab testing (with Stitches)

landondurnan:

CleanShot 2021-09-14 at 08 43 33

We need a way to standardize, pass and reuse branded values to all of our components. Our current focus is to make it easier to pass values like color, fonts, and shape which should also provide a foundation for all future theming we're likely to pursue.

This proof-of-concept is leveraging a new framework agnostic dependency called Stitches (https://stitches.dev/). What the core library gives us out of the box:

* a way to pass a Theme UI token spec to our components (https://stitches.dev/docs/tokens), 
* generate reusable CSS properties, 
* standardize our themeable variant props data (https://stitches.dev/docs/variants). 

This library also opens makes it easier to wrap override properties with classes which should help immensely with site and section level styles. In the future we could also benefit from the style utilities that could help standardize any reusable styles and tokens from the theme definition (https://stitches.dev/docs/utils).

The most straightforward example of a theme token is the brand color. We are likely to reuse this color more than any others as a default output. If a user sets their brand color to `#e5d7cc` we'd want to leverage that as the default color on Button and ActionBarButton. Currently this requires developers to inject color data to our components and manually setting that property.

Leveraging the Theme UI spec gives us a blueprint for how to store and pass colors, fonts, shape, and more. So looking at the simplest example of a Theme JSON object.

```
{
  "colors":{
    "primary":"#e5d7cc"
  }
}
```

Leveraging Stitches we can generate a theme from this object that is accessible in JS and also generates reusable CSS properties.

```
const { theme, createTheme } = createStitches({ theme: this.theme, prefix: 'maker' });
```

Where now `theme.colors.primary = "#e5d7cc"` and CSS properties are injected in `:root`. With the prefix defined we now also have a reusable CSS property: `--maker-colors-primary: #e5d7cc`

So we have several goals with the Theme component and all the Maker components that may or may not be children of the Theme.

* When using the Theme component, a developer should have access to the theme data, but by default should be blissfully unaware of the definition for most use cases.
* A component should have CSS fallbacks. The parent `Theme` component is optional and not required for basic component styles.
* Components should leverage default CSS. We should reduce the amount of computed JS necessary for static component rendering.
* We should do our best to maintain the nature of SFC and separation of concerns. The HTML, JS, and CSS for a component should be simple and easy-to-understand. 
* We still need to allow local overrides on components. The theme provides a global definition, the component allows a local override that can be set by the user.

In practice what might this look like? Let's look at an example. A developer wishes to add a Maker button to their view:

```
<m-button>Text</m-button>
```

Our button component should default to a primary button and with no color set we've decided in the CSS to set the color to black.

With the Theme component there's no change to the components API, but we can now leverage CSS property fallbacks for our default styling.

`MButton.vue`
```
<style>
.Button.variant_primary {
	background-color: var(--maker-colors-primary, #000);
}
</style>
```

`App.vue`
```
<m-theme :theme="themeDefinition">
[...]
</m-theme>
```

`Page.vue`
```
<m-button>Text</m-button>
```
Would now render a button with the background color defined in `colors.primary`. With local overrides we just prepend the local inline style from the prop. In our case `<m-button color="#32a852">Green button</m-button>` can still be done with CSS properties:

`MButton.vue`
```
<script>
  computed: {
    inlinePrimaryStyle() {
        return (this.color ? {
            '--inline-background-color': this.color,
            '--inline-color': getHighestContrast(this.color, ['#fff', '#000']),
        } : false);
    },
}
</script>
<style>
.Button.variant_primary {
        color: var(--inline-color, var(--maker-colors-on-primary, #fff));
	background-color: var(--inline-background-color, var(--maker-colors-primary, #000));
}
</style>
```

CleanShot 2021-09-14 at 12 44 27

This pattern is easy-to-understand and reusable leveraging basic CSS conventions. Global and/or inline CSS controls the visual style of our components. Button, ActionBarButton, active form input borders, every component -- can all set CSS defaults at the component level. They can all leverage the standardized CSS properties generated from the Theme UI spec. They can all support inline property definitions. Following this pattern simplifies the code we have dramatically and meets the goals we have for a theming foundation.

When we get into more complex computations for example with contrast checks, that's the appropriate place to leverage JS.

In the proof-of-concept here, I took these foundational parts a bit further: https://square.github.io/maker/lab/theme-lab-stitches/#/Theme/

The theme definition has the basic default colors that would represent the default site background, text color, and primary color. There's also basic definitions for heading and paragraph font-families. Those global defaults which generate these properties:

```
:root, .maker-t-hqpWCq {
    --maker-colors-background: #fff;
    --maker-colors-text: #000;
    --maker-colors-primary: #e5d7cc;
    --maker-colors-on-primary: #000;
    --maker-fonts-heading: "Merriweather", serif;
    --maker-fonts-text: "Open Sans", sans-serif;
    --maker-radii-default: 0px;
 ```

In addition, I wanted to see what challenges we may have in passing and generating section level overrides, what we've been calling color styles / profiles so far. Theme UI / Stitches supports device modes (typically light / dark modes), which allows you to generate in a way alternative themes. So this allows us to leverage those tools to solve a similar problem. In the proof-of-concept you can click any one of the section "cards" to change the global default. All of those cards represent the colors that can be customized in our six profiles.

CleanShot 2021-09-14 at 09 44 38

Is made from this data:
```
{
  "colors":{
    "background":"#fff",
    "text":"#000",
    "primary":"#e5d7cc",
    "surfaces":{
      "profile6":{
        "background":"#464e4e",
        "text":"#fff",
        "heading":"#e5d7cc",
        "primary":"#e5d7cc",
        "secondary":"#fff"
      }
    }
  },
  "fonts":{
    "heading":"\"Merriweather\", serif",
    "text":"\"Open Sans\", sans-serif"
  }
}
```

Which generates:
```
:root, .maker-t-hqpWCq {
    --maker-colors-background: #fff;
    --maker-colors-text: #000;
    --maker-colors-primary: #e5d7cc;
    --maker-colors-on-primary: #000;
    --maker-fonts-heading: "Merriweather", serif;
    --maker-fonts-text: "Open Sans", sans-serif;
    --maker-radii-default: 0px;
}

.profile6 {
    --maker-colors-background: #464e4e;
    --maker-colors-text: #fff;
    --maker-colors-heading: #e5d7cc;
    --maker-colors-primary: #e5d7cc;
    --maker-colors-on-primary: #000;
    --maker-colors-secondary: #fff;
}
```

A reusable class that can then be set at any level either globally or any other element and will then leverage the CSS properties in that class to override the `:root` properties. This is more conceptual future-proofing towards what we need to do with all color in all our components at the global and section level.

Overall this is an early proof-of-concept, but I'm hoping is a promising direction towards standardizing, simplifying, and creating reusable code that meets our goals for a theming foundation. For example, once I had the foundation in for color, I was able to reference the Theme UI spec for fonts, add a couple more definitions, and then leverage the generated properties within seconds. All of this can be done within the Maker components so end-users don't need to know anything about the theme, but the extra benefit here is that the theme object and all the CSS properties are available and reusable to make custom components.

## Other information

I only spent a little time on the contrast checks, but the inline style as seen in the demo works by choosing white or black text based on the color provided. For the theme spec, we could one of two things. One is leverage the Material concept of "primary / on Primary" determine the text color on a background outside of Maker and pass it as a static definition. Or we could have Maker define it at the theme level with some utilities, whitelist a set of colors from the spec and then generate the text contrast automatically. Either way, we need to conditionally check and define it, it just comes down to where we do that.

**Updated**: Added automatic contrast checks and setting on `primary`, `secondary`, and `background` color keys to generate `on-` keys for each color.

### CODE REVIEW

File: src/components/Button/src/Button.vue [hunk 1]
```diff
@@ -231,10 +141,25 @@ export default {
...
+			case 'secondary':
+				return this.inlineSecondaryStyle;
+			default:
+				return this.inlinePrimaryStyle;
+			}
+		},
+		inlinePrimaryStyle() {
+			return (this.color ? {
+				'--inline-background-color': this.color,
+				'--inline-color': getHighestContrast(this.color, ['#fff', '#000']),
```
landondurnan, line 156: this simplifies the text color contrast check when passing the color.

landondurnan:
Closing in favor or a newer version that doesn't conflict with master. New proof-of-concept coming soon.

PR #118, opened 2021-08-27
feat(Choice): Added selectedColor to Choice

mbove-square:

* Customization of `Choice` component

* Allows users to change the `selectedColor` of the `ChoiceOption` component
* Chooses a text color for best contrast depending on the selectedColor

Screen Shot 2021-08-27 at 3 50 16 PM
---
Screen Shot 2021-08-27 at 3 57 06 PM

## Other information

Styleguide updates
https://square.github.io/maker/styleguide/choice-option-color-contrast/#/Choice


src/components/Choice/README.md
```diff
@@ -8,7 +8,18 @@
...
 <template>
 	<div>
-		<m-choice v-model="selected">
+		<label>
+			Color picker
+			<input
+				v-model="color"
+				type="color"
+			>
+		</label><br><br>
```
mattmewton, line 17: Might want to put each of these on a new line

src/components/Choice/src/Choice.vue
```diff
@@ -47,12 +48,19 @@ export default {
...
+			default: '#222',
+			validator: (color) => chroma.valid(color),
+		},
 	},
 
 	data() {
 		return {
 			currentValue: this.selected,
 			isMultiSelect: this.mode === 'multi-select',
+			color: () => this.selectedColor,
```
mattmewton, line 63: Not sure why you need a function here
```suggestion
			color: this.selectedColor,
```

src/components/Choice/src/ChoiceOption.vue
```diff
@@ -4,15 +4,28 @@
...
 	>
 		<slot />
 	</button>
 </template>
 
 <script>
+import chroma from 'chroma-js';
 import key from './key';
 
+function getContrast(chromaBg, targetChromaFg) {
```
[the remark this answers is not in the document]
pretzelhammer: also, since all this logic does is generate some CSS variables I think we should move it into the parent Choice component instead of repeating this work in every ChoiceOption component (since it always evaluates to the same result anyway)

src/components/Choice/src/ChoiceOption.vue
```diff
@@ -32,6 +45,19 @@ export default {
 			}
 			return this.controlState.currentValue === this.value;
 		},
+
+		style() {
```
pretzelhammer, line 49: please move this to the Choice parent component, these styles only have to be calculated once and can be inherited by every ChoiceOption component via CSS inheritance

src/components/Choice/src/Choice.vue
```diff
@@ -47,12 +48,19 @@ export default {
 			default: 'single-select',
 			validator: (modeValue) => ['single-select', 'multi-select'].includes(modeValue),
 		},
+
+		selectedColor: {
+			type: String,
+			default: '#222',
```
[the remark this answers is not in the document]
mbove-square: @mattmewton This was previously the selected background color, and I didn't want to change the default
https://github.com/square/maker/blob/b1c894e13a1631ab8e42dee727978339b696f1c3/src/components/Choice/src/ChoiceOption.vue#L68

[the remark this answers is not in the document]
mattmewton: @pretzelhammer do we currently have a way of organizing these default colors? Wondering if this is something we should be tackling with theming on the way.

src/components/Choice/src/ChoiceOption.vue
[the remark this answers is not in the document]
mbove-square: @mattmewton Just made `getContrast` a util

src/components/Choice/src/Choice.vue
mbove-square: @mattmewton I have modified this so that all computed properties related to selected options have been moved to `getSelectedStyle()` to clarify that this is a function. This was done to make these provided properties reactive.

src/components/Choice/src/ChoiceOption.vue
mbove-square: @pretzelhammer Just moved this logic to `getSelectedStyle` in the Choice parent component

src/components/Choice/src/Choice.vue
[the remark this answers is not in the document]
pretzelhammer: @mattmewton hmmm organizing default theming data is a WIP, but just having a default string for this component in this PR for now is fine 👍

src/components/Choice/src/Choice.vue
```diff
@@ -47,15 +49,43 @@ export default {
...
+			validator: (color) => chroma.valid(color),
+		},
 	},
 
 	data() {
 		return {
 			currentValue: this.selected,
 			isMultiSelect: this.mode === 'multi-select',
+			getSelectedStyle: () => ({
```
pretzelhammer, line 66: i don't think you need to do this... CSS variables will cascade naturally, just set them on the parent and the child ChoiceOption component will inherit them, or is there a particular reason you chose to do it this way?

src/components/Choice/src/ChoiceOption.vue
pretzelhammer: correct me if i'm wrong, but i don't think this computed prop is necessary on the children components, all of the passed styles are already CSS custom properties so you can just set them on the parent and have the children inherit them naturally, i don't think a `getSelectedStyle` method is necessary

mbove-square: @pretzelhammer No you are correct, I was overthinking this. Just made this change

pretzelhammer (approved):

lokesh (approved):

PR #119, opened 2021-09-03
feat: Card component

ashjtan:
Missing a reusable Card component, which is used in [designs](https://www.figma.com/file/mFjFmyQdxkX2Lufcg6Wfs9/%F0%9F%8D%AA-Customer-Accounts-%E2%80%94-Squad-%E3%83%BCHannah%2FJoel-%E2%80%94-Q121?node-id=2337%3A33356) for in-progress projects.

Added the basis for the Card component, including the header `label` and `#actions` slot.
Screen Shot 2021-09-02 at 5 13 27 PM

## TODOs / outstanding questions
- should there be margins around the card?
- TODO: may want to add a footer section to match the designs.
- TODO: figure out the [font weight](https://www.figma.com/file/mFjFmyQdxkX2Lufcg6Wfs9?node-id=2337:33356#104135313) of the Label, Action button, etc.
  - there's a good chance that both should be `500`; however the current NoticeButton is `600`. either this should change, or i will need to make a new CardButton component (that is essentially the same).
  -  should we refactor the `NoticeButton` to be something like `TextButton` for more reusability and clearer role definition?


Ryguy11 (commented):
I would be a proponent of introducing a `text` variation to the `m-button` component. I think this would solve your outstanding questions. Also the `m-button` has built in loading states so those could be utilized as well. @landondurnan @pretzelhammer thoughts on the outstanding questions?

Ryguy11 (commented):
re: adding a footer, I agree that another prop should be added to match the designs but Im not sure the wording footer makes sense here. Maybe like `sublabel`? Even that could be a bit confusing so im not sure

ashjtan:
after talking to Shimaa, i will be closing this PR to make a more simplified card component, which will basically just be a white rectangle with a border and rounded corners, and a slot to drop stuff in it. i will open a separate PR for that. i'll save this branch for now, bc the styling will likely be incorporated into the future ItemCard component. 

additionally, @Ryguy11 we have decided to make a separate Text Button variant, so i will open a future PR for that.

PR #120, opened 2021-09-03
feat: Card component (simplified)

ashjtan:
##Describe the problem this PR addresses
Missing a reusable [Card component](https://www.figma.com/file/Fz10ZinFDOWgXlukmJjXKm/Maker-playground?node-id=3288%3A7588), which is used in designs for in-progress projects.

Created a basic Card component that acts as a container or wrapper, with one default `slot`.

Screen Shot 2021-09-03 at 3 33 22 PM

## Other information
Changes from #119: this is a much more simplified component than the first iteration. It basically is a white rectangle with a border and some inner padding. I initially wanted to have `header/label`, `default`, `footer`, and `actions` slots/props. But after talking with design, the use cases seem to warrant more flexibility. More discussion is needed to iron out the later direction, but this simple iteration is appropriate for the current project use case.


lokesh-squareup (approved):
Do we want to consider clickable cards and including styling for those in the component?

ashjtan:
> Do we want to consider clickable cards and baking in styling for hover and active behaviors into the component?

@lokesh i think this is something we will want to add in the future. however, it isn't needed for the project at hand, so any complex behavior will be delayed til later. i think design also needs to iron out what exactly they want out of the card more than just this simple use case, so there are def changes/additions to come.

pretzelhammer (approved):

PR #121, opened 2021-09-08
feat(Button): `text` variant

ashjtan:
A sort of inline, no-padding button was needed for use in the Card #120 component. See [designs](https://www.figma.com/file/Fz10ZinFDOWgXlukmJjXKm/Maker-playground?node-id=3495%3A20160)

Created a new `text` Button variant that is to be used within wrapping components, such as Notice or Card.

It has **no**:
- fill or border
- padding
- hover, focus, active styles

Screen Shot 2021-09-07 at 5 00 57 PM

Screen Shot 2021-09-07 at 4 57 33 PM

## Other information
- i would like to remove the NoticeButton in another PR (#122), but i will need to check what feature it was created for / if it's being used, as it could potentially be a breaking change.

### DIFF
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -671,18 +671,18 @@ body {
 
 Supports attributes from [`<button>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).
 
-| Prop       | Type      | Default     | Possible values                    | Description                           |
-| ---------- | --------- | ----------- | ---------------------------------- | ------------------------------------- |
-| type       | `string`  | `'button'`  | —                                  | Type of the button                    |
-| size       | `string`  | `'medium'`  | `small`, `medium`, `large`         | Size of the button                    |
-| full-width | `boolean` | `false`     | —                                  | Whether to make the button full-width |
-| color      | `string`  | `'#000'`    | —                                  | Background color of button            |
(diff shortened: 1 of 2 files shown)


src/components/Button/src/Button.vue
```diff
@@ -132,10 +134,28 @@ function ghost(tokens) {
...
+		'--color-contrast-active': transparentHex,
+		'--color-focus': transparentHex,
+	};
+}
+
 const VARIANTS = {
 	primary: fill,
 	secondary: outline,
 	tertiary: ghost,
+	text: transparent,
```
ashjtan, line 158: not confident that `transparent` is the best name since `ghost` is also transparent, but open to suggestions

src/components/Button/README.md
```diff
@@ -25,6 +25,9 @@
 					<th>
 						Tertiary / Ghost
 					</th>
+					<th>
+						Text / Transparent *
```
ashjtan, line 29: taking suggestions if text isn't a good name, e.g. "Inline"?

ashjtan:
After talking to @pretzelhammer, I will be redoing this task and making a separate TextButton component rather than adding a new Button variant. Since many of the Button props do not apply to the TextButton, it will be similar to the existing Button but with many of those props removed.

PR #126, opened 2021-09-10
feat(theme): almost all props on Button & ActionBarButton made themeable

pretzelhammer:

more theming

makes `textColor`, `align`, and `fullWidth` theme-able props in MButton and MAcitonBarButton

## Other information

nope


src/components/Button/src/Button.vue
```diff
@@ -283,10 +294,32 @@ export default {
 			assert.error(['squared', 'rounded', 'pill'].includes(shapeValue), `${shapeValue} resolved from ${shapeValueOrPointer} is not a valid shape value and cannot be used in the shape prop of Button`);
 			return shapeValue;
 		},
+		resolvedFullWidth() {
```
landondurnan, line 297: It's not clear to me why full width or align needs to be global default themeable properties. We don't have a current need for this and I can't imagine a future scenario where this configuration would be necessary.

pretzelhammer: I don't feel strongly either way about `fullWidth` and `align` being themeable or not, but this was originally requested by @laurenhu so maybe she can speak to it

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -183,10 +183,43 @@ export default {
...
 			return shapeValue;
 		},
+		resolvedTextColor() {
+			let textColorValueOrPointer;
+			if (this.textColor) {
+				textColorValueOrPointer = this.textColor;
+			} else {
+				textColorValueOrPointer = this.theme.actionbarbutton.textColor;
+			}
+			const textColorValue = this.theme.resolve(textColorValueOrPointer);
```
lokesh-squareup, line 193: For convenience, what do you think about this update to the the `resolve()` method:
```
/**
 * Resolves passed in val which can be a pointer to a theme prop. If val not set, falls back to themePath. 
 * Which will also be resolved if it is apointer.
 * @param {String} themePath
 * @param {*} [val]
 * @return {*} resolvedVal
 * @example
 * this.theme.resolve('actionbarbutton.textColor', this.textColor);
 */
```

It feels overloaded, but worth considering optimizing as this logic will be recurring across the codebase.

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -183,10 +183,43 @@ export default {
 			assert.error(['squared', 'rounded', 'pill'].includes(shapeValue), `${shapeValue} resolved from ${shapeValueOrPointer} is not a valid shape value and cannot be used in the shape prop of ActionBarButton`);
 			return shapeValue;
 		},
+		resolvedTextColor() {
+			let textColorValueOrPointer;
+			if (this.textColor) {
+				textColorValueOrPointer = this.textColor;
+			} else {
+				textColorValueOrPointer = this.theme.actionbarbutton.textColor;
+			}
```
landondurnan, line 192: why not:
`const textColorValueOrPointer = this.textcolor || this.theme.actionbarbutton.textColor;`

src/components/Button/src/Button.vue
laurenhu: I think it makes documentation for theming confusing if themes can access some props that control the UI for a given component but not others. Practically speaking, it means that developers using Maker components will need to be aware of applying some props while being able to ignore others (`fullWidth` on `ActionBarButton` being an example)

src/components/ActionBar/src/ActionBarButton.vue
pretzelhammer: if you have suggestions for changes on how the `resolve` utility function works please feel welcome to open a PR. I'd prefer to keep this PR focused and scoped to the task of making a few more props on ActionBarButton and Button themeable rather than going on tangentially-related refactor side-quests.

I updated this PR with some JSDoc comments anyway.

src/components/ActionBar/src/ActionBarButton.vue
pretzelhammer: i'll make a follow-up refactor PR to clean up the repetitive computed methods used to resolve themeable props.

src/components/Button/src/Button.vue
landondurnan: There's several components where the defaults are sufficient and any props to change the behavior of something is the exception to the rule meant for limited use cases. Theming is meant to be a global configuration, props on components are local configurations. There's certainly overlap in terms of global and local options, but they do not need to be 1-to-1.  As a consumer no one should have to go to a global configuration on a theme in order to customize the behavior of a components local behavior in a view.  If all buttons have the same shape that's a global configuration that can be set once. If one button is customized to have a different shape that's an exception and set locally. If every prop on every component is a global configuration then it's more confusing where you're supposed to configure things. What's themeable globally is the responsibility of a smaller subset of engineers. Everyone else should be agnostic to it and define their rules as if the theme didn't exist with local overrides as exceptions.

pretzelhammer: > As a consumer no one should have to go to a global configuration on a theme in order to customize the behavior of a components local behavior in a view.

This is still true even if we make `fullWidth` and `align` themeable props.

> If all buttons have the same shape that's a global configuration that can be set once. If one button is customized to have a different shape that's an exception and set locally.

Sounds very simple, intuitive, and straight-forward to me.

> If every prop on every component is a global configuration then it's more confusing where you're supposed to configure things. 

Yet you just explained how simple, intuitive, and straight-forward it was. How is it confusing?

> What's themeable globally is the responsibility of a smaller subset of engineers. 

This is still true even if we make `fullWidth` and `align` themeable props.

> Everyone else should be agnostic to it and define their rules as if the theme didn't exist with local overrides as exceptions.

This is still true even if we make `fullWidth` and `align` themeable props.

> [Our goal] should be to reduce the number of local props are set where in the global config works for the majority of use cases for the entire application.

This is still true even if we make `fullWidth` and `align` themeable props.

laurenhu (approved):

PR #137, opened 2021-09-21
feat(image-uploader): add image uploader component

### LINKED ISSUE #134: Image Upload/Picker
### Feature request

Implement an image upload component as designed here: https://www.figma.com/file/Fz10ZinFDOWgXlukmJjXKm/Maker-playground?node-id=3333%3A15435.

This component should allow users to select multiple images and trigger an upload (if needed). Upload progress should be tracked via a progress bar + a loading spinner.
The number of images selected should be able to be configured to a maximum (or unlimited if unspecified).

### Why?

We currently do not have a standard component to handle uploads.

### Alternatives

_No response_

### Additional context

Demo of working component hooked up to a mock upload function.
Screen Recording 2021-09-17 at 4 10 31 PM

tatems:
This PR adds a new component for handle image selection (and uploading, if required). Closes #134.

The component wraps a single file input, and listens to changes in order to push images.

If given an upload handler, the component will attempt to call it with each image selected, and set the upload progress/status on completion or if an error is ecountered. If a handler is not provided, the file is pushed into the list immediately.

The component also allows setting a maximum number of images and a maximum size (in bytes).

Screen Recording 2021-09-29 at 12 54 33 PM


src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,170 @@
...
+			const url = await new Promise((resolve) => {
+				const reader = new FileReader();
+				reader.onloadend = () => resolve(reader.result);
+				// TODO: Add error logging
+				// eslint-disable-next-line unicorn/prefer-add-event-listener
+				reader.onerror = () => resolve('');
+				reader.readAsDataURL(image.file);
+			});
+
+			Vue.set(image, 'url', url);
```
privatenumber, line 153: Use `this.$set`: https://vuejs.org/v2/api/#vm-set

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,170 @@
...
+			:class="$s.ImageUploaderItem"
+			@selectImages="selectImages"
+		/>
+		<m-image-selection
+			v-for="image of model"
+			:key="image.id"
+			:image="image"
+			:class="$s.ImageUploaderItem"
+			@removeImage="removeImage"
+		/>
```
privatenumber, line 14: You can unprefix the internal components with `m-`

src/components/ImageUploader/src/ImageSelection.vue
```diff
@@ -0,0 +1,174 @@
+<template>
+	<div
+		:class="{ [$s.ImageSelectionContainer]: true, [$s.ImageSelectionContainerError]: hasError }"
```
privatenumber, line 3: Should be multiline but ideally something caught by ESLint.

Will look into it

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -0,0 +1,60 @@
...
+	<div
+		:class="$s.ImagePickerInputContainer"
+		@click="$refs.input.click()"
+	>
+		<input
+			ref="input"
+			:class="$s.ImagePickerInput"
+			type="file"
+			multiple="true"
+			accept="image/jpeg, image/png"
```
privatenumber, line 11: I'm thinking this should be replaced with an `accept` prop that defaults to `image/*`

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+	<div
+		:class="$s.ImagePickerInputContainer"
+		@click="$refs.input.click()"
+	>
+		<input
+			ref="input"
+			:class="$s.ImagePickerInput"
+			type="file"
+			multiple="true"
```
privatenumber, line 10: I think this should be removed and let the user pass it in via `v-bind="$attrs"`. eg. What if the use-case only allows one image?

privatenumber (changes requested):

privatenumber:
Linking Orbit's Upload component for reference:
https://github.com/squareup/orbit-ui/blob/master/src/components/Upload/src/Upload.vue

src/components/ImageUploader/src/ImagePicker.vue
tatems: That, or since we have `max-images`, we set this to be `true` when that is set `>1`, then `false` if set to `1`

src/components/ImageUploader/README.md
```diff
@@ -0,0 +1,157 @@
...
+			reject,
+			uploadProgressHandler,
+			image,
+		}) {
+			let newProgress;
+			const nextStatus = nextStatuses[Math.floor(Math.random() * 3)];
+			setTimeout(() => {
+				if (nextStatus === IMAGE_SELECTOR_STATUSES.PENDING || progress < 75) {
+					const remainingProgress = 100 - progress;
+					newProgress = progress + Math.min(10, Math.random() * remainingProgress);
```
mbove-square, line 108: Not sure if using random here is a great idea, even if it is just for effect

src/components/ImageUploader/README.md
```diff
@@ -0,0 +1,157 @@
...
+		// Pay no mind to any weirdness here, this is to simulate an image taking time to upload
+		uploadImage({
+			progress = 0,
+			resolve,
+			reject,
+			uploadProgressHandler,
+			image,
+		}) {
+			let newProgress;
+			const nextStatus = nextStatuses[Math.floor(Math.random() * 3)];
```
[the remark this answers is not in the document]
tatems: Yeah good idea, I'll split errors out

src/components/ImageUploader/README.md
tatems: I'm not too worried about it in this context. Otherwise, yeah, I'd use something more secure.

src/components/ImageUploader/README.md
```diff
@@ -0,0 +1,157 @@
...
+				this.uploadImage({
+					resolve,
+					reject,
+					uploadProgressHandler,
+					image,
+				});
+			});
+		},
+
+		// Pay no mind to any weirdness here, this is to simulate an image taking time to upload
```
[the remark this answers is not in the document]
tatems: Sure thing

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,180 @@
...
+export default {
+	name: 'MImageUploader',
+
+	components: {
+		ImagePicker,
+		ImageSelection,
+	},
+
+	props: {
+		model: {
```
mattmewton, line 38: Why would we call this `model` and not `images`? Concerned that this could be conflated with `v-model` which has its own mechanisms.

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,180 @@
...
+		},
+		maxImages: {
+			type: Number,
+			default: () => undefined,
+		},
+		maxSize: {
+			type: Number,
+			default: () => undefined,
+		},
+		accept: {
```
mattmewton, line 54: `acceptHeaders` maybe?

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,180 @@
...
+			type: Array,
+			default: () => [],
+		},
+		uploadHandler: {
+			type: Function,
+			default: () => undefined,
+		},
+		maxImages: {
+			type: Number,
+			default: () => undefined,
```
mattmewton, line 48: Why not `null` here? Same question for `maxSize`

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,180 @@
...
+		ImagePicker,
+		ImageSelection,
+	},
+
+	props: {
+		model: {
+			type: Array,
+			default: () => [],
+		},
+		uploadHandler: {
```
mattmewton, line 42: Can we call this `getUploadHandler` to better denote that this is a function?

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,180 @@
...
+			return this.maxImages - this.model.length;
+		},
+
+		canUploadMultiple() {
+			// eslint-disable-next-line no-magic-numbers
+			return !this.maxImages || this.maxImages > 1;
+		},
+	},
+
+	methods: {
```
mattmewton, line 86: Would like to see docblocks on all of these methods

src/components/ImageUploader/src/ImageSelection.vue
```diff
@@ -0,0 +1,177 @@
...
+			/>
+		</div>
+
+		<button
+			v-if="!isUploading"
+			type="button"
+			:class="$s.ImageSelectionRemoveButton"
+			@click="$emit('removeImage', image.id)"
+		>
+			<!-- TODO: Icon -->
```
mattmewton, line 33: Remove?

src/components/ImageUploader/src/ImageSelection.vue
```diff
@@ -0,0 +1,177 @@
...
+			};
+		},
+
+		showProgressBar() {
+			return this.image.progress < MAX_PROGRESS;
+		},
+
+		progressContainerStyle() {
+			return {
+				// eslint-disable-next-line no-magic-numbers
```
mattmewton, line 77: Is there a way we can write this without needing the ESLint disable?

src/components/ImageUploader/src/ImageUploader.vue
tatems: ESLint complained

src/components/ImageUploader/src/ImageUploader.vue
tatems: Hmmm, not sure about that, as the input is the handler function itself. Maybe `uploadHandlerFn` or something similar?

src/components/ImageUploader/src/ImageUploader.vue
tatems: It isn't for headers though, it's for the field `accept` on the `input` element.

src/components/ImageUploader/src/ImageUploader.vue
tatems: Yup good call.

src/components/ImageUploader/src/ImageUploader.vue
mattmewton: Weird. Never heard of that happening.

src/components/ImageUploader/src/ImageUploader.vue
mattmewton: Ah ok

src/components/ImageUploader/src/ImageUploader.vue
mattmewton: Sure, that works

src/components/ImageUploader/src/ImageSelection.vue
```diff
@@ -0,0 +1,174 @@
...
+			type: Object,
+			required: true,
+		},
+	},
+
+	computed: {
+		bgImageStyle() {
+			const url = `url("${this.image.url}")`;
+			const loadingGradient = 'linear-gradient(0deg, rgba(255, 255, 255, 0.5), rgba(255, 255, 255, 0.5))';
+			const loadedGradient = 'linear-gradient(180deg, rgba(0, 0, 0, 0.5) 0%, rgba(0, 0, 0, 0) 50%)';
```
landondurnan, line 67: We have a loading component that could replace this. Should simplify this code a bit.

https://github.com/square/maker/tree/master/src/components/Loading

landondurnan: You can see how it's used in Button for example: https://github.com/square/maker/blob/master/src/components/Button/src/Button.vue#L20-L23

landondurnan (changes requested):

src/components/ImageUploader/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+```
+
+<!-- api-tables:start -->
+## Props
+
+| Prop              | Type     | Default           | Possible values | Description                                                                                                                              |
+| ----------------- | -------- | ----------------- | --------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| images            | `array`  | `[]`              | —               | List of images that have been selected                                                                                                   |
+| upload-handler-fn | `func`   | `() => undefined` | —               | Function called to trigger an upload. Called immediately on image selection, provided max size and max number image constraints are met. |
```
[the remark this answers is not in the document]
tatems: Not sure how we would handle that since I don't want to be too prescriptive about whatever API the consumer is using, as well as error states.

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -0,0 +1,59 @@
...
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	box-sizing: border-box;
+	width: 96px;
+	height: 96px;
+	border: 1px solid rgba(0, 0, 0, 0.15);
+	border-radius: 8px;
+	cursor: pointer;
+	transition: background-color 150ms linear;
```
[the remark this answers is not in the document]
tatems: Good catch, this is a holdover from earlier.

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -0,0 +1,59 @@
...
+</script>
+
+<style module="$s">
+.ImagePickerInputContainer {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	box-sizing: border-box;
+	width: 96px;
+	height: 96px;
```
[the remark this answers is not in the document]
tatems: That's the only size we have the design for right now.

src/components/ImageUploader/src/ImageSelection.vue
tatems: [We do actually use `MLoader` here](https://github.com/square/maker/pull/137/files/e3d8eceac94d7572e35a343cb2c012370dbe44fc#diff-9336009bb13b843a4ee64eefba32b0cd7682e64580c2c810c6e22e51abb1148aR14), this is meant as an overlay for the component while it is being uploaded.

src/components/ImageUploader/README.md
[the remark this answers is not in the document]
tatems: I'm open to suggestions for alternatives. This seemed (to me) that it was the best balance between managing the upload state without too much overhead while also allowing some flexibility in implementation.

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+	<div
```
lokesh, line 2: Couple of suggested edits to make this usable with a keyboard:
```
<div
  :class="$s.ImagePickerInputContainer"
  @click="$refs.input.click()"
  role="button"
  tabindex="0"
  @keydown.space.enter.prevent="$refs.input.click()"
```

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,226 @@
...
+			type: Array,
+			default: () => [],
+		},
+		// Not disabling and breaking the line causes the doc generator to break.
+		/* eslint-disable max-len */
+		/**
+		 * Function called to trigger an upload. Called immediately on image selection, provided max size and max number image constraints are met.
+		*/
+		/* eslint-enable max-len */
+		uploadHandlerFn: {
```
lokesh, line 51: Do you have a use case in mind for this? Could we get by with just emitting an event when the file input fires it's change event?

src/components/ImageUploader/README.md
```diff
@@ -0,0 +1,225 @@
...
+| max-images        | `number` | `() => undefined` | —               | The maximum number of images allowed to be selected.                                                                                     |
+| max-size          | `number` | `() => undefined` | —               | The maximum file size allowed (in bytes)                                                                                                 |
+| accept            | `string` | `() => 'image/*'` | —               | Allowed file types, must be an image type (eg image/jpeg)                                                                                |
+
+
+## Events
+
+| Event | Type | Description |
+| ----- | ---- | ----------- |
+| input | -    | —           |
```
lokesh, line 224: When we have events with names that overlap standard HTML names, we normally prepend them with the component name to make it clear that it is a custom event. So in this case the event could be named `image-uplaoder:input`

src/components/ImageUploader/src/ImageUploader.vue
tatems: Yeah the main thing I was trying to go for was to abstract away managing the progress and pending/complete/error states. If we use just an emitter, the consumer would be responsible for patching all those values and would add a bunch of overhead.
This standardizes that process while still providing enough flexibility to the consumer of this component to interact with their API.

src/components/ImageUploader/README.md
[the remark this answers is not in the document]
mattmewton: What sorts of actions are happening in this function? Perhaps if we break down what is happening piece by piece, we can more easily determine what can be offloaded to Maker internally and what parts should be required on the application layer.

src/components/ImageUploader/src/ImageSelection.vue
mattmewton: Should this be a new Maker component? `MOverlay`?

tatems: Interesting idea. I'll run it by Shimaa. Though I think that could come as a follow up.

src/components/ImageUploader/README.md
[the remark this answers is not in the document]
tatems: @lokesh ccing here
Essentially, all this function should be is one that accepts a) an image, and b) a handler to push progress updates. Here's the basic steps:

- Parent component sets up an async handler function and passes it to the `uploadHandlerFn` prop on the uploader
- A user selects an image (or set of images), [and the uploader will call this function with the image file, plus a function to enable pushing progress updates](https://github.com/square/maker/pull/137/files#diff-439689f507e38af60f86ff36d2a8e5011ebecebd826c93da8a529a58e52b5c45R172)
  - When the promise resolves, [the uploader updates the upload state of the image](https://github.com/square/maker/pull/137/files#diff-439689f507e38af60f86ff36d2a8e5011ebecebd826c93da8a529a58e52b5c45R176)
  - If an error is encountered, [the uploader sets the image state to have an error](https://github.com/square/maker/pull/137/files#diff-439689f507e38af60f86ff36d2a8e5011ebecebd826c93da8a529a58e52b5c45R180)

I had written a better set of docs on this but it got wiped away (I didn't realize these were autopopulated at the time). I can put this into the doc block as well, and into the README itself.

src/components/ImageUploader/src/ImagePicker.vue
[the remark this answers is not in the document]
tatems: Or even just `small`, `medium`, `large` like we do in a lot of places.

src/components/ImageUploader/README.md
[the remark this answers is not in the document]
tatems: I've added some more documentation + some printing out of the internal values to help explain what is happening.

src/components/ImageUploader/src/ImageSelection.vue
landondurnan: This is mostly an optimization, but I think between the existing loading / progress bar components a lot of these styles can go away. The overlay is an interesting additional component, but not that much of an optimization over the additional code to show the loading state of an image. This isn't necessarily a blocker to move forward though. We can either do it now or later.

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,226 @@
...
+			type: Function,
+			default: () => undefined,
+		},
+		/**
+		 * The maximum number of images allowed to be selected.
+		 */
+		maxImages: {
+			type: Number,
+			default: () => undefined,
```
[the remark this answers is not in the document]
tatems: I could have sworn I fixed these already, but yeah will change

src/components/ImageUploader/src/ImageUploader.vue
```diff
@@ -0,0 +1,311 @@
...
+		 * Transforms an image File into a base64 URL. Used to display on the UI.
+		 *
+		 * @param {Object} image formatted image object
+		 */
+		async buildImageURL(image) {
+			const url = await new Promise((resolve) => {
+				const reader = new FileReader();
+				reader.onloadend = () => resolve(reader.result);
+				// eslint-disable-next-line unicorn/prefer-add-event-listener
+				reader.onerror = () => resolve('');
```
landondurnan, line 198: Wouldn't we be missing potential errors here? Why resolve and not reject? https://stackoverflow.com/a/46568146

src/components/ImageUploader/src/ImageUploader.vue
landondurnan: so one thing I've learned is that this function is completely optional, you just miss out on the realtime data (progress) from the server. 

Then just subjective opinion here, but `uploadHandler` is just as meaningful and more concise. But I'm wondering if we want to reference server / service in the name.

landondurnan (commented):
It's still not clear to me how the consumer will manage status changes with the output. I think we should be emitting events for each image status change -- specifically `complete` and `error` events. There is additional information we will need to provide to users, particularly on  error states, and those events can emit, pass data, and allow the consumer to react directly to it more clearly.

src/components/ImageUploader/index.js
```diff
@@ -0,0 +1,4 @@
+export { default as MImageUploader } from './src/ImageUploader.vue';
+export { default as MImagePicker } from './src/ImagePicker.vue';
+export { default as MImageSelection } from './src/ImageSelection.vue';
```
landondurnan, line 3: I don't think we want to export these to be available as public components. They aren't documented or recommended for usage. We'd likely break these down further in the future to be even more atomic, revealing these now isn't necessary.

src/components/ImageUploader/src/ImageSelection.vue
```diff
@@ -0,0 +1,177 @@
...
+				inline
+				:class="$s.ImageSelectionRemoveIcon"
+			/>
+		</button>
+	</div>
+</template>
+
+<script>
+import { MLoading as Loading } from '@square/maker/components/Loading';
+import { MProgressBar as ProgressBar } from '@square/maker/components/ProgressBar';
```
landondurnan, line 40: Why import them with a different name? The `M` prefix on the component will help clarify which components they are when others see them in the template.

tatems: I had misunderstood another change request above, thought they meant all maker components. But I've reverted it.

src/components/ImageUploader/README.md
```diff
@@ -0,0 +1,328 @@
...
+| -------------- | -------- | ----------- | --------------- | ---------------------------------------------------------------------------------------------------------------------------------------- |
+| upload-handler | `func`   | —           | —               | Function called to trigger an upload. Called immediately on image selection, provided max size and max number image constraints are met. |
+| max-images     | `number` | —           | —               | The maximum number of images allowed to be selected.                                                                                     |
+| max-size       | `number` | —           | —               | The maximum file size allowed (in bytes)                                                                                                 |
+| accept         | `string` | `'image/*'` | —               | Allowed file types, must be an image type (eg image/jpeg)                                                                                |
+
+
+## Events
+
+| Event                    | Type | Description |
```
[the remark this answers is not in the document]
tatems: Sure! How do I do that?

lokesh (approved):
LGTM

We discussed in-person some of the trade-offs with this current API, but we are sensitive to the project timeline and noted that there is the option to iterate in the future.

Some of the concerns noted:
- **Readability of the API:** When possible, we try to use primitive types for prop values as this makes the component's features and scope clear. As an example, having distinct props for `color` and `font`, instead of `config` prop that takes an obj that specifies values for both. Not a hard rule, as objects are necessary at times. The concern was around the uploadHandlerFn prop and its arg struct.
- **Modularity** - We default to small, composable components as Maker's dreams are to be long-lived. So it needs to be able to adapt to changing product & design needs. So in this case, separating out the file upload input (and making it non-image specific) and the image thumbnail display/selection was discussed.
- **Adding developer conveniences later** - This component handles much of the plumbing -- reducing the amount of code required to be written by developers. Our normal approach is to release small, flexible components and see usage patterns develop, and then look for opportunities to add conveniences vs adding them ahead of time. 

The Figjam diagram we used to discuss an alternative approach that adopted some of the ideas noted above.

Screen Shot 2021-09-29 at 10 14 46 AM

PR #138, opened 2021-09-22
feat(progress-bar): add progress bar component

### LINKED ISSUE #136: Progress Bar
### Feature request

Add a progress bar as specced here: https://www.figma.com/file/Fz10ZinFDOWgXlukmJjXKm/Maker-playground?node-id=3344%3A13972

Highlights:
- 4 sizes: Large (16px tall); Medium (12px tall); Small (8px tall); Extra Small (4px tall)
- The main bar colour should be configurable (with black as the default)
- The bar should have a light grey background
- The corners should be able to be rounded or sharp based on current site settings

### Why?

We currently have no standard progress bar to use in our projects, leading to differing manual implementations.

### Alternatives

_No response_

### Additional context
This can eventually be used by the Image Selector in #135

tatems:
Closes #136 

This PR introduces a reusable progress bar component. It has 2 styles (square and rounded/pill), is colour configurable, and 4 different sizes.

The `ProgressBar` component has 4 main inputs:
- `shape`: is either square or rounded/pill (sites that have rounded or pill designs will have the same progress bar shape)
- `color`: controls the color of the inner bar
- `size`: xsmall (4px), small (8px), medium (12px), or large (16px)
- `progress`: a 0-100 number that controls the display of the inner bar

The inner progress bar is a full width bar that uses a transform to match the progress input. It does this instead of a width for performance reasons, as animating width is not nearly as performant.

Screen Recording 2021-09-22 at 11 52 26 AM

### DIFF
--- a/src/components/ProgressBar/src/ProgressBar.vue
+++ b/src/components/ProgressBar/src/ProgressBar.vue
@@ -29,21 +29,33 @@ export default {
 	name: 'ProgressBar',
 
 	props: {
+		/**
+		 * Size (height) of the progress bar
+		 */
 		size: {
 			type: String,
 			default: 'medium',
 			validator: (size) => ['xsmall', 'small', 'medium', 'large'].includes(size),
 		},
+		/**
+		 * Shape of the progress bar
+		 */
 		shape: {
 			type: String,
 			default: 'rounded',
 			validator: (shape) => ['squared', 'rounded', 'pill'].includes(shape),
 		},
+		/**
+		 * Color of the progress bar (not background)
+		 */
 		color: {
 			type: String,
 			default: '#000',
 			validator: (color) => chroma.valid(color),
 		},
+		/**
+		 * Progress/width of the bar (0-100)
+		 */
 		progress: {
 			type: Number,
 			default: 0,
@@ -93,10 +105,7 @@ export default {
 }
 
 .ProgressBarContainer.shape_rounded,
-.ProgressBar.shape_rounded {
-	border-radius: 8px;
-}
-
+.ProgressBar.shape_rounded,
 .ProgressBarContainer.shape_pill,
 .ProgressBar.shape_pill {
 	border-radius: 16px;
--- a/src/components/ProgressBar/src/ProgressBar.vue
+++ b/src/components/ProgressBar/src/ProgressBar.vue
@@ -1,0 +1,104 @@
+<template>
+	<div
+		:class="[
+			$s.ProgressBarContainer,
+			$s[`size_${size}`],
+			$s[`shape_${shape}`],
+		]"
+	>
+		<div
+			:class="[
+				$s.ProgressBar,
+				$s[`shape_${shape}`],
+			]"
+			:style="{
+				backgroundColor: color,
+				transform: `translateX(-${progressTranslate}%)`
+			}"
+		/>
+	</div>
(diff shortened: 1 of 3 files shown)


src/components/ProgressBar/src/ProgressBar.vue
```diff
@@ -0,0 +1,113 @@
...
+		:class="[
+			$s.ProgressBarContainer,
+			$s[`size_${size}`],
+			$s[`shape_${shape}`],
+		]"
+	>
+		<div
+			:class="[
+				$s.ProgressBar,
+				$s[`shape_${shape}`],
```
landondurnan, line 12: The shape class is already on the parent and doesn't need to be set again.

src/components/ProgressBar/src/ProgressBar.vue
```diff
@@ -0,0 +1,113 @@
...
+			]"
+			:style="{
+				backgroundColor: color,
+				transform: `translateX(-${progressTranslate}%)`
+			}"
```
landondurnan, line 17: We should move declarations to the CSS and the calculated values to be computed css properties.

`:style="computedStyles"` then a computed method that returns two properties:
```
{
  '--bar-color': this.color,
  '--fill-percent: MAX_PROGRESS - this.progress
}
```

Then in the styles declare the background property and transform you're doing inline and leverage the inline properties instead.

This is a lot cleaner way of keeping the template tight, the computed parts in JS, and the CSS declarations in styles so everything is where it's expected.

src/components/ProgressBar/src/ProgressBar.vue
```diff
@@ -0,0 +1,113 @@
...
+<style module="$s">
+.ProgressBarContainer {
+	width: 100%;
+	overflow: hidden;
+	background-color: #f5f4f4;
+}
+
+.ProgressBar {
+	height: 100%;
+	transition: transform 100ms linear;
```
landondurnan, line 83: I think a similar thing can be done `transition: width ...` and it's less expensive than transforms. See orbit's implementation for ideas: https://github.com/square/orbit/blob/master/src/components/ProgressBar/src/ProgressBar.vue

src/components/ProgressBar/src/ProgressBar.vue
```diff
@@ -0,0 +1,113 @@
...
+
+.ProgressBarContainer.shape_squared,
+.ProgressBar.shape_squared {
+	border-radius: 0;
+}
+
+.ProgressBarContainer.shape_rounded,
+.ProgressBar.shape_rounded,
+.ProgressBarContainer.shape_pill,
+.ProgressBar.shape_pill {
```
landondurnan, line 110: These shape selectors will get simpler by just leveraging the parent class.

landondurnan (changes requested):
Pretty close, some suggestions for improvements.

landondurnan (changes requested):
@tatems One other suggestion. Could you split out the demos, it's a bit intense with the interactions and doesn't add much value. Suggestion: One demo that has the color / progress interaction as you have (but remove all the different sizes). Then have a separate demo for sizes with static props.

tatems:
@landondurnan Shimaa actually liked the demo as-is, so I think I'll keep it for now.

landondurnan (approved):

PR #139, opened 2021-09-22
refactor(theme): drier themeable props

pretzelhammer:

makes code more DRY

makes code more DRY

## Other information

nope


landondurnan (approved):
Tested and works as expected.

I still think we should not have global options for every component property and`fullWidth` and `align` should be removed.

PR #144, opened 2021-09-28
feat(theme): theme level profiles

pretzelhammer:

adds theme-level profiles

adds theme level profiles

## Other information

visit https://square.github.io/maker/styleguide/theme-level-profiles/#/Theme and scroll down to the **Customizing the theme within subsets of the app** section.

### DIFF
--- a/src/components/Theme/README.md
+++ b/src/components/Theme/README.md
@@ -1,6 +1,6 @@
 # Theme
 
-Add the Theme component _anywhere_ top-level to encapsulate all library components. In this example, the Theme component wraps the entire App.
+You can add the Theme component anywhere to encapsulate a group of components you would like to theme. If you would like to theme all the components in your entire app then you would add the Theme component to the root level of your app, like so:
 
 ```vue
 <template>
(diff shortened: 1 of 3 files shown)


laurenhu (approved):

lokesh (approved):
I prefer the conceptual simplicity of m-theme taking in theme objects and applying overrides w/o awareness of profiles, as profiles feel specific to our app at this point in time. But I also see that it's not a big technical cost to add support. We can vet further in product as there might be other pros/cons we uncover, and this code supports both approaches so we can shift as needed. 👍

src/components/Theme/README.md
```diff
@@ -367,13 +369,144 @@ export default {
 </style>
 ```
 
+## Customizing the theme within subsets of the app
+
+Theme components can be nested, and when nested the child Theme can override its parent Theme in two ways:
+1. Providing its own theme data, which will be merged with the theme data provided by the parent or
+2. Selecting a profile from the parent theme, which will be merged with the theme data provided by the parent
+
```
lokesh, line 377: We'll need to think about how to standardize the usage. I can imagine devs adding their own profiles in subsets of the app, which it looks like we are discouraging.

landondurnan:
I don't want to discourage moving forward with this, but wanted to suggest an alternative approach for future consideration. My overall concern with nested themes is opening the doors to increasingly complex ways to override default output, rather than there being a single source of truth for the entire application, any theme override becomes another injection point and another thing to manage.

The primary reason we are doing this is to support setting different colors at section (or potentially page level). This is unique to our use cases of what we've discussed as "color profiles". So this solution meets that need, but it also means any other data could be passed in that we allow "theming". There's no restrictions on just color settings and the data structure would currently support that. I don't think it should. I think profiles are restricted to colors and the data structure should reflect that.  `theme.colors.profiles` rather than `theme.profiles` the distinction isn't extreme, but it is meaningful.

Rather than having sub-themes that could promote endless nesting, I think it would be ideal to follow a pattern that emphasis a application structure where: Theme = Global / App-level / Grandparent. Section = Local container(s) on a page that can leverage the profiles defined in the theme (Parent).  Button, Text, etc = Atomic (children) components that inherit the styles from its parent section or grandparent theme.

We don't need to do any of that right now, but I wanted to express general concern over nested themes with long-term maintenance and usage.

PR #149, opened 2021-09-29
feat(image-uploader): add drag and drop functionality

tatems:
This PR adds the ability to drag files on top of `ImagePicker` and emit added files. It can handle both singular/multiple files, plus directory trees. Only valid image types will be added.

Most of the plumbing comes from the [implementation found in Orbit](https://github.com/squareup/orbit-ui/blob/master/src/components/Upload/src/Upload.vue#L202).
Some small changes have been made to fit the structure of the Maker components + use of `async`/`await`.

Screen Recording 2021-09-29 at 1 57 54 PM


tatems:
Closing in favour of #150

PR #150, opened 2021-09-29
feat(image-uploader): add drag/drop to image picker

tatems:
This PR adds the ability to drag files on top of `ImagePicker` and emit added files. It can handle both singular/multiple files, plus directory trees. Only valid image types will be added.

Most of the plumbing comes from the [implementation found in Orbit](https://github.com/squareup/orbit-ui/blob/master/src/components/Upload/src/Upload.vue#L202).
Some small changes have been made to fit the structure of the Maker components + use of `async`/`await`.

Screen Recording 2021-09-29 at 1 57 54 PM


src/components/ImageUploader/src/ImageSelection.vue
```diff
@@ -145,7 +140,7 @@ export default {
 }
 
 .ImageSelectionContainerError {
-	border: 2px solid var(--color-error);
+	border: 1px solid var(--color-error);
```
Ryguy11, line 143: i found this pretty hard to see on the styleguide, was this a design requested change?

Ryguy11 (approved):
Overall looks good, though its interesting to me that you can drag and drop a directory into the uploader but you cant manually select a directory from clicking.

I would also like to see some updated styleguide docs explaining that directory drag and drop is supported

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -22,16 +35,118 @@
...
+			this.dragged = dragged;
+		},
+
+		async handleDrop(event) {
+			this.setDragged(false);
+
+			if (event.dataTransfer.items) {
+				const fileEntries = await Promise.all(
+					[...event.dataTransfer.items]
+					.map((item) => traverseEntry(item.webkitGetAsEntry && item.webkitGetAsEntry())),
```
pretzelhammer, line 95: so `traverseEntry(item.webkitGetAsEntry && item.webkitGetAsEntry())` only works on webkit browsers? what about non-webkit browsers?

src/components/ImageUploader/src/ImageSelection.vue
pretzelhammer: once you have checked with design please update the PR and/or resolve this comment thread

src/components/ImageUploader/README.md
```diff
@@ -7,8 +7,8 @@ Allows JPEG, PNG, and GIF file formats. HEIC is converted to JPEG by iOS with th
 ## Basic Usage With Upload Handler
 The upload hander (passed in as the `uploadHandlerFn` prop) should accept an object as the first argument. This object will contain the fields:
```
pretzelhammer, line 8: huh? but it's not called `uploadHandlerFn` it's called `uploadHandler`: https://github.com/square/maker/blob/image-uploader-drag-n-drop/src/components/ImageUploader/src/ImageUploader.vue#L47

please update the doc and/or the code to match

src/components/ImageUploader/README.md
```diff
@@ -7,8 +7,8 @@ Allows JPEG, PNG, and GIF file formats. HEIC is converted to JPEG by iOS with th
...
 
-- `image`: A `File` instance for the image to upload
-- `uploadProgressHandler`: A `Function` to push progress updates to the uploader, takes a number as an argument.
+- `imageFile`: A `File` instance for the image to upload
+- `uploadProgress`: A `Function` to push progress updates to the uploader, takes a number as an argument.
```
pretzelhammer, line 11: also huh? there's no `uploadProgress` property, but there is a `setImageProgress` property: https://github.com/square/maker/blob/image-uploader-drag-n-drop/src/components/ImageUploader/src/ImageUploader.vue#L203

please update the docs and/or code so they match

also, changing the property names since the last release is a breaking change and requires a library major version bump, if you intend to stick with these name changes please communicate that these changes are breaking by:
1) renaming the PR title from `feat(image-uploader): add drag/drop to image picker` to `feat(image-uploader)!: add drag/drop to image picker`. please notice the added `!` which communicates to Maker to make a new major version release when the PR is merged in.
2) add a **Migration Guide from 6.x to 7.x** section to your PR description which gives people instructions on how to migrate their code to be compatible with the new major release, if you want some examples of prior written migration guides you can look at these PRs:
- https://github.com/square/maker/pull/148
- https://github.com/square/maker/pull/35

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -22,16 +35,118 @@
 
 <script>
 import PlusIcon from '@square/maker-icons/Plus';
+import ArrowIcon from '@square/maker-icons/ArrowUp';
+import {
+	isEqual, uniqWith, escapeRegExp, partition,
+} from 'lodash';
+
+const getEntryFile = (entry) => new Promise((resolve) => entry.file(resolve));
```
pretzelhammer, line 43: does the `reject` case not need to be handled here?

src/components/ImageUploader/src/ImagePicker.vue
```diff
@@ -22,16 +35,118 @@
...
+	}
+	return getAllEntries(reader, aggregated.concat(entries));
+});
+function traverseEntry(entry) {
+	if (entry) {
+		if (entry.isFile) {
+			return getEntryFile(entry);
+		}
+		if (entry.isDirectory) {
+			return getAllEntries(entry.createReader()).then(
```
pretzelhammer, line 58: instead of using `then` here you can convert this function into an async function and use await

src/components/ImageUploader/README.md
tatems: The current version has the proper names, just the documentation was not updated. I don't think we need make it a breaking change.

src/components/ImageUploader/src/ImageSelection.vue
tatems: Got confirmation, we're g2g

PR #152, opened 2021-09-30
chore: add issue template for new components

landondurnan:
As more engineers are contributing new components to Maker I thought it would be helpful for communication and collaboration to ensure we're getting more information about a component _before_ engineering work starts.

### DIFF
--- a/.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
+++ b/.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
@@ -1,0 +1,53 @@
+name: New Component
+description: Provide details to help define a new component for the system
+labels: [feature request, pending triage]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        **⚠️ This is a public repository. Please be aware of sensitive information and prefer referencing internal links.**
+
+        Please leverage this template to provide as much details about the visual and functional requirements for this new component. Doing so will help use communicate and collborate on a solution effectively.
+
+  - type: text
+    attributes:
+      label: Design File
+      description: Provide a Figma link as a reference
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Description
+      description: |
+        Provide as much details as possible about the visual and functional requirements.
+      placeholder: |
+        <!--
+          Consider UI/UX like affordance, user interaction, and different states
+        -->
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Research
+      description: |
+        How similar is this new component to any existing component?
+      placeholder: |
+        <!--
+          Link to any existing components from Maker or other design systems (Material, Apple, Carbon, etc.) and describe their similarities
+        -->
+      validations:
+        required: true
+  - type: textarea
+    attributes:
+      label: API
+      description: |
+        Outline considerations around props, slots, events, and other important aspects of how developers will use this component.
+  - type: textarea
+    attributes:
+      label: Theme support
+      description: |
+        How will the Theme impact this component? Describe how color, font, shape and other design levers will be used.
+      placeholder: |
+        <!--
+          Considerations around how your component will work in scenarios with different colored backgrounds and contrast is important
+        -->


privatenumber:
As with any change, it's important to outline the problems it's addressing so others can provide better feedback (hence the PR template -- please don't delete it next time).

Right now the PR just describes a hope or an intention. But PRs should address something more concrete.

That said, I'm wondering what specific problem/frustration you're addressing that the current Feature request template doesn't. And if there are any changes that should be made to the Feature request template instead (eg. delete it, rename it to "Component feature request", etc.)

landondurnan:
Updated description with more detail

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        **⚠️ This is a public repository. Please be aware of sensitive information and prefer referencing internal links.**
+
+        Please leverage this template to provide as much details about the visual and functional requirements for this new component. Doing so will help use communicate and collborate on a solution effectively.
+
+  - type: text
+    attributes:
+      label: Design File
```
privatenumber, line 14:
```suggestion
      label: Figma link
```

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
+name: New Component
```
privatenumber, line 1:
```suggestion
name: Request a new component
```

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+description: Provide details to help define a new component for the system
+labels: [feature request, pending triage]
```
privatenumber, line 3:
```suggestion
labels: [new component request, pending triage]
```

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+    attributes:
+      label: Design File
+      description: Provide a Figma link as a reference
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Description
+      description: |
+        Provide as much details as possible about the visual and functional requirements.
```
privatenumber, line 22:
```suggestion
        Provide as much detail as possible about the visual and functional requirements.
```

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+      placeholder: |
+        <!--
+          Consider UI/UX like affordance, user interaction, and different states
+        -->
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Research
```
privatenumber, line 31:
```suggestion
      label: Prior art
```

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+      description: |
+        How similar is this new component to any existing component?
```
privatenumber, line 33: This sounds like you're asking for similarities between other components in Maker.

And specifically, I don't think we care about "how similar" it is to other 3rd party components, but we just want references.

```suggestion
        Are there similar components in 3rd party component libraries? 
```

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+      placeholder: |
+        <!--
+          Link to any existing components from Maker or other design systems (Material, Apple, Carbon, etc.) and describe their similarities
```
privatenumber, line 36: Add links to common design systems to make it easier to find.

Some contributors may not know the common design systems, or what to search, if they're new to UI.

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
```diff
@@ -0,0 +1,53 @@
...
+  - type: textarea
+    attributes:
+      label: API
+      description: |
+        Outline considerations around props, slots, events, and other important aspects of how developers will use this component.
+  - type: textarea
+    attributes:
+      label: Theme support
+      description: |
+        How will the Theme impact this component? Describe how color, font, shape and other design levers will be used.
```
privatenumber, line 49:
```suggestion
        How will theming impact this component? Describe how color, font, shape and other design levers will be used.
```

privatenumber (changes requested):
- A lot of these questions (UI states, prior art, theming) seem like they should be answered by the component designers, rather than the engineer tasked with implementing it. Since we're using issue templates now, it shouldn't be hard for non-technical people to fill out. I wonder if it's worth making a "Component request" issue template intended for designers (no API discussion).

- I think the "Feature request" issue template will be confusing with this change, and should be edited to clarify "Component feature request".

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
landondurnan: I was attempting to do both. For example, there's a new component getting queued up for "pin codes" that are very similar to visual and functional requirements of an input or password component. Referencing our existing input component would likely be a good start.  However, some of the functional requirements are still new and 3rd party references are relevant too.

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
landondurnan: That's fair, but a list of links to a few design systems starts to feel prescriptive and non-exhaustive. There's dozens of great systems out there and as placeholder text you can't click on any links. I agree it's worth potentially a bit more guidance, but I don't have a better alternative idea.

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
landondurnan: Changed to be a bit more specific.

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
privatenumber: I think this section sounds prescriptive because it comes off as a task and doesn't explain the goal (eg. why do we care about other design systems?).

I'm actually not sure what your intention is here, because I think it would vary dramatically depending on whether this template should be filled out by an engineer tasked with implementation, or the component designer.

Making the intention and end-goal clear would give the contributor more creative freedom.

----

FYI As an engineer that implemented many components, I didn't often get value from learning what other UI libraries are doing. I might even go as far as saying it's distracting from listening to our own users and use-cases.

privatenumber:
Chatted with @landondurnan on Slack and I think the next move is to take a step back and publicly discuss the root problem on Issues/Discussion. That way we can get feedback from contributors on what works best for them.

---

I think the main problem with this template is it's not clear who it's for (engineer vs designer). It has questions for both designers and engineers, and that makes it difficult for one person to answer.


If we want to add a new Issue template for components, let's start by establishing a process that involves PM/designer + engineers so we can split up expected deliverables by target audience.

.github/ISSUE_TEMPLATE/NEW_COMPONENT.yml
landondurnan: Several recent components we've referenced Orbit implementations since we have similar problems and solutions. 

From an engineering standpoint, I do think it provides value to research a little bit around the conventions of reusable components in other systems as it helps compare and/or contrast implementation details that often get overlooked until implementation starts. 

Knowing up front what visual / functional requirements may be needed or missed in the original design provides everyone an opportunity to align and iterate on those requirements. Having other real-world examples can speed up that process and make communication easier.

If any new component is so unique that there's no relevant comparison then the answer to this question can simply be "None"

PR #153, opened 2021-10-01
Add FluidCarousel component

mbove-square:

### DIFF
--- a/src/components/FluidCarousel/README.md
+++ b/src/components/FluidCarousel/README.md
@@ -1,0 +1,204 @@
+# FluidCarousel
+
+## Example
+```vue
+<template>
+	<div class="container">
+		<m-fluid-carousel
+			is-draggable
+		>
+			<div
+				v-for="i in 25"
+				:key="i"
+				class="carousel-item"
+			>
+				{{ i }}
+			</div>
+		</m-fluid-carousel>
+	</div>
+</template>
+
+<script>
+import { MFluidCarousel, CAROUSEL_BACK_EVENT } from '@square/maker/components/FluidCarousel';
+
+export default {
+	components: {
+		MFluidCarousel,
+	},
+	mounted() {
+		console.log(CAROUSEL_BACK_EVENT);
+	},
+};
+</script>
+<style scoped>
+.container {
+	max-width: 600px;
+}
+
+.carousel-controls {
+	display: flex;
+	justify-content: flex-end;
+	margin-bottom: 12px;
+}
+
+.carousel-item {
+	width: 180px;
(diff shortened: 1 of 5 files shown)


mattmewton:
What is the reasoning behind moving this to Maker now? Just want to make sure that we are aligned with the site design folks on components that we deem to be reusable.

src/components/FluidCarousel/README.md
```diff
@@ -0,0 +1,299 @@
...
+
+
+## FluidCarousel Events
+
+| Event                      | Type      | Description                                             |
+| -------------------------- | --------- | ------------------------------------------------------- |
+| overflowing                | `boolean` | Event emitted when carousel content overflows container |
+| carousel:mounted           | -         | Event emitted during mounted lifecycle hook             |
+| scroll:end                 | -         | Emitted when the scroll position is near the end        |
+| carousel:at-first-position | `boolean` | Emitted when carousel moves                             |
```
ashjtan, line 290: as a general question / to open discussion, have we considered hiding the left/right controls when at the first/last position accordingly by default? it's a behavior we tend to add when using `FluidCarouselControls`, n i can't rly imagine a case in which we would be strongly opposed to that behavior.

landondurnan (changes requested):
@mbove-square It's really hard to evaluate the visual and functional requirements of this component without a design reference. Can you add one?

I'm also unclear about the name `FluidCarousel` since internally we've been discussing a `Carousel` component. Is there going to be another component called `Carousel`


Many of the props you've added seem like they could be either computed or automatically enabled. For example, item-count and pixels-to-scroll are both calculated values that should probably be internal to the component. Another alternative worth considering is a child component like carousel-item which could simplify some things. 

It's not clear why the header and footer slots are needed. You can already add any content you wish above or below a component.

It's not clear based on the demos the benefit of the emitted events. I'm having a hard time understanding how they might be used.

I know we have a few different use cases with arrows being visible, not visible, and in different locations. Depending on the situations, that might make sense to have a child component for the controls, but I think we need to clarify those scenarios as the interaction, positioning, and designs may not be consistent or be misinterpreted -- again this is where having more design references can be helpful.

Generally, the amount of props, slots, events, and child components really need to be scrutinized. The larger those options grow it may be a warning sign that this component isn't atomic enough and might need to be broken down into smaller components. We can get into sticky situations where one prop conflicts with another prop and the code gets very complex to manage.

Let's get another look at the designs and prioritize the scenarios in which this component would provide the most value. Once we understand the most basic version of this component we can look at edge cases and see how we can add on to this or explore some other alternatives.

PR #155, opened 2021-10-04
feat(theme): add color modes/contrast colors for components

laurenhu:
This PR takes a first pass at implementing the color modes proposed by design to allow components to render over any background color.

Notes:
- The [original test](https://github.com/square/maker/pull/154) followed the design spec and this used leonardo to generate 9 color contrast levels. There was feedback that leonardo did not product good contrast colors for mid-range colors like (e.g. #3499fe or #ff6732). Additionally, 9 contrast colors seemed excessive for the initial design changes, which only requires 4.

- This version uses a simpler formula to generate colors as a starting point. The `contrastColors` util function currently uses chroma's `mix` function in the `'lab'` colorspace to produce contrast colors but can be adjusted to use any formula we want.

- This version generates 4 contrast colors, that correspond the the original design's levels 100, 300, 800, 900 - the colors used in the proposed design. These can be renamed but for now I kept the level names to match the designs.

- The components updated are all showcased in the ColorModes lab experiment

Screen Shot 2021-10-04 at 12 17 15 PM
Screen Shot 2021-10-04 at 12 17 28 PM
Screen Shot 2021-10-04 at 12 17 52 PM


src/components/Heading/src/Heading.vue
```diff
@@ -119,6 +119,7 @@ export default {
 
 .Heading {
 	margin: 0;
+	color: var(--color-900);
```
pretzelhammer, line 122: Hmmm, I'm wondering what this might look like in the near future one we add color props to MHeading and allow local overrides... maybe it'll be `color: var(--text-color, var(--color-900, #000))` where `--text-color` is some CSS var computed locally... also, do we need a default here, e.g. `var(--color-900, #000)`?

src/components/Stepper/src/Stepper.vue
```diff
@@ -95,6 +103,12 @@ export default {
 		minVal() {
 			return Number.parseInt(this.min, 10);
 		},
+		resolvedColor() {
+			return this.color || this.theme.colors.contrast['color-100'] || '#cccccc';
+		},
+		resolvedTextColor() {
+			return this.textColor || this.theme.colors.contrast['color-800'] || '#000000';
+		},
```
pretzelhammer, line 111: if you'd like to follow the convention established in some of the other components, in order to resolve a potential theme value you would use the `resolveThemeableProps` utility function, button example: https://github.com/square/maker/blob/beta/src/components/Button/src/Button.vue#L231

And you would add a `stepper` property to the default theme which would look something like this:
```js
// in default theme object
stepper: {
  color: '@colors.contrast.color-100',
  textColor: '@colors.contrast.color-800',
}
```
button example: https://github.com/square/maker/blob/beta/src/components/Theme/src/default-theme.js#L13-L21

and then add in the defaults for those colors in the theme itself as well:
```js
// in default theme object
colors: {
  color-100: '#cccccc',
  color-800: '#000000',
}
```

src/components/Text/src/Text.vue
```diff
@@ -81,6 +81,7 @@ export default {
 }
 
 .Paragraph {
+	color: var(--color-800);
```
pretzelhammer, line 84: Does this need a fallback? like `var(--color-800, inherit)`?

src/components/Heading/src/Heading.vue
landondurnan: This might be confusing due to the fact that this is a generated value, but we need to consider that our existing profiles allow explicit customization of the heading color that's not tied to the background color.

src/components/Text/src/Text.vue
landondurnan: This will have the same issue as heading where a profile can have its unique explicit value for the text color.

landondurnan (commented):
I think we need to really consider the naming conventions, the scale with this limited set of generated colors is subjectively causing more confusion than clarity. I think mapping out the colors and usage could provide us more insight into what more meaningful names might be. Noted in internal conversations, but we should be looking at the https://theme-ui.com/theme-spec/#color for some assistance in guidelines and leverage similar patterns to define color usage.

I think there's also a disconnect right now in the fact that the profiles being passed outside of Maker already have explicit definitions for heading, text, primary, secondary buttons and is not using any of those values.

landondurnan (changes requested):
I created a table that outlines all the changes in each component including the component variables, the generated colors used, and the current default (or fallback) where relevant. There are inconsistencies with the colors and variable names, but primarily I wanted to identify the patterns established and see if we can come up with more meaningful names instead of the disparate color scale.

|                           | Variable / property            | Generated Color     | Current Default / Fallback |
| ------------------------- | ------------------------------ | ------------------- | -------------------------- |
| Button                    | \--focus-border                | \--color-300        | white                      |
| CheckboxControl           | \--color-foreground            | \--color-100        | white, 95%                 |
|                           | \--color-border                | \--color-300        | black, 90%                 |
|                           | \--color-border-active         | \--color-800        | black, 90%                 |
|                           | \--color-backgorund            | \--color-800        | black, 90%                 |
| Divider                   | background-color               | \--color-300        | black, 8%                  |
| Heading                   | color                          | \--color-900        |                            |
| InputControl - Fill       | \--color-background            | \--color-100        | #f6f7f9                    |
|                           | \--color-placeholder           | \--color-800        | black, 55%                 |
|                           | \--color-foreground            | \--color-800        | #7f7f7f, 90%               |
|                           | \--color-disabled              | \--color-300        | black, 30%                 |
|                           | \--color-background-disabled   | \--color-800        | black, 5%                  |
|                           | \--color-error                 | \--color-error      | #ce3217                    |
|                           | \--border-color-hover          | \--color-300        | #222                       |
| InputControl - Outline    | Styles removed -- need to test |                     |                            |
| Notice                    | \--color                       | \--color-800        | #181818                    |
|                           | \--color-icon                  | \--color-800        | #a9a9a9                    |
|                           | \--color-bg                    | \--color-100        | #f1f1f1                    |
| Notice > .Icon            | stroke                         | \--color-bg         | white                      |
| RadioControl              | \--color-border                | \--color-300        | black, 30%                 |
|                           | \--color-fill                  | \--color-800        | black, 90%                 |
|                           | \--color-focus                 | \--color-800        | black, 90%                 |
|                           | \--color-error                 | \--color-error      | #ce3217                    |
| SelectControl -- Fill     | \--color-background            | \--color-100        | #f6f7f9                    |
|                           | \--color-foreground            | \--color-800        | #020101, 90%               |
|                           | \--color-disabled              | \--color-300        | black, 30%                 |
|                           | \--color-background-disabled   | \--color-800        | black, 5%                  |
|                           | \--color-border-active         | \--color-300        | #222                       |
|                           | \--color-error                 | \--color-error      | #ce3217                    |
| SelectControl - Outline   | Styles removed -- need to test |                     |                            |
| SelectControl > .Icon     | color                          | \--color-foreground |                            |
| Stepper                   | computed(resolvedColor)        | \--color-100        | #ccc                       |
|                           | computed(resolvedTextColor)    | \--color-800        | #000                       |
| Text                      | color                          | \--color-800        | inherit                    |
| TextareaControl - Fill    | \--color-background            | \--color-100        | #f6f7f9                    |
|                           | \--color-placeholder           | \--color-800        | black, 55%                 |
|                           | \--color-foreground            | \--color-800        | #7f7f7f, 90%               |
|                           | \--color-disabled              | \--color-300        | black, 30%                 |
|                           | \--color-background-disabled   | \--color-800        | black, 5%                  |
|                           | \--color-error                 | \--color-error      | #ce3217                    |
|                           | \--color-border-active         | \--color-300        | #222                       |
| TextareaControl - Outline | Styles removed -- need to test |                     |                            |
| TextareaControl > .Icon   | color                          | \--color-foreground |                            |

Summary of usage looks like this:
- --color-100 = Almost always background color of component
- --color-300 = Borders and Disabled
- --color-800 = Leveraged primarily as "text" color (foreground)
- --color-900 = Heading color
- --color-error = always red, doesn't need to be redefined in each component

My recommendation is we do not use a partial numeric scale. Rather we should define meaningful names that can be shared across components. We also need to account for the existing profile definitions already being passed.  The current Heading and Text component should leverage the defaults in the profile.

Every profile already has the following definitions:

- background
- heading
- text
- primary
- secondary

We fill the gaps with the new colors leveraged for output we can use more meaningful names taking inspiration from https://theme-ui.com/theme-spec/ 

- accent = A contrast color for emphasizing UI
- muted = A faint color for backgrounds, borders, and accents that do not require high contrast with the background color"
- borders = component borders and divider

.postcssrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
 			preserve: false,
 			features: {
 				'custom-properties': false,
+				'focus-within-pseudo-class': false,
```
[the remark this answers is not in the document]
laurenhu: [Postcss focus within](https://github.com/csstools/postcss-focus-within) needs to be paired with the [focus-within polyfill](https://github.com/jsxtools/focus-within) in order to work. Otherwise it just swaps the selector for an attribute that cannot be targetted. Considering that [:focus-within](https://caniuse.com/css-focus-within) is supported by all browsers except for IE, it made more sense to just turn off the postcss feature but correct me if I'm wrong and we're supporting IE 11.

Issue #172, opened 2021-10-28
Popover Component


tatems:
### Figma link

https://www.figma.com/file/Fz10ZinFDOWgXlukmJjXKm/Maker-playground?node-id=4178%3A63689

### Description

The goal here is to build out a set of components to provide a way to display more contextual information or actions that can be triggered by an action from another component. Inspiration for these components comes from the [Orbit UI Popover](https://github.com/squareup/orbit-ui/tree/master/src/components/Popover).

The overall functionality of this component should be similar to the Orbit counterpart. The main requirements are:

- Be triggered by an action that could be inside or outside the `o-popover` component.
- The popover should 'tether' itself to the element that triggers it (most of the time)
  - Unlike Orbit, this component should be able to be triggered from elsewhere in the DOM, and be able to be tethered to another element entirely
- The tethered component should be able to re-position itself to stay in view
- The popover content should conform to some height/width/scroll behaviours (detailed in the Figma)

### API

In most cases, this will be the most common implementation pattern for the popover:
```html
<template>
  <m-popover>
    <!-- this action slot will be the element the popover will tether to -->
    <template #action="popover">
      <m-button @click="popover.toggle()">
        Popover Toggle
      </m-button>
    </template>

    <!-- Option #1 automatically put this content in the container -->
    <template #content>
      Some content here
    </template>

    <!-- Option #2 For more customizability, have consumers put the container in -->
    <template #content>
      <m-popover-container>
        Some content here
      </m-popover-container>
    </template>
  </m-popover>
</template>
```

However, it will be necessary to expose a way to trigger the popover from _outside_ the popover component context:

```html
<template>
  <div>
    <m-button @click="togglePopover()">
      Popover Toggle
    </m-button>
  </div>

  <div ref="popoverTether">
    Tether to me!
  </div>

  <m-popover
    ref="popover"
    :tether-element="popoverTether"
  >
    <template #content>
      Some content here
    </template>
  </m-popover>
</template>

<script>
export default {
  ...
  mounted() {
    this.someEventBus.$on('someTrigger', () => this.$refs.popover.open());
  },

  methods: {
    togglePopover() {
      this.$refs.popover.toggle();
    },
  },
};
</script>
```

### Theme support

At this point in time, we will be supporting light and dark themes.
![image](https://user-images.githubusercontent.com/3402466/139260774-b8e9b62f-1e8a-4ac2-9b1d-f2d6f75d2a3d.png)

tatems:
Questions/Feedback:
[...]

PR #176, opened 2021-11-03
feat(popover): add popover component

tatems:
This PR adds a set of components to support popovers. These popovers will allow consumers to provide additional context or options/actions to users without taking them completely out of context via a modal/dialog.

image

There are 5 major parts to this system:

1. `popoverApi` mixin: Creates a popover layer provider. This stores the current popover and other contextual information and handles outside clicks/blurs to close the modal.
2. `PopoverLayer`: This tracks the currently open popover, and also contains the target `portal`.
3. `PopoverInstance`: This manages the `Popper.js` instance, and contains a `Portal` for the popover content.
4. `Popover`: This is the component that accepts the content for the popover. It accepts `tether` (what the popover attaches to) and `content` (the stuff inside the popover) slots. It uses `Portal` to push content of the `PopoverInstance` to the `PopoverLayer` when open.
5. `PopoverContainer`: This component provide a default style for the popover content container, and should be used inside the `content` slot of the `Popover` component.

## Other information
While this code was initially built as a fairly 1-to-1 copy of Orbit's implementation, there was a bit of a journey to get to the current state:

- Orbit's heavy use of VNodes meant that reactivity was broken for content inside the popover. That meant that content would not be changed once the popover was opened. This was necessary to fix for upcoming `Menu` and `MenuSelect` support.
- As an attempt to fix this, we tried `portal-vue`, which worked better for reactive props/values, but broke `provide`/`inject`. Again, this was needed for `MenuSelect`.
- We settled on `vue-simple-portal`, which worked for both reactivity and `provide`/`inject`, and is the basis for Vue 3's `teleport` feature. However a couple key changes were needed to support this:
  - The `PopoverLayer` is no longer responsible for dynamically rendering popover instances. Instead, it tracks the currently open popover, and contains a static portal target for any popover opened in the layer.
  - The `Popover` controls the creation/destruction of it's own popover. It determines this state from the `popoverApi` provider (if the currently open `id` matches it's own `id`, it is open).


src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,237 @@
...
+			},
+		},
+	],
+});
+
+export default {
+	name: 'Popover',
+
+	inject: {
+		popoverAPI: {
```
ashjtan, line 56:
```suggestion
		popoverApi: {
```

minor, but if u can, i would change the casing to match existing `dialogApi`, `modalApi`, etc.

landondurnan:
@tatems Not sure if this is intended to be a draft or you're looking for feedback. If so please update the PR descriptions and reference the original issue: https://github.com/square/maker/issues/172

src/components/Modal/src/ModalLayer.vue
```diff
@@ -86,14 +88,20 @@ const apiMixin = {
...
 						if (!(await this.state.options.beforeCloseHook())) {
 							return; // cancel
 						}
 					}
 
-					vm.currentLayer.state.vnode = undefined; // close modal
+					// Close the open popover (if present) and then close the modal in the next tick.
+					// Closing at the same time will result in the popover content becoming inline and
+					// causes a weird content shift as the modal fades away.
+					vm.popoverApi.closePopover();
```
landondurnan, line 101: This is likely necessary because we don't have a sophisticated layer manager, but want to call this out as potential future code-smell as we're creating internal dependencies between unrelated components. The popover may not be used ever, but now the modal needs to control closing something that might not exist.

tatems:
@landondurnan my bad, I opened as a draft awhile ago and forgot to go back and fill it in. This is ready for review.

tatems:
@landondurnan converting back to a draft. I found an issue where `provide`/`inject` is broken when using `vue-portal`, and that needs to be addressed for `Menu` to work how we want.

src/components/Popover/src/PopoverBubble.vue
```diff
@@ -0,0 +1,55 @@
...
+				'--popover-color': this.color,
+				'--popover-bg-color': this.bgColor,
+			};
+		},
+	},
+};
+</script>
+
+<style module="$s">
+.PopoverBubble {
```
landondurnan, line 48: I think we need to account for a default output that includes the desired padding in the container. Right now, the out-of-box solution requires additional styling. Rather than enforcing additional styling, let's default the padding and add support to override it. The default should have 24px padding:

CleanShot 2021-11-09 at 15 18 56

But in scenarios like Menu you will need a prop to override that padding.

package.json
```diff
@@ -122,5 +122,11 @@
   },
   "release": {
     "preset": "conventionalcommits"
+  },
+  "dependencies": {
+    "@linusborg/vue-simple-portal": "^0.1.5",
+    "@popperjs/core": "^2.10.2",
+    "portal-vue": "^2.1.7",
+    "uuid": "^8.3.2"
```
landondurnan, line 130: Which of these dependencies are truly needed? Vue-simple-portal is a smaller repro of portal-vue. uuid is for generating unique IDs, but I don't think we need it.

src/components/Modal/src/ModalLayer.vue
```diff
@@ -2,16 +2,16 @@
 	<div :class="$s.Layer">
 		<m-transition-fade-in>
 			<div
-				v-if="currentLayer.state.vnode"
+				v-if="currentModalLayer.state.vnode"
```
landondurnan, line 5: Why the name change?

We may need to consider that more than modals can exist at this layer stack in the future and currentLayer may be more descriptive.

src/components/Popover/index.js
```diff
@@ -0,0 +1,4 @@
+export { default as MPopoverLayer } from './src/PopoverLayer.vue';
+export { default as MPopover } from './src/Popover.vue';
+export { default as MPopoverBubble } from './src/PopoverBubble.vue';
```
landondurnan, line 3: I don't think we should make this public. Rather as an internal only component the public component would still be "Popover" and we'd pass any relevant props from the single-use popover component down into the internal child.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
+<template>
+	<div style="display: contents;">
+		<slot
```
landondurnan, line 3: The tether API feels like an additional layer of configuration that isn't needed right now. By restricting the current use-cases to specific elements in the same view we can reduce the complexity.

Are there current use cases we need to consider to wrap other elements?

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+				'right', 'right-start', 'right-end',
+				'bottom', 'bottom-start', 'bottom-end',
+				'left', 'left-start', 'left-end',
+			].includes(placement),
+		},
+
+		/**
+		 * Distance from tether element
+		 */
+		distanceOffset: {
```
landondurnan, line 110: This feels extraneous right now and can likely be removed. There isn't another scenario where we're configuring this output differently.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+			type: Number,
+			default: 8,
+		},
+
+		/**
+		 * Offset from base position (Y for left/right placement, X for top/bottom placement)
+		 */
+		skiddingOffset: {
```
landondurnan, line 118: Similar to the distanceOffset -- I don't think there are any scenarios here that we need to account for an alternative to the default.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+			type: Number,
+			default: 0,
+		},
+
+		/**
+		 * Absolute min width of popover, overrides tetherMinWidth
+		 */
+		minWidth: {
```
landondurnan, line 126: I think this could also be removed. Don't we want to calculate the width based on the element we're wrapping? There'd be a static min-width if the element is under a certain threshold. Otherwise the popover container would be the same width as the element we're wrapping.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+			type: Number,
+			default: 0,
+		},
+
+		/**
+		 * Set min width of popover to tether width, overridden by minWidth
+		 */
+		tetherMinWidth: {
```
landondurnan, line 134: Extraneous in reference to both the minWidth and that there isn't a current need to tether the popover to anything but the wrapping element.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+			type: Boolean,
+			default: true,
+		},
+
+		/**
+		 * Configures the popover to hide itself if the tether element goes out of view
+		 */
+		shouldHideOnOverflow: {
```
landondurnan, line 142: I think it's reasonable to make this default behavior always and not have a configuration for it.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+			type: Boolean,
+			default: true,
+		},
+
+		/**
+		 * ID of popper
+		 */
+		id: {
```
landondurnan, line 150: Only one popover should be visible at a time and then a unique ID isn't really needed for it. (Similar to how there can only be one modal on one layer, we don't need a unique ID to manage it's state)

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,278 @@
...
+			type: String,
+			default: () => uuid(),
+		},
+
+		/**
+		 * Custom popover config
+		 */
+		customConfig: {
```
landondurnan, line 158: I'm hesitant to reveal a third-party configuration option just yet until we have more scenarios that are accounted for in our base props. Having multiple props that configure the same type of options can cause issues as you have to make internal choices of which one to use if multiple are set.

src/components/Popover/src/PopoverBubble.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<div
+		:class="$s.PopoverBubble"
```
landondurnan, line 3: I know this was carry over, but on second look "Bubble" is so much of a metaphor of its visual style. Something more generic like "PopoverContainer" may provide a bit more meaning and clarity.

src/components/Popover/src/PopoverBubble.vue
```diff
@@ -0,0 +1,55 @@
...
+	</div>
+</template>
+
+<script>
+import chroma from 'chroma-js';
+
+export default {
+	name: 'PopoverBubble',
+
+	props: {
```
landondurnan, line 17: Rather than direct access to the popover container as an additional child component, I think we should have a single "Popover" component and route these props through the parent to this internal child.

We'll also need to add some "padding" option, more details below.

landondurnan (changes requested):
Overall, good implementation. But overall feedback is looking for ways to simplify defaults and configurations. Limiting the scope of the component right now to the simplest version can help us reduce complexity and be more purposeful about changes later.

src/components/Modal/src/ModalLayer.vue
tatems: I can probably put it back. There was a naming conflict with the popover layer but I ended up not using it. 

Though I'm not sure what you mean that there could be more than modals? This is specifically a modal layer.

src/components/Popover/index.js
tatems: I have to disagree here. I don't think we should be too prescriptive on the content of the popover. Exposing this provides a good/useful default, but something like the mini cart will need to have something different (or at least add additional styles).

In my opinion, the popover component should _only_ be responsible for creating the popper instance, and what goes inside that instance should be up to the consumer.

src/components/Popover/src/Popover.vue
tatems: The tether slot here is the same as the action slot in orbit. I actually removed the external tether api that I mentioned in the issue.

tatems: I kept it as 'tether' because this version allows a parent component to trigger the open/close of the popover, so there may not actually be an action inside the slot

src/components/Popover/src/Popover.vue
tatems: Fair enough, I can remove these and use defaults. Since we have a config prop consumers can override this anyway.

package.json
tatems: We can drop uuid and portal-vue (I thought I already had for the latter). 

We need simple portal since this version doesn't break provide/inject.

src/components/Popover/src/Popover.vue
tatems: Unfortunately we need this for mini cart, especially if you're having me drop the other props. Otherwise we'll have to roll our own solution for it.

src/components/Popover/src/Popover.vue
tatems: So we'll for sure need this as part of the mini cart, and unfortunately doesn't seem like it's something that can be changed via modifiers.

src/components/Popover/src/Popover.vue
tatems: Yup so what I've done is remove all that and just exposed a `modifiers` prop that takes in an array.

src/components/Popover/src/Popover.vue
landondurnan: I see now, just nomenclature change. I had forgotten about that detail in Orbit. But now I'm trying to remember why we needed it to begin with since at first glance the default slot and scoped slot would work the same way.

src/components/Popover/src/Popover.vue
tatems: I found a way to do it via modifiers, so I removed it

src/components/Popover/src/PopoverBubble.vue
tatems: Added!

src/components/Dialog/src/DialogLayer.vue
```diff
@@ -16,6 +16,7 @@
 					:class="$s.disableScroll"
 				/>
 				<v :nodes="dialogApi.state.vnode" />
+				<!-- <m-popover-layer /> -->
```
pretzelhammer, line 19: can you please remove this line if it's not needed?

package.json
pretzelhammer: if we can drop uuid and portal-vue let's please do so

src/components/Dialog/src/DialogLayer.vue
tatems: Done!

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,220 @@
...
+		destroyPopper() {
+			this.popperToDestroy?.destroy();
+			this.popperToDestroy = undefined;
+		},
+	},
+};
+</script>
+
+<style>
+/* Keep */
```
pretzelhammer, line 219: can you please add more information to this comment? why is it necessary to keep this empty `style` tag?

tatems: I forgot to go back and clean these up. Will remove.

src/components/Popover/src/PopoverLayer.vue
```diff
@@ -0,0 +1,141 @@
...
+			default: undefined,
+		},
+	},
+
+	provide() {
+		/**
+		 * This is to avoid name collisions for the popover portal if
+		 * multiple exist at the same 'level'.
+		 */
+		const layerId = uuid();
```
pretzelhammer, line 40: this is way overkill, lol. there's gonna be like a few popovers in the same layer at most, we don't need to use v4 uuids to prevent collisions, if we need a random string of characters we can just use `Math.random().toString(36).slice(2, 12)` which will work plenty well.

tatems: Lol sounds good. I'll do that.

pretzelhammer (approved):

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,216 @@
...
+
+	data() {
+		const vm = this;
+
+		return {
+			/**
+			 * This is used to track which popover is open. When opening, this
+			 * will be pushed into the popoverApi state as the opened popover.
+			 */
+			id: generateRandomId(),
```
privatenumber, line 123: Creating a `randomId` util for this is not worth it.

You'll get guaranteed uniqueness by closing over an `id` variable and just incrementing it each time.

package.json
```diff
@@ -122,5 +122,9 @@
   },
   "release": {
     "preset": "conventionalcommits"
+  },
+  "dependencies": {
+    "@linusborg/vue-simple-portal": "^0.1.5",
+    "@popperjs/core": "^2.10.2"
```
privatenumber, line 128: These should be peer dependencies so they're not duplicated in the consuming app.

src/utils/debug.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-console */
+const isProduction = (typeof process === 'object') && (typeof process.env === 'object') && (process.env.NODE_ENV === 'production');
+const logStyles = 'background:#212121; padding:2px; border: 1px solid #646464; border-radius:3px; color:#fff';
+
+const componentNamespace = (componentName) => (componentName?.trim() ? ` [${componentName}]` : '');
```
privatenumber, line 5: I think it should be _required_ to pass in a component name.

Have you encountered any cases where it shouldn't be required?

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,216 @@
...
+			styles.popper.minWidth = `${rects.reference.offsetWidth}px`;
+		},
+		effect({ state: { elements } }) {
+			elements.popper.style.minWidth = `${elements.reference.offsetWidth}px`;
+		},
+	},
+];
+
+export default {
+	name: 'Popover',
```
privatenumber, line 68: Remove.

This can be automatically derived from the file path.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,216 @@
...
+		};
+	},
+
+	computed: {
+		tetherEl() {
+			if (this.$el.children.length !== MAX_TETHER_VNODE) {
+				return undefined;
+			}
+
+			return this.$el.children[0];
```
privatenumber, line 162: You should be able to replace this with subslot. It's a utility to filter slots, but declaratively in the template.

src/components/Popover/src/PopoverLayer.vue
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		return {
+			[PopoverAPIKey]: api,
+		};
+	},
+};
+
+export default {
+	name: 'PopoverLayer',
```
privatenumber, line 81: Remove.

package.json
```diff
@@ -122,5 +122,9 @@
   },
   "release": {
     "preset": "conventionalcommits"
+  },
+  "dependencies": {
+    "@linusborg/vue-simple-portal": "^0.1.5",
```
privatenumber, line 127: This is a Vue.js plugin so it must be a peer dependency for the consuming app to register it to their app.

I'm hesitant to introduce this plugin just for one use-case, passing around a vNode—which is rather trivial to do. It adds another installation step to Maker and also relies on the DOM by using ids (global pollution) instead of deferring to Vue to handle rendering.

privatenumber (changes requested):

src/components/Popover/src/Popover.vue
tatems: @landondurnan thoughts?

package.json
tatems: I listed the issue with using VNodes in the PR description. Unfortunately, things like reactivity _must_ work for the upcoming `Menu` feature.

src/components/Popover/src/Popover.vue
tatems: Interesting, I didn't know this was a thing!

src/utils/debug.js
tatems: I'm ok with either. I'll make it required.

package.json
tatems: And again, this tool will be part of Vue 3 proper.

src/components/Popover/src/Popover.vue
landondurnan: I agree that we don't need to pull in another dependency to generate a unique ID.

package.json
landondurnan: @privatenumber With this component in particular I thought it would be advantageous for us to look ahead at patterns that would be established in Vue 3 and leverage those for scenarios where layering and DOM injection can become challenging to manage. Portals appear to be the best tool for the job and something that could provide the foundation for layer management. 

Perhaps it's premature to think that far out, but the patterns that are being established in other frameworks and [eventually HTML itself](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/portal) will make it easier to define, document, and potentially refactor into other languages (i.e. consideration of Maker as Web Components)

mattmewton: @landondurnan @privatenumber I don't think it's premature, especially considering that we will be upgrading to Vue 3 in 2022. I'm an advocate for introducing features that will make the migration easier, and this seems to fit that bill.

src/components/Popover/src/Popover.vue
mattmewton: +1 for removing this

src/components/Modal/src/ModalLayer.vue
```diff
@@ -26,6 +26,7 @@
 				>
 					<v :nodes="currentLayer.state.vnode" />
 				</div>
+				<m-popover-layer />
```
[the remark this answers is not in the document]
tatems: I was actually just rethinking this. I don't know if there needs to be. I'll try removing it.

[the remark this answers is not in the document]
tatems: Nope, we don't! The `ModalLayer` now injects the root-level `popoverApi`.

Ryguy11 (commented):
One thing that I noticed is that if you have an open modal with the `closeOnOutsideClick` prop passed in, clicking on a popover will close the modal. Not sure if there is an easy solution there and maybe it is just recommended to not have `closeOnOutsideClick` functionality when you have a popover.

tatems:
@Ryguy11 that's an interesting point. I'll see what I can do about that.

tatems:
@Ryguy11 actually it seems like it works fine, the click is prevented on the popover layer

package.json
tatems: Seems like general consensus is to stick with portals here.

src/components/Popover/src/Popover.vue
tatems: Moved to an incremental ID

package.json
landondurnan: I do think think there's some risks of DOM pollution without a proper layer manager and there's opportunities to clean up internally as it relates to managing the open / close state particularly between unrelated components. 

While we could use vnodes, I'm not sure the benefits of a leveraging those internals make sense over the simplicity and standardization of portals.

Either way, these are all internal API decisions that can be optimized over time. The extra peer dependencies are small and well-used and with a Vue 3 upgrade in the future likely to be further optimized. I think we can move forward with this solution for now to unblock and revisit later if necessary.

landondurnan (approved):

src/components/Popover/index.js
landondurnan: While I still don't think the child component is necessary, it is a pattern we established with Modal / Dialog and is reasonable to continue. With the changes for padding, this also makes it easier to override the default padding on that component directly. Making note in this resolution we changed the name to PopoverContainer.

package.json
privatenumber: Any new technology should be vetted before considering it for adoption. I'm not against using portals but I haven't used it yet so I'm not sure how it would address the same problems Orbit's layering solution does.

Has anyone done an evaluation on it? Vue 3 has a lot of "hype" momentum (especially at ECOM). We should be careful not to let it drive engineering decisions as if it's some sort of silver bullet.

If we eventually upgrade to Vue 3, and we learn that portals suits our needs, changing something like this would be trivial. However, if we take a long time to upgrade (no one currently has a strategy), or we learn portals does not address our needs, this incurs debt.

src/components/Popover/src/Popover.vue
```diff
@@ -0,0 +1,214 @@
+<template>
+	<div style="display: contents;">
+		<!-- @slot Element that the popover is tied or 'tethered' to. Must be a single node. -->
+		<slot
+			name="tether"
```
privatenumber, line 5: Rename to "action" (consistent with Orbit).

I understand the popover is tethered to the content of this slot, but this slot does not contain the "tether". It contains the "button", "trigger", or "action" that the popover is tethered to.

privatenumber (changes requested):

src/components/Popover/src/Popover.vue
landondurnan: I was going back and forth on this too, my first reaction was also that `action` made more sense, but it might have been a bias since we used that term often in Orbit.

package.json
tatems: Even if we never upgrade to Vue 3 (which is doubtful), that doesn't mean we can't take advantage of the teleport/portal pattern. And I believe that this is a good proof of concept to help us decide whether we take the plunge for other aspects of layering, on top of the other benefits @landondurnan mentioned.
And if it doesn't work, we can always do the opposite of what you've suggested and move back to vnodes.

src/components/Popover/src/Popover.vue
tatems: Changed!

package.json
privatenumber: Again, best engineering practice is vetting before considering for adoption. There is no technical analysis of why this is a preferable approach.

We should answer:
- What is the motivation for this change?
 	_(The original motivation that you reactivity does not work has been debunked)_
- What are the advantages of this implementation (over what we have working)?
 	_(The work necessary to migrate a vNode approach to be Vue 3 compatible is trivial because VDOM exists in Vue 3)_
- How does this handle multiple layers? (eg. Popover in Modal over another Modal, or Popover on page under a Modal)
- Are there performance limitations? After all, one of the biggest benefits of using Vue.js is for the performance gains of interacting with a virtual DOM instead of DOM.
- What are the drawbacks of this approach?
- Are there any SSR concerns?


If you need to move forward without an assessment, you can dismiss my review.

landondurnan: Appreciate the considerations and write-up @privatenumber. We should definitely look further into these and be prepared to swap out for a vnode alternative internally if necessary. I think the scope of this work and the current risk of introducing this method of handling DOM injection is suitable for the time being. Using this version and testing further in other components and for some upcoming features can provide us with additional opportunity to test and validate these issues further.

I think there's a lot more we need to do as @privatenumber mentioned to validate those concerns and also think more holistically about our overall layer management. Given some demands to move forward with some features now I think we can unblock with this method and keep revisiting these issues in the future. An internal change to anything shouldn't impact the public API right now, so we can keep discussing and move forward on this solution.

src/components/Modal/src/ModalLayer.vue
sohnj877: @tatems This is throwing an error. I think we need to do `vm.popoverApi?.closePopover();` here

PR #177, opened 2021-11-05
Color audit

landondurnan:


src/components/Theme/src/default-theme.js
```diff
@@ -41,19 +44,17 @@ export default function defaultTheme() {
 		text: {
 			fontFamily: 'inherit',
 			size: 0,
-			textColor: undefined,
 		},
 		heading: {
 			fontFamily: 'inherit',
 			size: 2,
-			textColor: undefined,
```
laurenhu, line 49: If we set this and `text` to `textColor: '@colors.heading'` and `textColor: '@colors.text'`, we won't need to create additional variables for `--color-text` and `--color-heading`.  I also like showing undefined values in this default theme because it makes clear what all the themable props are for each themeable component.

src/components/Heading/src/Heading.vue
```diff
@@ -160,7 +160,7 @@ export default {
 
 .Heading {
 	margin: 0;
-	color: var(--color-900);
+	color: var(--color-heading);
```
laurenhu, line 163: We can drop this altogether if we set the heading textcolor in the default theme.

src/components/Text/src/Text.vue
```diff
@@ -118,7 +118,7 @@ export default {
 }
 
 .Paragraph {
-	color: var(--color-800);
+	color: var(--color-text);
```
laurenhu, line 121: Same here with text color. Though I do see the utility of a `--color-text` variable that can be used on the base theme wrapper for global background and color.

lab/experiments/ThemeTest.vue
```diff
@@ -340,15 +353,30 @@ function contrastColors(bgHex) {
...
 
 	Object.entries(levels).forEach(([name, level]) => {
-		colors[`color-${name}`] = chroma.mix(bgHex, contrastColor, level, 'lab').hex();
+		colors[`neutral-${name}`] = chroma.mix(bgHex, contrastColor, level, 'lab').hex();
 	});
 
 	return {
 		...colors,
-		'color-elevation': isLight ? '#ffffff' : colors['color-300'],
+		'color-elevation': isLight ? '#ffffff' : colors['neutral-30'],
```
laurenhu, line 379: `colors['neutral-20']` for the dark backgrounds

laurenhu (commented):
This all looks great! The scale and the naming are great. Design has reviewed and approved all the color ratios. Just a minor bug on the lab and a note on the where the heading/text default colors are set.

lab/experiments/ThemeTest.vue
landondurnan: missed this one, thanks

src/components/Text/src/Text.vue
landondurnan: That's primarily why I set that up. I also just made some changes here to reflect other ways this could be used between both the heading and text property: https://github.com/square/maker/pull/177/commits/0f525951a7f95edf28318eae8b973d3fe0669de8

My suggestion is that all text against the background should use the same text color by default, so that would update things like radio / checkboxes:
CleanShot 2021-11-12 at 16 13 35

With something like the calendar which has both a heading (Date Year) and the cells it seems more appropriate as well to leverage the heading and text color defined in the profile vs a neutral set.
CleanShot 2021-11-12 at 16 13 41

lokesh:
Question about the intended contrast behavior for Maker components: Since text and headings will have static color values, unlike the neutrals, is it expected that the consuming app handle the contrast checking?

image

lokesh:
Incorrect color var used for the selected buttons text color:
image

landondurnan:
the issue with the selected button color has been fixed, missing "elevation" color setting that @laurenhu pointed out. The lab hasn't been updated with the changes yet, going to do a bit more work on disabled states and will push an update.

src/components/Theme/src/default-theme.js
landondurnan: This was intentional, but may be slightly subversive. The colors for heading, text, and other color tokens that are part of our color profiles aren't always a one-to-one relationship between the token and the component. Global generic color tokens should be reusable.

For example, the Calendar component has a header that does not use the Heading component. The text color of our inline forms should inherit the text color, but they aren't using the text component.

So an opinionated approach here is rather than explicitly declaring the defaults on components in the theme definition we should look at the theme definition as setting global reusable tokens. This will be something I'll push here and keep on pushing, but this was my gentle nudge regarding this direction.

Considering future scenario we will need to build, the "shape" token that should define the visual style of more than just the button. Given our current approach we have three distinct and named shapes in the button component. The implication is that we will duplicate these three shapes in every component. Then the theme definition will have a declaration for every component to then set the same prop on every component. It's an incredible amount of extra code that isn't necessary and would be better served through a single prop like borderRadius that defaults to a global token(s). This is the pattern established in other themable design systems like Material and defined in the Theme UI spec as `radii`.  https://github.com/square/maker/blob/ba223d1a518cc156d265d79715369d04df4ae3fa/lab/experiments/Theme/themes.js#L64-L66

This simple generic definition is far easier to understand, more flexible, and provides far less of a footprint for us and others to have to manage as we expand on our theme definitions.

laurenhu (approved):
Since we allow website users to customize their paragraph text color, I think `--color-text` is useful for components that don't have a color prop (such as form inputs, calendar, etc). Not sure if `--color-heading` will be as widely used since MHeading has a themeable color prop.

We also might want to consider just using MHeading anywhere we would need a heading (like in calendar) so that future themeable heading tyles in terms of font family, weight, spacing, etc are applied correctly.

Just food for thought - not at all worth holding up the release!

pretzelhammer (changes requested):
all of this looks good, there's a few small but important things Lauren pointed out in her comments in this PR which I think are worth changing and I've implemented those changes here: https://github.com/square/maker/pull/206

I think we should merge that PR into this one and once that's done we can merge this PR into beta

pretzelhammer (approved):

PR #181, opened 2021-11-05
feat(modal): add drag/swipe to dismiss on mobile modals

laurenhu:
PR for drag/swipe to dismiss functionality on mobile modals as prototyped below.

- Modal is only swipe/draggable in mobile
- `on-swipe-down` is a single point touch event, min 30px in distance, max 300ms in duration. It closes the Modal
- `on-drag-down` is a single point touch even
- `on-drag-end` is the end of a drag event. Modal closes if it's been dragged more than 30% client height. Otherwise it bounces back into place.
- Touch to dismiss events only work when Modal is scrolled to the top. When modal is scrolled up/down, there's a 800ms delay before dismiss events are active again.
- The distance/timing thresholds  swipe events can be adjusted, as can the minimum drag to close

Design:
screencast 2021-10-06 11-24-20
Lab:
https://square.github.io/maker/lab/swipe-dismiss/#/ActionBarTest

### DIFF
--- a/lab/experiments/ActionBarTest.vue
+++ b/lab/experiments/ActionBarTest.vue
@@ -1,8 +1,10 @@
 <template>
-	<m-action-bar-layer>
-		<router-view />
+	<div>
+		<m-action-bar-layer>
+			<router-view />
+		</m-action-bar-layer>
 		<m-modal-layer />
-	</m-action-bar-layer>
+	</div>
 </template>
 
 <script>
--- a/lab/experiments/ActionBarTest/CartModal.vue
+++ b/lab/experiments/ActionBarTest/CartModal.vue
@@ -1,10 +1,14 @@
 <template>
 	<m-modal>
-		<div class="cover-photo">
-			<m-image
-				src="https://i.picsum.photos/id/507/900/900.jpg?hmac=NDltE7xXtFlZjUoyDqGjehzY5ORPtj4-d42qbFgAFkk"
-			/>
-		</div>
+		<template
+			#dismiss
+		>
+			<div class="cover-photo">
+				<m-image
+					src="https://i.picsum.photos/id/507/900/900.jpg?hmac=NDltE7xXtFlZjUoyDqGjehzY5ORPtj4-d42qbFgAFkk"
+				/>
+			</div>
+		</template>
 
 		<div>
 			<h1>Cart modal content</h1>
--- a/lab/experiments/ActionBarTest/CartModal.vue
+++ b/lab/experiments/ActionBarTest/CartModal.vue
@@ -3,23 +3,36 @@
 		<template
 			#dismiss
 		>
-			<div class="cover-photo">
+			<div
+				v-if="showImage"
+				class="cover-photo"
+			>
 				<m-image
 					src="https://i.picsum.photos/id/507/900/900.jpg?hmac=NDltE7xXtFlZjUoyDqGjehzY5ORPtj4-d42qbFgAFkk"
 				/>
 			</div>
-		</template>
-
-		<div>
-			<h1>Cart modal content</h1>
-			<div
-				v-for="i in 100"
-				:key="i"
+			<m-segmented-control
+				v-else
+				v-model="selected"
 			>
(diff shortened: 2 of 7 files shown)


src/components/TouchCapture/index.js
```diff
@@ -0,0 +1 @@
+export { default as MTouchCapture } from './src/TouchCapture.vue';
```
landondurnan, line 1: Personal mental note, we should categorize these helper componenta a bit more in the styleguide. Maybe it's time to organize our list beyond alphabetical.

src/components/TouchCapture/src/TouchCapture.vue
```diff
@@ -0,0 +1,158 @@
+<template>
```
landondurnan, line 1: This looks great, but I am curious if there's any small third-party packages that'd give us all this and more? I could see us doing a lot more with these touch events, particularly with other things we're seeing like carousel (swipes) and zooms (pinch).

landondurnan (commented):
Looking great @laurenhu. Threw some ideas out as I think we might be able to optimize a bit around the dismiss container. Also I'm not as well-versed in some of the issues we saw with the actionbar so hoping Kirill can weigh-in there.

lab/experiments/ActionBarTest.vue
```diff
@@ -1,8 +1,10 @@
 <template>
-	<m-action-bar-layer>
-		<router-view />
+	<div>
```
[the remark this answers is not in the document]
laurenhu: This is actually how we have it set up in prod...I think the experiment was out of date. I had to update it to get the modals working properly in both this and the action bar branch.

```
		<m-action-bar-layer
			id="app"
			:class="classes"
		>
			<router-view />
			<layer-manager />
			<checkout-layer />
		</m-action-bar-layer>
		<m-modal-layer
			:class="modalClasses"
		/>
		<m-dialog-layer />
```

src/components/TouchCapture/src/TouchCapture.vue
landondurnan: I poked around a bit and found a number of libraries but ultimately their benefits is mostly around a cleaner looking code. The downsides is we introduce an additional dependency, additional filesize, and hitching our wagon to something that isn't likely to get many updates later as the community is moving to Vue 3.

https://openbase.com/categories/js/best-vue-touch-events-libraries

lab/experiments/ActionBarTest/CartModal.vue
```diff
@@ -1,10 +1,14 @@
 <template>
 	<m-modal>
-		<div class="cover-photo">
-			<m-image
-				src="https://i.picsum.photos/id/507/900/900.jpg?hmac=NDltE7xXtFlZjUoyDqGjehzY5ORPtj4-d42qbFgAFkk"
-			/>
-		</div>
+		<template
+			#dismiss
+		>
```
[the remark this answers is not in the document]
landondurnan: What was the resolution? The dismiss slot is still needed?

src/components/Modal/src/Modal.vue
```diff
@@ -1,16 +1,44 @@
 <template>
-	<div :class="$s.Modal">
+	<div
+		:class="$s.Modal"
+		:style="modalStyles"
+	>
+		<template v-if="$slots.dismiss">
```
landondurnan: I think we could simplify this by always having the the touch capture wrap the default slot. Only have one of them (don't need separate ones for mobile / desktop) and then simply hide invisible container on desktop.

laurenhu: I'm not sure why but the PRs on maker don't  auto-hide code that's outdated...just adds an Outdated tag. But all the code around dismiss slots have been removed - the actions are just on the modal itself...there are no touch actions on desktop so it's only on mobile.

landondurnan: Oh weird! I might have dug into a cached commit and didn't see the updates.

landondurnan (approved):
Great work!

PR #183, opened 2021-11-08
feat(menu): add menu component

tatems:
This PR provides some utility components based on the `Popover` component to provide contextual actions or selections for users.

### Menu
The `Menu` component provides contextual clickable actions using `Popover`.
 
image

### MenuSelect
The `MenuSelect` component is designed to be a replacement for the native `select` (and by extension the Maker `MSelect` component).
`select` is fairly limited in it's styling capability. `MenuSelect` takes advantage of `Menu` and `Popover` to provide a better UI design, while also allowing for multiselect and other customizations.

image


landondurnan (commented):
I'm really on the fence about separating the functional requirements into the two different components with `Menu` and `MenuSelect`. There's a part of me that believes they share far more in common than they are different and even if there are misc props that may not be used in certain context that's okay. However, I also see that we're internally using the Menu and MenuItem component in the MenuSelect so most of what they'd share in common is still managed (or passed). This is a gray area for me and I'm hoping we can debate this a bit more, we've been okay with duplication and internal component usage -- this checks the same boxes as other components we have. So maybe this is following existing philosophy.

I think the primary concern I have, and this is a larger philosophical issues we need to align on, is our approach to the combinations of child component, slots, and props.  We've done all of these at different times and ultimately controlling more in the template than internally in the component has provided the most flexibility.

For example, the `options` prop provides a simpler way of passing data to the component. But internally this creates more complexity. We're having to do checks and evaluate if there's data, if there's a child component, how do we deal with the v-model between both. The alternative is provide one way to implement this output and lean on controlling in the template. For consumers, it's more code -- but it's less complex and error prone and provides the maximum amount of flexibility.

I'm hoping some others can weigh in here, particularly @pretzelhammer as I've had these debates with him in the past. There's no wrong answer here, overall this code looks fine. It's just our larger philosophy. Also I reached out to design on context of defining these components with different names so we can keep our documentation / nomenclature in sync.

src/components/Menu/src/MenuSelect.vue
```diff
@@ -0,0 +1,190 @@
...
+		 * Minimum width of popover items
+		 */
+		popoverModifiers: {
+			type: Array,
+			default: () => undefined,
+		},
+	},
+
+	watch: {
+		isMultiselect: {
```
landondurnan, line 131: Why do we need to watch this? Shouldn't v-model handle all the required update values?

landondurnan (commented):
I think it's worth considering a simplification of the API, particularly around the props and data. We have very flexible component in Orbit called "Tiles" that has a succinct approach to the way we approach the functional props, v-model, and template flexibility:

https://github.com/squareup/orbit-ui/blob/master/src/components/Tiles/src/Tiles.vue

The code is pretty straightforward with the props with `mode` (single / multi-select) and `mandatory` (required value). The model and methods are very simple and easy to understand.

PR #184, opened 2021-11-09
feat(inline-action-bar): delayed spring transition

laurenhu:
This PR adds a delayed bouncy spring up transition to the inline action bar, as well as a lab experiment to preview different delay/stiffness/damping/mass values for the transition.

- `ActionBarTransition`: experiment to demo inline action bar spring transition values
- `transitions`: add  `springUpBounceFn` and `springDownBounceFn` for the new default inline action bar transitions
- `InlineActionbar`: add `enterAnimation`, `leaveAnimation`, and `enterDelay` props

Nov-09-2021 15-32-36

### DIFF
--- a/src/components/ActionBar/src/InlineActionBar.vue
+++ b/src/components/ActionBar/src/InlineActionBar.vue
@@ -1,16 +1,23 @@
 <template>
 	<div :class="$s.ActionBarWrapper">
-		<atomic-action-bar
-			v-bind="$attrs"
-			v-on="$listeners"
+		<m-transition
+			:enter="enterAnimation"
+			:leave="leaveAnimation"
 		>
-			<!-- @slot ActionBar items -->
-			<slot />
-		</atomic-action-bar>
+			<atomic-action-bar
+				v-if="loaded"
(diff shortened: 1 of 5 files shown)


src/components/ActionBar/src/InlineActionBar.vue
```diff
@@ -21,9 +28,37 @@ import AtomicActionBar from './AtomicActionBar.vue';
 export default {
 	components: {
 		AtomicActionBar,
+		MTransition,
 	},
 
 	inheritAttrs: false,
+
+	props: {
```
landondurnan, line 36: If we make these public we'll need to add documentation for them. 

In the short-term I'm wondering if we should hold our cards a bit closer to our chest and keep the configurations as internal only. This way we could explore more global configurations across several components and consider how theming might enable different styles of animations.

laurenhu: Removed the props from InlineActionbar so that there's just the default transition and delay, similar to ActionBar.

landondurnan (approved):
Excited to see this in an integration!

Issue #169, opened 2021-10-22
[Actionbar] Reduce padding at standalone mode


sohnj877:
### Feature request

We should remove the extra deadclick padding on the actionbar when site is running in standalone mode.

### Why?

Extra padding is unnecessary since the standalone mode hides browser UI.

### Additional context

![Screen Shot 2021-10-22 at 1 31 03 PM](https://user-images.githubusercontent.com/7245055/138501994-0b9e7977-c91b-4001-9ba1-18f460009807.png)

https://user-images.githubusercontent.com/7245055/138502040-de8499ce-fd00-4290-9a6c-5f1a44c9044d.mp4

landondurnan:
I'm going to make an attempt on this. I had a version here (https://github.com/square/maker/pull/80) but code conflicts prevented merge at the time.

lokesh:
@landondurnan Is `env(safe-area-inset-*)` not working because of our implementation, or is it not a viable option to solve our problem? If it is viable, are there examples online using the css property that we can test?

landondurnan:
Resolved in 6.3.1

PR #186, opened 2021-11-10
fix(actionbar): adjust padding to be 24 and include env inset calc

landondurnan:

https://github.com/square/maker/issues/169

Adjusted padding to be the desirable 24px all the way around (original design intent)
CleanShot 2021-11-10 at 15 18 28

Simplified the padding calculations with the [env()](https://developer.mozilla.org/en-US/docs/Web/CSS/env())

Added fallbacks in case env isn't supported.

## Other information

Previously we had a significant amount of extra padding added to get around the nefarious Safari "deadclick" zone. The usage of the env safe zone should have resolved that, but it looks some of those calculations resulted in roughly the same less than ideal positioning. Tested thoroughly on desktop + mobile on Chrome and Safari. Even did a built test on our PWA version:
Screen Shot 2021-11-17 at 10 51 09 AM

### DIFF
--- a/src/components/ActionBar/README.md
+++ b/src/components/ActionBar/README.md
@@ -38,7 +38,7 @@ There are 2 versions of ActionBar:
 							key="confirm"
 							full-width
 						>
-							Confirm Action
+							Confirm action
 						</m-action-bar-button>
 					</m-inline-action-bar>
 				</div>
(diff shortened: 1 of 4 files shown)


landondurnan:
Appears to work as expected in Safari and Chrome. Looking into a lab deploy to test the "Save to homescreen" app-like experience.

landondurnan:
Found some issues on desktop. Will re-open when they're fixed.

lokesh (approved):
The code refactor using `safe-area-inset-bottom` makes this 10x simpler. 

I didn't test myself, so approving with the assumption this was tested on mobile. It's worth testing in iOS Safari with both the address bar at the top and at the bottom (the new default in the latest major iOS release)

sohnj877 (approved):
LGTM 👍

PR #191, opened 2021-11-17
feat(transition-staggered): staggered transition group

laurenhu:
This PR adds a transition group component that allows for staggered list transitions.

- `ContentAnimation`: experiment to demo the staggered transition
- `transitions`: add  `springSubtle`, `toSubtleFloatyY` constants and `staggeredFloatUpFn` for the designed transition that accepts a stagger
- `TransitionStaggered`: transition group component that takes a stagger item count by viewport and passes back a function to calculate the load index.

For a mobile stagger count of 1 item at a time and desktop stagger count of 4 at a time:

mobile

desktop

### DIFF
--- a/src/components/TransitionStaggered/src/TransitionStaggered.vue
+++ b/src/components/TransitionStaggered/src/TransitionStaggered.vue
@@ -1,0 +1,91 @@
+<template>
+	<transition-group
+		v-bind="$attrs"
+		tag="div"
+		@enter="handleEnter"
+		@leave="handleLeave"
+	>
+		<slot
+			:data-load-index="loadIndex"
+		/>
+	</transition-group>
+</template>
+
+<script>
+import { staggeredFloatUpFn, floatDownFn } from '@square/maker/utils/transitions';
+
+const defaultItemCount = [
+	{
+		minWidth: 0,
+		itemCount: 1,
+	},
+];
+
+export default {
+	props: {
+		// Number of items to stagger at a time for a viewport
+		// E.g. [
+		// 	{
+		// 		minWidth: 0,
+		// 		staggerItemCount: 1,
+		// 	},
+		// 	{
+		// 		minWidth: 840,
+		// 		staggerItemCount: 4,
+		// 	},
+		// ]
+		staggerItemCount: {
+			type: Array,
+			default: () => defaultItemCount,
+			validator: (staggerItemCount) => {
+				// cannot be empty
+				if (staggerItemCount.length === 0) {
+					return false;
+				}
+				// must have default catch-all count
+				if (staggerItemCount[0].minWidth !== 0) {
+					return false;
+				}
+				return staggerItemCount.every((itemCount) => (itemCount.minWidth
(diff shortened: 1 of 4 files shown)


src/components/TransitionStaggered/src/TransitionStaggered.vue
```diff
@@ -0,0 +1,91 @@
...
+		// 	{
+		// 		minWidth: 0,
+		// 		staggerItemCount: 1,
+		// 	},
+		// 	{
+		// 		minWidth: 840,
+		// 		staggerItemCount: 4,
+		// 	},
+		// ]
+		staggerItemCount: {
```
lokesh, line 37: What about a name like `itemsPerRow` for this prop? Still doesn't capture the media query support, but more clear about what the number means.

src/components/TransitionStaggered/src/TransitionStaggered.vue
```diff
@@ -0,0 +1,91 @@
...
+			let items;
+			staggerItemCount.forEach((count) => {
+				if (count.minWidth < viewportWidth) {
+					items = count.itemCount;
+				}
+			});
+			if (items === defaultItemCount[0].itemCount) {
+				return index;
+			}
+			const remainder = index % items;
```
lokesh, line 86: Posted a design suggestion in [Slack](https://square.slack.com/archives/C0255HML1PD/p1637192756102800) to consider row based staggering on desktop. This can be tested locally by updating the return value to: `Math.floor((index - 1) / items);`
stagger-row

src/components/TransitionStaggered/src/TransitionStaggered.vue
landondurnan: There's something about the name bugging me too -- also the complexity of the data object we're passing in. I'm wondering if we can split it up into multiple props, but I'm at a loss right now on what it should be. 🤔

landondurnan: I'm thinking about how you would construct a grid like this with flex or css grid -- perhaps we can model the props the same way you'd do with regular CSS.  Rather than doing calculations based on the viewport width and passing data about the width the consumer should just consider the grid they want and pass those details.

For that matter, now I'm kind of wondering if a list or collection component responsible for the layout of that type of content should be used and the transition is more baked in with the skeleton loaders.

So many ways to skin this cat...

laurenhu: I tried an earlier version that had a `skeleton` slot and more layout handling. But it would have required refactoring of how BX views handle loading states and grids/grid props currently. It felt like a task for a second pass once we can cover all current use cases.

For now, in order to have a Maker integration that can be used for testing on Website, the simplest component I needed was a transition that handles a staggered load animation for a group of elements. It would still be useful within a more comprehensive loading component.

As far as checking the window viewport - that's also how Maker handles applying different transition animations for Modals on mobile and web. For now the transition component has no awareness of how a list of elements is displayed, it just needs to know how to group them for animation sequencing.

PR #194, opened 2021-11-30
ci: use setup-node with node-version-file

privatenumber:

Previously, GitHub's official Node.js action, `setup-node` did not support reading from `.nvmrc` so we had to use a custom bash instance to use nvm to setup node.

However, they recently added support for  [`node-version-file`](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file) which means we can go back to using `setup-node`.

`setup-node` is GitHub's [official recommendation](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs-or-python#specifying-the-node.js-version) for installing a specific version of Node.js.

- Use `setup-node` with `node-version-file`

## Other information

Will follow up with a dependency caching PR

### DIFF
--- a/.github/workflows/deploy-styleguide.yml
+++ b/.github/workflows/deploy-styleguide.yml
@@ -20,14 +20,16 @@ jobs:
           repository: square/maker
           ref: deploys
           path: .dist
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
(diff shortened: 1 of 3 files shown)


landondurnan (approved):
Nice!

PR #196, opened 2021-12-02
feat(transition-staggered): staggered transition

laurenhu:
This PR adds a transition group component that allows for staggered list transitions.

- `ContentAnimation`: experiment to demo the staggered transition
- `transitions`: add  `springSubtle`, `toSubtleFloatyY` constants and `staggeredFloatUpFn` for the designed transition that accepts a stagger
- `TransitionStaggered`: transition group component that takes a stagger item count by viewport and passes back a function to calculate the load index.

For a mobile stagger count of 1 item at a time and desktop stagger count of 4 at a time:

mobile

desktop

### DIFF
--- a/lab/experiments/ContentAnimateIn.vue
+++ b/lab/experiments/ContentAnimateIn.vue
@@ -5,7 +5,7 @@
 			:class="$s.ProductGrid"
 		>
 			<div
-				v-for="index in 12"
+				v-for="index in 16"
 				:key="index"
 				:class="$s.Product"
 			>
@@ -19,7 +19,7 @@
 		>
 			<template #default="{ dataLoadIndex }">
 				<div
-					v-for="index in 12"
+					v-for="index in 16"
 					v-show="reveal"
 					:key="index"
 					:data-load-index="dataLoadIndex(index)"
--- a/lab/experiments/ContentAnimateIn.vue
+++ b/lab/experiments/ContentAnimateIn.vue
@@ -30,7 +30,7 @@
 					:src="`https://picsum.photos/600/300?${i}`"
 					height="200px"
 				>
-				<div>This is item number {{ Math.random() }}</div>
+				<div>This is item number {{ i }}</div>
 			</div>
 		</transition-group>
 	</div>
--- a/lab/experiments/ContentAnimateIn.vue
+++ b/lab/experiments/ContentAnimateIn.vue
@@ -1,0 +1,105 @@
+<template>
+	<div :class="$s.ProductLoad">
+		<div :class="$s.ProductGrid">
+			<div
+				v-for="i in 12"
+				v-show="!reveal"
+				:key="i"
+				:class="$s.Product"
+			>
+				<m-skeleton-block :class="$s.ProductImage" />
+				<m-skeleton-text />
+			</div>
+		</div>
+		<transition-group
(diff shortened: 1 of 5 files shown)


laurenhu:
Replaces https://github.com/square/maker/pull/191 - we went a different direction with the transition effect.

pretzelhammer:
wow, this is pretty cool, can you add a readme.md page to document this component because it's difficult to understand how it works and should be used from just looking at the code itself

src/utils/transitions.js
```diff
@@ -228,6 +242,25 @@ export function delayedFloatUpFn({ element, onComplete }) {
 	}, springDelay);
 }
 
+export function staggeredFloatUpFn({ element, onComplete }) {
+	const elementStyler = styler(element);
+	const styleFn = toStaggeredFloatyY;
+	const animationDirection = animateUp;
+	const distanceBase = 20;
+	const distanceMultiplier = 5;
+	const endValue = distanceBase + (element.dataset.loadIndex * distanceMultiplier);
```
pretzelhammer, line 251: oh interesting, so `data-load-index` has to be added to each DOM element because that tells the JS how much to animate it by

lab/experiments/ContentAnimateIn.vue
```diff
@@ -0,0 +1,105 @@
...
+				:class="$s.Product"
+			>
+				<template
+					v-if="!reveal"
+				>
+					<m-skeleton-block :class="$s.ProductImage" />
+					<m-skeleton-text />
+				</template>
+				<m-transition-staggered
+					:stagger-item-count="staggerItemCount"
```
landondurnan, line 18: I'm always looking for ways that we can simplify public api and make it easier to understand. My primary concern at the start was this complex prop and understanding how to use it. When testing locally and removing this prop, I didn't noticed a difference on either mobile or desktop with how the staggered transition loaded.

If this wasn't a configuration, what are we missing out on?

laurenhu: This property tells it how many items per row are being staggered in the animation. If it's not provided, it just animates all the items in at the same time without staggering.

with prop
staggered

without prop
unstaggered

laurenhu:
@pretzelhammer Good call - documenting everything in a README helped highlight several ways of simplifying the component so that everything could be handled with simple number props. Now nothing needs to be applied to the element itself.

@landondurnan Hopefully these changes make the transition simpler/more user-friendly.

src/components/TransitionStaggered/README.md
```diff
@@ -0,0 +1,98 @@
+# Transition Staggered
+
```
pretzelhammer, line 2: the example is pretty clear, but i would maybe write a couple sentences here at the top, something like "use MTransitionStaggered when you wanna animate a group of items (typically in a list or a grid) into view after lazy-loading them. the MTransitionStaggered components **has to be** wrapped around every item in the grid/list and it will know how to animate the item given the `item-index` prop"

pretzelhammer (approved):
code-wise looks good! just left one small comment with a suggestion on how to improve the docs a little bit more

landondurnan (changes requested):
I have some suggestions and questions about the props, but doing some local testing and will follow up once I'm able to replicate intended behavior with less code.

PR #199, opened 2021-12-03
docs: reduce markup and CSS for StaggeredTransition Demo

landondurnan:

The demo used a bit more markup and styles than necessary.

does the same thing with less code.

## Other information

I added an output of the index to make it a bit easier to see which item is transitioning in when.

### DIFF
--- a/lab/experiments/ContentAnimateIn.vue
+++ b/lab/experiments/ContentAnimateIn.vue
@@ -5,7 +5,7 @@
 			:class="$s.ProductGrid"
 		>
 			<div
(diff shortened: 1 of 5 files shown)


landondurnan:
@laurenhu That was my intent with the grid design, but I realize now that it wouldn't produce consistent transitions at each "breakpoint". Given we're not using grid in production now this may be premature to push, but I'm hoping we can start moving in that direction. With that being said it would likely require rethinking the transition from the ground up. We discussed that potential future state so I'm just going to close this out for now.

PR #202, opened 2021-12-09
fix(modal): fix background site drag for modal on safari

laurenhu:
This PR addresses the previously observed bug where dragging the modal also dragged the underlying site on mobile Safari.  This produces an unwanted visual effect and can also cause the page to reload.

Bug:
https://user-images.githubusercontent.com/1486885/145483791-21495336-b1cc-46ce-a146-965f6b3a6f89.mov

Fix:
https://user-images.githubusercontent.com/1486885/145483830-456e08c0-6c19-4bcc-8b7c-bb6b6b579af5.mov

### DIFF
--- a/src/components/TouchCapture/src/TouchCapture.vue
+++ b/src/components/TouchCapture/src/TouchCapture.vue
@@ -27,6 +27,10 @@ export default {
 	name: 'TouchCapture',
 
 	props: {
+		preventDefault: {
+			type: Boolean,
+			default: false,
+		},
 		minSwipeDistance: {
 			type: Number,
 			default: 30,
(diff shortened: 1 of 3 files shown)


src/components/TouchCapture/src/TouchCapture.vue
```diff
@@ -107,6 +111,9 @@ export default {
 
 	methods: {
 		handleTouchEvent(event) {
+			if (this.preventDefault) {
+				event.preventDefault();
```
mattmewton, line 115: Based on the description in the ticket, it actually sounds like you want `event.stopPropagation()` instead of `preventDefault`.

laurenhu: I tested this but it doesn't look like `event.stopPropagation` does the trick here because scrolling is a default behavior which stopPropagation doesn't prevent.
https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

mattmewton: TIL!

mattmewton (approved):

PR #203, opened 2021-12-13
feat(transitions): transitions changes for modal, actionbar and staggered transition

laurenhu:
This PR combines previously reviewed components and changes for the Transitions release, including:
- Modal swipe or drag to close functionality on mobile using the `TouchCapture` component
- New `springUpBounceFn` and `springDownBounceFn` transitions on the default and inline action bar
- New `TransitionStaggered` component that takes a row count and item index and animates in item with a delay


landondurnan (approved):
All previously reviewed. Will help unblock a lot of important transition polish.

PR #206, opened 2022-01-11
refactor: Color audit feedback

pretzelhammer:
implemented feedback resolving convos in this pr: https://github.com/square/maker/pull/177

once this PR is merged into its parent `color-audit` then its parent will be able to be merged into its grandparent `beta` which will then be able to be merged into its great grandparent `master` (after we do some internal QA for the Maker upgrade in Website) and we'll finally have theming in the mainline branch!!


src/components/Theme/src/Theme.vue
```diff
@@ -79,10 +78,3 @@ export default {
 	},
 };
 </script>
-
-<style module="$s">
-	.Theme {
```
landondurnan, line 84: If we're not setting the color and background color here, where is it done? @pretzelhammer @laurenhu

pretzelhammer: MTheme should only theme Maker components. A lot of native HTML components have `color: inherit` or `background-color: inherit` as default styles which would be affected by these styles set from within MTheme. The mechanism to customize the color & background-color of different individual Maker components should be via passing props which is how it's currently done in every component.

pretzelhammer: if this breaks the code in ThemeTest lab then that code should be changed to set these styles explicitly in the lab code instead of doing it through the MTheme component.

pretzelhammer: I just tested the ThemeTest lab locally and it works the same as it does in the parent branch, removing these did not introduce any regressions.

laurenhu: I don't think we should remove these. It feels strange to tell Maker components to expect that they are rendering against a give background color, but not actually apply that background to the theme wrapper.

In terms of ease of use, it makes more sense to apply background and text color to MTheme as specified in `@colors` automatically, rather than pass them for the benefit of child Maker components but also have to add them as styles to the`<m-theme>` parent.

pretzelhammer: >  It feels strange to tell Maker components to expect that they are rendering against a give background color, but not actually apply that background to the theme wrapper.

It only feels strange if we assume that Maker is the only component library being used to render an entire app. I agree it is strange in that specific scenario. However, since Maker is being integrated into other much larger codebases which have a lot of non-Maker components then I feel that MTheme implicitly setting a `color` and `background-color` styles in the root of the app (which is where we will be instructing devs to place the MTheme component) will be surprising to devs as it will likely clash with the non-Maker parts of their app. Both `color: some-value` and `background-color: some-value` are leaky CSS styles and will leak into the non-Maker parts of their app.

I only feel strongly about this _because it has already bitten me_. I was playing with the `color-audit` branch locally while working on the MTheme docs and noticed the `text-color` prop was affecting text outside of the MText component and found it really unexpected and jarring. I honestly thought it was a bug, but was surprised to learn that it was done this way intentionally while trying to debug it. I think this will confuse more devs than me so I think it should be removed from the beta. 

@laurenhu @landondurnan  If I've convinced you please approve this PR, but if not please let me know and I'll add these styles back in. I think these will get removed eventually, but whether we do it now or in 2 months from now as a breaking change in a Maker major version release are both okay with me.

landondurnan: I think nearly all design systems, particularly ones that are themeable, take the approach of setting global CSS at the app level. This isn't something we should shy away with, it's something we should embrace. Static CSS variables set at the global level provide the easiest, fastest, and cleanest way to style the entire application. Setting the background and color at the top level with CSS is extremely common and expected.  It's a reasonable expectation that the design system will define these rules globally and it's going to make everyone's life easier while we're in a hybrid of replacing existing code with Maker.

landondurnan: Before
CleanShot 2022-01-11 at 17 09 26


After
CleanShot 2022-01-11 at 17 10 23

The text color of the radio and checkbox labels should be the same. In this PR they're not set to the same color, this isn't ideal.  The alternative you're proposing is either all of our components that use text will need to internally leverage the text component or we will need to add color props to every component that has text. Neither sound ideal.

pretzelhammer: >Nearly all design systems...

I don't think using "but other design systems do this" in every discussion as an argument is productive. If other design systems already solve all of our problems then why aren't we just using them?

> Static CSS variables set at the global level provide the easiest, fastest, and cleanest way to style the entire application.

Exactly, thank you for reaffirming my point. Maker is used in the context of the Website codebase, of which at most is only like ~2% Maker right now. However, if we set these styles in MTheme they will theme the ~2% of the app which is Maker BUT they will also bleed into the other ~98% of the app which is not Maker. This is likely going to cause problems.

> The alternative you're proposing is either all of our components that use text will need to internally leverage the text component or we will need to add color props to every component that has text.

The latter is ideal as it protects one of our already agreed upon technical design considerations: components should be able to be used and styled standalone without a parent MTheme component. The current implementation is not ideal because it creates a hard dependency on styling MRadio labels, MCheckbox labels, and MCalendar headings on having a parent MTheme component. There's now a discrepancy between what can be styled at the theme-level and at the component-level for these handful of components. People will get confused and ask _"Why do these components have a textColor prop and these don't?"_ and et cetera.

Anyway, I added these styles back in so I'm resolving this comment thread. If everything else in the PR looks good please approve so we can merge.

laurenhu: I see what you're saying in terms of mixed contexts where Maker is not the only library providing styles. I can say that at least on website, we set background and text color on the global, then block level based on color profile settings. So transferring these behaviors to Maker won't be a departure in terms of rendered output Most website components that don't have explicit color options inherit these colors

laurenhu (approved):

landondurnan (approved):
Approving to push forward. But these types of things are going to keep coming up and I'm going to keep pushing the idea that we should prefix and define all the global tokens and define all styles with CSS rather than defining as props and injecting the styles inline.

The more global theming options we add the more challenges we're going to run into with how to define things globally across multiple components. CSS is the best solution to that.

PR #211, opened 2022-01-19
Revert "refactor(actionbar): Add --chrome-bottom-offset to bottom padding"

landondurnan:
Reverts square/maker#209 as it's not a refactor and instead a _temporary_ fix


pretzelhammer:
you have to update the title to follow the commit message format, so it has to begin with "refactor" (or some equivalent term) and not "Revert refactor"

Issue #201, opened 2021-12-06
Action bar scrolling out of viewport on IOS when referred from another site


wkashdan:
### Bug description

When referred to a site with a modal + action bar in Chrome IOS, the action bar scrolls out of the viewport. 

### Reproduction

1. Click a link to a site that has a Modal + action bar, I used Slack on IOS as a place to refer from
2. In this case I used https://square.github.io/maker/styleguide/latest/#/Modal
3. Launch the modal + action bar, scroll and notice the action bar is covered by the browser bar 

https://user-images.githubusercontent.com/8744227/144893845-66a44a55-ba07-4276-a107-80cb0e7f79a7.MP4

### Environment

```shell
This bug happened on IOS 14.8.1, on an IPhone 12 Mini in Chrome 

Unfortunately I'm not sure what version of Chrome IOS this was happening on. I just updated to the most recent and the bug is not happening anymore.
```

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

landondurnan:
I've been crawling through chromium ios releases to see if there was a recent bugfix in the past couple months. Their release and git commit messages haven't provided a lot of insight yet (looked through the last three releases).

Unfortunately, there's probably a judgement call here in terms of reverting the previous code to work on older versions of Chrome vs leaving things as-is as a won't fix since it's work as expected on current versions. If we can get some more details of when Chrome made the change this will be easier.

landondurnan:
Making note of something here that might be relevant (thanks to @mattmewton) https://medium.com/@draganeror/iphone-x-layout-features-with-css-environment-variables-d57423433dec

> You will find in many sources on the internet that env() will work anywhere var() works, but that is not the case. From experience, sometimes it will not work inside calc(). In that case, use Custom CSS Properties (CSS native variables).

That was one change I made that could be impacting some of the experience (particularly between different versions of iOS Chrome). I was trying to make our css around the padding more succinct and removed a lot of the previous variables by using calc. We could put that back in perhaps to resolve issues where browsers aren't able to `calc`

The article noted above seems like what we should do to be safe.

landondurnan:
On second thought, the iOS versions mentioned in the article are for browsers several years old at this point and several major version behind. Keeping consistent with our `.browserlist` we don't need to fallback that far. We could change the `calc` to `vars`, but it'd be still good to know when Chrome fixed it so we can make note of it.

sohnj877:
https://github.com/square/maker/pull/209

[...]

PR #212, opened 2022-01-19
fix(actionbar): temp fix for chrome ios; add --chrome-bottom-offset to padding

sohnj877:

https://github.com/square/maker/pull/209 had wrong commit type so it was reverted

This reverts https://github.com/square/maker/pull/211 and put back https://github.com/square/maker/pull/209

## Other information

This is a temporary fix for https://github.com/square/maker/issues/201 and `--chrome-bottom-offset` will be set in MTheme in the future

### DIFF
--- a/lab/experiments/ActionBarTransition/ItemModal.vue
+++ b/lab/experiments/ActionBarTransition/ItemModal.vue
@@ -171,6 +171,7 @@ export default {
 </style>
 
 <style module="$s">
+/* tempfix: chrome-bottom-offset - value set outside of maker */
 /* stylelint-disable length-zero-no-unit */
 .ActionBarWrapper {
 	--regular-bottom-padding: 32px;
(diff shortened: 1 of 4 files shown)


lab/experiments/ActionBarTransition/ItemModal.vue
```diff
@@ -180,6 +181,7 @@ export default {
 			var(--regular-bottom-padding)
 			+ var(--extra-bottom-padding-for-deadclick)
 			+ var(--safe-area-inset-padding)
+			+ var(--chrome-bottom-offset, 0px)
```
pretzelhammer, line 184: can you add a comment here, and everywhere else where `--chrome-bottom-offset` is used to explain that this is a temp fix / undefined var that comes from outside of Maker? thanks

pretzelhammer (approved):
looks good, as before, but please add some comments by the usage of the new undefined CSS vars

landondurnan (approved):

PR #213, opened 2022-01-20
docs: Themes lab demo

landondurnan:

We needed a way to test Maker theming in some more "real-world" scenarios. We also have some upcoming work related to additional typography and shape theming that we need to do more discovery with.

CleanShot 2022-01-20 at 11 53 27

This is a hackweek project based on our current Theme beta that does the following:
* Creates a more robust theme editor and canvas preview to test color, typography, and shape.
* Explores support for typography including changing font-family, weights, base size, and a scale control.
* Adds shape support to several components to understand challenges and work on solutions to have more components have the same / similar shape output.
* Outputs theme data as JSON, so it's easy to see and get all the values. Should help designers explore and define individual themes and pass that data to a theme library.

Lab Demo accessible here: https://square.github.io/maker/lab/themes-lab-demo/#/Themes/

## Other information

This is marked as a draft for now as it implements some new things with typography and shape that are not ready to be merged as-is. The hope here is this acts as a helpful reference with some ongoing work and eventually we can update the theme lab playground and ship this lab with that code.

Some more "to do" items that would be nice-to-have:
- Add color profiles, right now this lab only supports a single global profile.
- Add some other color definitions like secondary, link, and any others we support.
- Better colorpicker -- can't copy / paste hex values with browser default control.
- Better font-family and weight definitions, including an understanding of what "pairing" would look like.
- Need to understand how we could specify font fallbacks, particularly between serif and sans-serif fonts.
- Need to figure out how to segment dependencies for the lab environment as I added pinia for the lab, but that shouldn't be a dependency for Maker. (wondering if we should just have separate packages between lab, docs, and the components)


landondurnan:
I'm going to close this out and create a new PR with a cleaner history

PR #214, opened 2022-01-20
fix(theme): fix color fallbacks on form components without MTheme

laurenhu:
This PR fixes the fallback styles for form inputs used without an MTheme wrapper.

Without fallbacks:
Screen Shot 2022-01-20 at 2 24 50 PM

With fallbacks:
Screen Shot 2022-01-20 at 2 25 06 PM


landondurnan (approved):
Nice

PR #215, opened 2022-01-21
feat!: theming typography

pretzelhammer:

no theming typography

adds theming typography:
- adds themeable `weight` prop on MHeading and MText which allows controlling the font weight
- refactors/fixes the `size` prop on MHeading & MText to go from integer number steps to string named steps (the names came from Design)
- adds `baseSize` and `sizeScale` properties to the theme spec which gives users control over the typographical hierarchy used by the MHeading and MText at a theme-level
- updates the MHeading & MText docs to show off all the new features/refactors

## Other information

updated MHeading docs: https://square.github.io/maker/styleguide/theming-typography/#/Heading
updated MText docs: https://square.github.io/maker/styleguide/theming-typography/#/Text

### DIFF
--- a/src/components/Theme/src/default-theme.js
+++ b/src/components/Theme/src/default-theme.js
@@ -19,8 +19,6 @@ export default function defaultTheme() {
 		fonts: {
 			baseSize: '16px',
 			sizeScale: 1.15,
-			baseLineHeight: '1.5em',
-			lineHeightScale: 0.95,
 		},
 		profiles: [
 			{
--- a/src/components/Theme/src/default-theme.js
+++ b/src/components/Theme/src/default-theme.js
@@ -17,7 +17,7 @@ export default function defaultTheme() {
 			primary: '#000000',
 		},
 		fonts: {
-			baseSize: '16px',
+			baseSize: '14px',
 			sizeScale: 1.15,
 		},
 		profiles: [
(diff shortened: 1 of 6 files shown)


src/components/Theme/src/default-theme.js
```diff
@@ -16,6 +16,10 @@ export default function defaultTheme() {
 			overlay: 'rgba(0, 0, 0, 0.3)',
 			primary: '#000000',
 		},
+		fonts: {
+			baseSize: '14px',
+			sizeScale: 1.15,
```
landondurnan, line 21: Changing the base font size and scale is going to have negative consequences on sites that are currently using 16px as the base font size. We should not change these defaults at this time.

landondurnan (changes requested):
After some initial testing of the impact of changing the base font size, this will have negative consequences and increase the amount of work necessary to rectify a change like this.

I recommend we keep the base size, scale, and steps as they currently are to maintain status quo for our output. Let's revisit the primary issues we're trying to solve with a change like this and determine a safer way to support it given the current state of this code and existing code in consuming apps.

pretzelhammer:
alternative PR: https://github.com/square/maker/pull/216

landondurnan:
Is this PR still relevant with https://github.com/square/maker/pull/215 open? Are both meant to be reviewed?

pretzelhammer:
@landondurnan i guess not, i'll close this one

PR #216, opened 2022-01-28
feat!: theming typography compatible with Website

pretzelhammer:

no theming typography

yes theming typography

## Other information

[MHeading docs](https://square.github.io/maker/styleguide/theming-typo-website-compat/#/Heading)
[MText docs](https://square.github.io/maker/styleguide/theming-typo-website-compat/#/Text)

### DIFF
--- a/src/components/Heading/README.md
+++ b/src/components/Heading/README.md
@@ -88,12 +88,31 @@
 		>
 		<br>
 		<br>
+		type scale
+		<br>
+		<br>
 		<m-heading
 			v-for="size in [-2, -1, 0, 1, 2, 3, 4, 5, 6, 7]"
 			:key="size"
 			:size="size"
 		>
-			Size {{ size }}
+			Size {{ size }} -
+			<span
+				class="showSize"
+				:mobileSize="stepToMobilePx(size)"
+				:desktopSize="stepToDesktopPx(size)"
+			/>
+		</m-heading>
+		<br>
+		named sizes
+		<br>
+		<br>
+		<m-heading
+			v-for="namedSize in namedSizes"
+			:key="namedSize.name"
+			:size="namedSize.size"
+		>
+			{{ namedSize.name }}
 		</m-heading>
 	</m-theme>
 </template>
@@ -111,12 +130,58 @@ export default {
 		const defaults = defaultTheme();
 		return {
 			baseFontSize: Number.parseInt(defaults.fonts.baseSize, 10),
-			fontSizeScale: defaults.fonts.sizeScale,
+			fontSizeScale: Number.parseFloat(defaults.fonts.sizeScale),
 			baseLineHeight: Number.parseFloat(defaults.fonts.baseLineHeight),
-			lineHeightScale: defaults.fonts.lineHeightScale,
+			lineHeightScale: Number.parseFloat(defaults.fonts.lineHeightScale),
 			fontFamily: defaults.heading.fontFamily,
 			fontWeight: defaults.heading.weight,
 			textColor: defaults.resolve(defaults.heading.textColor),
+			namedSizes: [
+				{
+					name: 'headline-1',
+					size: 7,
+				},
+				{
+					name: 'headline-2',
+					size: 6,
+				},
+				{
+					name: 'headline-3',
+					size: 5,
+				},
+				{
+					name: 'headline-4',
+					size: 4,
+				},
+				{
+					name: 'headline-5',
+					size: 3,
+				},
+				{
+					name: 'title-1',
+					size: 4,
+				},
+				{
+					name: 'title-2',
+					size: 3,
+				},
+				{
+					name: 'title-3',
+					size: 2,
+				},
+				{
+					name: 'title-4',
+					size: 1,
+				},
+				{
+					name: 'title-5',
+					size: -1,
+				},
+				{
+					name: 'title-6',
+					size: -2,
+				},
+			],
 		};
 	},
 	computed: {
@@ -124,9 +189,9 @@ export default {
 			return {
 				fonts: {
 					baseSize: `${this.baseFontSize}px`,
-					sizeScale: this.fontSizeScale,
-					baseLineHeight: this.baseLineHeight,
-					lineHeightScale: this.lineHeightScale,
+					sizeScale: Number.parseFloat(this.fontSizeScale),
+					baseLineHeight: Number.parseFloat(this.baseLineHeight),
+					lineHeightScale: Number.parseFloat(this.lineHeightScale),
 				},
 				heading: {
 					fontFamily: this.fontFamily,
@@ -136,8 +201,50 @@ export default {
 			};
 		},
 	},
+	methods: {
+		stepToMobilePx(step) {
+			return this.adjustRawPx(
+				Number.parseFloat(this.baseFontSize)
+				* Number.parseFloat((this.fontSizeScale) ** step),
+				step,
+			);
+		},
+		stepToDesktopPx(step) {
+			const DESKTOP_ADJUST = 0.13;
+			return this.adjustRawPx(
+				Number.parseFloat(this.baseFontSize)
+				* ((Number.parseFloat(this.fontSizeScale) + DESKTOP_ADJUST) ** step),
+				step,
+			);
+		},
+		adjustRawPx(rawPx, step) {
+			const MINUS_ONE_STEP = -1;
+			const MINUS_TWO_STEP = -2;
+			const MINUS_ONE_MIN_CLAMP = 14;
+			const MINUS_TWO_MIN_CLAMP = 12;
+			let adjustedPx = Math.round(rawPx);
+			if (step === MINUS_ONE_STEP) {
+				adjustedPx = Math.max(MINUS_ONE_MIN_CLAMP, adjustedPx);
+			} else if (step === MINUS_TWO_STEP) {
+				adjustedPx = Math.max(MINUS_TWO_MIN_CLAMP, adjustedPx);
+			}
+			return `${adjustedPx}px`;
+		},
+	},
 };
 </script>
+
+<style scoped>
+.showSize::before {
+	content: attr(mobileSize);
+}
+
+@media (min-width: 600px) {
+	.showSize::before {
+		content: attr(desktopSize);
+	}
+}
+</style>
 ```
 
 ## Unthemed
--- a/src/components/Heading/README.md
+++ b/src/components/Heading/README.md
@@ -1,6 +1,85 @@
 # Heading
 
-## Themed
+Use the Heading component for heading text. There are 9 heading sizes: -2 to 7. The font family, font weight, and text color are customizable.
+
+```vue
+<template>
+	<div>
+		font family
+		<br>
+		<select v-model="fontFamily">
+			<option value="inherit">
+				inherit
+			</option>
+			<option value="arial">
+				arial
+			</option>
+			<option value="serif">
+				serif
+			</option>
+			<option value="sans-serif">
+				sans-serif
+			</option>
+			<option value="monospace">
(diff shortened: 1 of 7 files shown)


src/components/Heading/README.md
```diff
@@ -1,36 +1,298 @@
 # Heading
 
+## Themed
+
```
landondurnan, line 4: I think the base component documentation should simply be the default options for that component. Theming demos are more appropriate in the the theme component documentation.

src/components/Heading/src/Heading.vue
```diff
@@ -90,9 +100,19 @@ export default {
...
 				color: this.resolvedTextColor,
+				fontWeight: this.resolvedWeight,
+				'--mobile-base-font-size': fonts.mobileBaseSize,
+				'--mobile-font-size-scale': fonts.mobileSizeScale,
+				'--mobile-base-line-height': fonts.mobileBaseLineHeight,
+				'--mobile-line-height-scale': fonts.mobileLineHeightScale,
+				'--desktop-base-font-size': fonts.desktopBaseSize,
+				'--desktop-font-size-scale': fonts.desktopSizeScale,
+				'--desktop-base-line-height': fonts.desktopBaseLineHeight,
+				'--desktop-line-height-scale': fonts.desktopLineHeightScale,
```
landondurnan, line 115: We need a better way of globally setting these values so they can be shared across multiple components. It's inefficient and costly to performance to set so many values inline.

I recommend we move all the css properties that can be set globally across multiple components into one source and finally tackle this issue: https://github.com/square/maker/issues/39

src/components/Theme/src/default-theme.js
```diff
@@ -16,6 +16,16 @@ export default function defaultTheme() {
 			overlay: 'rgba(0, 0, 0, 0.3)',
 			primary: '#000000',
 		},
+		fonts: {
+			mobileBaseSize: '16px',
+			mobileSizeScale: 1.17,
+			mobileBaseLineHeight: '1.5em',
```
landondurnan, line 22: line height units should be baseless, but I'm not sure how that impacts calculations.

landondurnan (changes requested):
Now is the time to remove the code duplication and inline styling on these components. We're now triplicating the same code across multiple components and impacting performance by running this calculations and injecting them inline.

These CSS properties can be scoped and defined globally once and reused without polluting each components rendering and the DOM with inline styles with essentially the same repeated values over and over again.

I think it's also a good idea to move all theming demos to the theme component page. The Heading and Text components are relatively simple and straightforward on their own and the ability to customize them further can largely be summarized in the Theme documentation.

pretzelhammer:
@landondurnan 

> Now is the time to remove the code duplication and inline styling on these components. We're now triplicating the same code across multiple components and impacting performance by running this calculations and injecting them inline.

I knew you would complain about this which is why I brought it up on Slack 2 weeks ago: https://square.slack.com/archives/C0287HVEXD5/p1642616147034000

And you replied with: https://square.slack.com/archives/C0287HVEXD5/p1642617415038800

TL;DR: I asked if we want the components to work without a parent MTheme and you said yes. This implementation is the cost of meeting that requirement.

> These CSS properties can be scoped and defined globally once and reused without polluting each components rendering and the DOM with inline styles with essentially the same repeated values over and over again.

Not if you want MHeadings and MTexts to work without a parent MTheme.

> I think it's also a good idea to move all theming demos to the theme component page. The Heading and Text components are relatively simple and straightforward on their own and the ability to customize them further can largely be summarized in the Theme documentation.

But if we remove the "duplicated" CSS code it becomes impossible to use the MHeading & MText components without a parent MTheme component so it's not possible to do non-themed demos of the components anymore.

Furthermore, even if that wasn't the case, I don't think it's a good idea to cram demos of every component into the MTheme docs because it will become a kitchen-sink-style page where it'll be hard for people to find what they're looking for and it'll be hard for us as devs to maintain it. It's much simpler and easier, for both consumers and maintainers of Maker, if the themed demos go into all of the individual component's docs pages.

src/components/Heading/src/Heading.vue
pretzelhammer: > We need a better way of globally setting these values so they can be shared across multiple components.

We already have a way, it's called the MTheme component.

> It's inefficient and costly to performance to set so many values inline.

I think the performance cost is negligible.

> I recommend we move all the css properties that can be set globally across multiple components into one source

We already do that, it's called the MTheme component.

> and finally tackle this issue: #39

That issue only concerns static CSS. Even if that issue is resolved it would not remove the inline styles used in the MHeading and MText components which are for run-time customizable dynamic CSS and required to be there if the components are meant to be usable without a parent MTheme component.

src/components/Heading/README.md
pretzelhammer: To quote myself from [another comment](https://github.com/square/maker/pull/216#issuecomment-1028356473), just for visibility's sake:

> I don't think it's a good idea to cram demos of every component into the MTheme docs because it will become a kitchen-sink-style page where it'll be hard for people to find what they're looking for and it'll be hard for us as devs to maintain it. It's much simpler and easier, for both consumers and maintainers of Maker, if the themed demos go into all of the individual component's docs pages.

landondurnan: It's far more effective to keep theming and documentation about theming separate from each component. We can create any number of labs from the theme documentation the demonstrates how it all works. The theme is exposing controls for color, fonts, and shape and we already have a "kitchen sink" demo. This is extremely common to have separate theming docs that don't bleed into or overtake the component documentation:

- https://material.io/design/material-theming/overview.html#material-theming
- https://m3.material.io/foundations/customization
- https://www.carbondesignsystem.com/guidelines/themes/overview/
- https://www.carbondesignsystem.com/guidelines/typography/overview/

When you're looking at a specific component the ideal scenario is a clear and static version of that component. What are the common variants / props a person would customize. Make it easy to copy and paste a single use-case. 

We're trying to build Maker so that consumers don't even need to think about theming, so therefore the component documentation should mimic how we believe most people to understand how that component works out-of-the-box. We can link to Theming documentation if it's relevant, but you shouldn't have to learn about Theming to understand a component.

landondurnan:
It's possible to generate CSS that can be shared regardless if those components are used inside a Theme or not. A stylesheet generated with the library would provide us the ability to define global defaults that can be shared across multiple components. This practice is exceptionally common in every component library. The theme component can update any relevant defaults.

Triplicating code across three components isn't necessary and is called out as an the issue you created here: https://github.com/square/maker/issues/39

Are we forever going to avoid creating shared styles? Why not solve this problem now?

laurenhu:
Do we want have a shared base component for MHeading/MText computed font size/styling props and styles? To at least share some computed props and css? I don't think there's a huge risk from duplicating this between just two components, but it might make adjustments and feedback from Design QA once it's integrated into website quicker?

Otherwise, I'm happy to look into #39 - I have some time this week.

pretzelhammer:
@landondurnan the fluid scale has been added, take a look at it and if it looks alright to you then this PR should be good to merge, we can address CSS sharing in another PR.

laurenhu (approved):

src/components/Theme/src/default-theme.js
```diff
@@ -16,6 +16,12 @@ export default function defaultTheme() {
 			overlay: 'rgba(0, 0, 0, 0.3)',
 			primary: '#000000',
 		},
+		fonts: {
+			baseSize: 16,
+			sizeScale: 1.15,
+			baseLineHeight: 1.5,
+			lineHeightScale: 0.95,
```
landondurnan, line 23: We currently have no plans to support either of these being values that can change at the theme level. We should avoid adding values at this level if they're not configurable.

src/components/Theme/src/default-theme.js
```diff
@@ -45,11 +51,13 @@ export default function defaultTheme() {
 			fontFamily: 'inherit',
 			size: 0,
 			textColor: '@colors.text',
+			weight: 300,
```
landondurnan, line 54: The default font weight for `regular` text is 400.

landondurnan (changes requested):
Some suggested changes for the global defaults.

I also still think we need to move all the "Themed" demo for text to the theme component and keep the Heading and Text component documentation agnostic to the Theme. I'm not sure why that conversation was resolved.

src/components/Theme/src/default-theme.js
pretzelhammer: The default for regular text according to Shimaa's design files is 300. I would ask her to clarify but she's out today. I'll ask tomorrow.

landondurnan: Here's a code reference for it: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight and the design file here references 400 for paragraphs: https://www.figma.com/file/SHY7Z7PbcVssjIJVuJ3IwU/%F0%9F%8D%81-Maker?node-id=12421%3A6938

pretzelhammer: i was referring to the actual typography project design files here: https://www.figma.com/file/RvCNKalYNiOB0iyw6jYJM7/Typography-system-%E2%80%94-Shimaa-%E2%80%94-Q421?node-id=603%3A41589

anyway, I talked to Shimaa and she's cool with bumping it to 400.

pretzelhammer:
@landondurnan 

> I also still think we need to move all the "Themed" demo for text to the theme component and keep the Heading and Text component documentation agnostic to the Theme. I'm not sure why that conversation was resolved.

I marked it resolved because I explained my reasons for having the theming demos in the components' styleguide pages instead of in the Theme component's styleguide page [here](https://github.com/square/maker/pull/216#issuecomment-1028356473):

> I don't think it's a good idea to cram demos of every component into the MTheme docs because it will become a kitchen-sink style page where it'll be hard for people to find what they're looking for and it'll be hard for us as devs to maintain it. It's much simpler and easier, for both consumers and maintainers of Maker, if the themed demos go into all of the individual component's docs pages.

I edited the Heading & Text docs pages so that they show unthemed examples first and then the themed stuff is at the bottom.

landondurnan:
I do agree the current documentation in the Beta on Theme is pretty overwhelming. 
https://square.github.io/maker/styleguide/theming-typo-website-compat/#/Theme

Adding more to it would definitely make it feel like a Kitchen Sink. So we should consider revising the Theme documentation in a way that's more consumable. 

Primarily -- we are adding theming so that globally we can control Colors, Fonts, and eventually shape.  The current docs are very technically focused on everything that's possible, but not necessarily a concise way to understand how and what should be changed.  We don't need separate demos for how you "theme" every component. We can have one high level demo that demonstrates the most highly customizable components in a single demo.

We can eliminate the separate demos currently there and the majority of the options we're demonstrating as themeable as they are not relevant to the typical consumer. Let's define what we want to demonstrate both in options and components.

Taking some inspiration from here: https://square.github.io/maker/lab/themes-lab-demo/#/Themes/
CleanShot 2022-02-15 at 12 29 21

The first demo can show these theme variables and a collection of components that leverage them. Something like this:
CleanShot 2022-02-15 at 12 53 47

This contains more than half a dozen components that all leverage the options for colors and typography. We don't need to have separate demos for how theming works for buttons, actionbar buttons, and text buttons -- they all work the same way. What we're trying to get across in the very first demo of every component is "look how easy this is to use".

Additional demos after that can get into the important nuances between each, but generally that first demo should capture most of it.

I've been looking closer at the docs for Heading and Text and have noticed some mistakes and some opportunity to add more clarity in our props table. It will be easier to show in code so I'll submit a PR against this one with some suggested changes.

landondurnan (approved):
Approving the changes for themed typography, but we need to test the line-height calculations some more with the new fluid scale.

We can also tackle the Theme documentation revision in a separate PR.

PR #222, opened 2022-02-09
build(breakpoints): define breakpoints as custom media queries

laurenhu:
Partial solution for issue #39 - specifically addressing standardizing breakpoints across components with custom media queries.


pretzelhammer (approved):
yay, thank you! 🎉

Issue #220, opened 2022-02-08
Pin Code Input


Ryguy11:
### Figma link

https://www.figma.com/file/SHY7Z7PbcVssjIJVuJ3IwU/%F0%9F%8D%81-Maker?node-id=10790%3A5819

### Description

The Pin Code Input component will be created to handle the input of different one time passwords or pin code entries. It will be a dynamic length input that is adjustable via prop. The Pin Code will need to auto focus the next input when a number is entered, and emit the code when completed. The Pin Code Input will need an error state and error styling to let the user know if the Pin they entered is invalid.

### API

Props:
- pinLength: number, default 6 - length of inputs to display
- disabled: boolean, default false - should disable code inputs

Event:
- complete: emit the completed PIN code

Slots:
- error: error slot that will display below the pin input

### Theme support

Given that the Pin Inputs will be a collection of Maker inputs, theme'ing should be applied in a similar manner as the inputs.

landondurnan:
Thanks @Ryguy11 

Primarily we should try to determine if we're going to internally use the Input component and then weigh the costs of either doing it or not.

If not, then we may want to mimic the same API noted here, which I think you've done for the most part:
https://github.com/square/maker/tree/master/src/components/Input#props

Probably need an invalid state. Not sure if we want to support different variants. And we don't need prefix / suffix output. 

If we do use the input component internally:
Pros: much of the styling and props are supports and it will inherit any theming controls we add.
Cons: likely need different events and styling for "pin code" that isn't exactly the same as an input with a password.

Gut feel on this is we should simply duplicate what we need rather than use the input internally.

landondurnan:
@Ryguy11 You had mentioned in passing this component exists externally in a couple other components, would you be willing to create a Draft PR pushing that component into Maker? Include any changes you think necessary to match the API suggestions above -- but depending on bandwidth we can determine resourcing to finalize the code around it.

pretzelhammer:
ryguy11 merged his PR into Maker which closes this issue

PR #223, opened 2022-02-10
feat(pin-input): add pin input component

Ryguy11:
Implement the Pin Input component per the specs outline in this issue: https://github.com/square/maker/issues/220


src/components/Dialog/src/DialogLayer.vue
```diff
@@ -175,6 +175,7 @@ export default {
 }
 
 .disableScroll {
+	position: fixed;
```
Ryguy11, line 178: This is a change outside of the pin-input component, but it prevents unnecessary body scrolling and jumpiness on dialogs when using the pin-input component inside of them.

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,426 @@
...
+			const TIMEOUT_LENGTH_MS = 1000;
+			setTimeout(() => {
+				this.isShaking = false;
+			}, TIMEOUT_LENGTH_MS);
+		},
+	},
+};
+</script>
+
+<style module="$s">
```
Ryguy11, line 219: Styling to maker conventions and for theming are something this PR is lacking - would love some examples or feedback here on convention

src/components/PinInput/README.md
```diff
@@ -0,0 +1,305 @@
+# PinInput
+
+### Test code for following inputs is 123456
```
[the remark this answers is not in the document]
Ryguy11: I can reword it - I actually meant to keep it because I wanted people playing around to be able to have a successful input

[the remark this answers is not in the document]
pretzelhammer: ohhh, i would not making it a heading but just make it a note somewhere on the page, as regular sized text.

usually, at the top of some component docs, you introduce the component and explain it's use-cases. You might say something like "The PinInput component should be used when you want a buyer to input a pin number and have that number validated. Note: in the examples use pin number 123456 to see the component's success state and any other pin number to see the component's failure/error state."

it doesn't have to be exactly that of course but whatever you think is best for introducing people to the component.

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+					:disabled="disabled"
+					:maxlength="i === 0 ? pinLength : 1"
+					:class="{
+						[$s.CodeInput]: true,
+						[$s.filled]: variant === 'fill',
+						[$s.error]: invalid,
+					}"
+					type="text"
+					inputmode="numeric"
+					pattern="[0-9]*"
```
pretzelhammer, line 29: is this i18n friendly? it's okay if the answer is no, but just wanna make sure we're only supporting PINs that are single-digit numbers only and that we're not shipping this to any countries that have PINs with weird stuff like letters or whatever in them. (yes i know the N in PIN stands for Number so common sense would tell us PINs only consist of numbers but I just wanna be extra careful and double-check).

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+				this.focusOnCodeCell(newIndex);
+			}
+		},
+
+		onFocusCode(event, index) {
+			event.preventDefault();
+			this.$set(this.code, index, '');
+		},
+
+		resetcode() {
```
pretzelhammer, line 201: should this be stylized as `resetCode`?

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+<style module="$s">
+.PinInputContainer {
+	display: flex;
+	flex-wrap: nowrap;
+	align-items: center;
+	justify-content: space-between;
+	padding-bottom: 8px;
+}
+
+.CodeInput {
```
pretzelhammer, line 228: can you add
```
box-sizing: border-box;
```
to this class and then change the `width` and `height` styles to this
```
width: 48px;
height: 48px;
```
i think this makes reasoning about the component's size a lot easier rather than looking to see if it has a border and then having to do mental math in our heads to figure out it's size 😄

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+	&:valid,
+	&:hover {
+		border: 2px solid rgba(0, 0, 0, 0.9);
+	}
+}
+
+/**
+ * Wrapper class allows for pin input border size to change without causing other pins to shift
+ */
+.CodeInputWrapper {
```
pretzelhammer, line 267: if you add
```
box-sizing: border-box;
```
to a few other classes is this utility-class still necessary? let's try to get rid of this utility class if we don't need it

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+
+.CodeInput {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	width: 46px;
+	height: 46px;
+	padding: 0;
+	font-weight: 500;
+	font-size: var(--font-step-0-size);
```
pretzelhammer, line 236: this CSS variable is not defined in the `master` branch, it's only in the `beta` branch. Since Maker Theming is still in `beta` I wouldn't worry about this for now, you can set this to `font-size: inherit` if you'd like it to be controllable from the client OR to `font-size: 16px` which is sensible default for now.

src/components/PinInput/src/PinInput.vue
pretzelhammer: Maker Theming is still in beta so you don't have to worry about making this component themeable for now since you're trying to get this into master.

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+			return false;
+		},
+
+		handleComplete(code) {
+			this.$emit('complete', code);
+		},
+
+		onPasteCode(event, index) {
+			event.preventDefault();
+			const pastedValue = (event.clipboardData || window.clipboardData).getData('text');
```
pretzelhammer, line 181: just asking to double-check, but is `(event.clipboardData || window.clipboardData)` a reliable way to get the clipboard data on all of our supported browsers, i.e. Chrome, Firefox, Safari, Edge & all of their mobile versions?

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,308 @@
...
+};
+</script>
+
+<style module="$s">
+.PinInputContainer {
+	display: flex;
+	flex-wrap: nowrap;
+	align-items: center;
+	justify-content: space-between;
+	padding-bottom: 8px;
```
pretzelhammer, line 225: this style adds an awkward unnecessary bottom padding to the component. it's only necessary when an error message is present, right? in that case let's only conditionally apply it when there's an error message and leave it out otherwise

pretzelhammer:
The component is called `PinInput` but all of the internal code refers to a "PIN" as a "code", e.g. `resetCode`, `onFocusCode`, etc which is kinda confusing. Can we rename all of the internal methods, functions, and variables to be `resetPin`, `onFocusPin`, and so on?

src/components/PinInput/src/PinInput.vue
Ryguy11: The pins that our services send are always numbers so I want to say that this is i18n friendly yeah

pretzelhammer:
I've been messing around with the component in [the styleguide](https://square.github.io/maker/styleguide/pin-input/#/PinInput) and just wanna clarify some functional and design requirements:

1. the PinInput never wraps or resizes, even if the container it's in is too narrow to fit all of it. currently it just overflows its container but a lot of containers have `overflow: hidden` on them and in those cases the PinInput becomes truncated and inaccessible. if you're aware of this drawback and are okay with it that's fine i just wanted to double-check that you were aware
2.  if i insert a number in the middle cell and then start inserting numbers from the leftmost cell the middle cell is skipped instead of re-visited. is this intended behavior?
3. if i try entering a non-digit character into a middle cell the cursor jumps back to the leftmost empty cell. is this intended behavior?
4. i see there's a `complete` event on the component but unlike other input components there doesn't seem to be any props that allow me to programmatically set or pre-populate cells with values. I think the `complete` event should be removed and replaced with a `v-model` to make this input component behave consistently with all other input components _unless_ there's a strong reason why not to do it this way that you can provide.
5. PinInput only allows users to insert digits & numbers but returns the results as a string. is this intended?
6. `shakeAndClearInput()` only ever seems to be called when the `invalid` prop is toggled from false to true, in which case, instead of exposing that method from the component, can we just add a watcher to the `invalid` prop and call it automatically when its value toggles from false to true? would it ever be necessary to `shakeAndClearInput()` when the input is not invalid?

src/components/PinInput/src/PinInput.vue
Ryguy11: seems to be reliable yeah

Ryguy11:
> 

1. Yup aware - given that our use-case for pins larger than 6 is none, I felt that it was fine to leave as is for now. If a use-case arises where its essential to have wrapping blocks we can add that later.
2. Intended
3. Intended
4. can you ever really think of a use case where the parent component would need to change the input values or even pre-populate it? I think it simplifies this component to not need to v-model it, nor should there ever really be a use-case or need for that. If there is a need for it you should be using a regular maker input.
5. Yeah intended
6. I thought of this dont worry. The issue here is you can be invalid on multiple entries, watchers only watch for _changes_ so if you fail once and then fail again the watcher would never actually trigger because the invalid state was never reset.

pretzelhammer:
@Ryguy11 

regarding 4.: I can't think of a use-case where a `v-model` would be necessary, I was suggesting it for the sake of following conventions and maintaining consistency with the other input-like components. I'm cool with leaving it out of this component because it's so specialized it's probably not gonna be used in a ton of places so if we need to make a breaking change to this component in the future it's probably not going to create a lot of refactor/migration work for its users so we don't need to nail down the perfect API for it right now.

regarding 6.: ah, i see what you're saying. what if, when a user starts typing a new pin into a input that's in an invalid state, we just toggle the invalid state off as they start typing and then toggle it back on if the complete typing another invalid pin? or would that not work for some design or product reason?

Ryguy11:
To elaborate more on point 6:
I actually spent a fair amount of time trying to make this work as intended and not having to use the parent component to trigger the event. Including storing the previous code tested, the current attempt number, or even using the updated lifecycle hook. However, I kept ending up on buggy behavior on each attempt and figured just having the parent trigger the event is fine.

pretzelhammer:
@Ryguy11 

regarding 6. okay, fair enough, it sounds like you already put in the work trying to remove the method but couldn't find a way to do it without introducing buggy or unexpected behavior so I'm cool with keeping it then.

Ryguy11:
@pretzelhammer I ended up deleting my comment cuz we commented at the same time and I was going to address what you said more, but yeah what you just followed up with is correct.

Ryguy11:
To add in some context regarding things I tried to make the `clearAndShakeInputs` handled on the component itself instead of needing it from the parent:
- using the updated() lifecycle hook, along with keeping a record of what the previous attempted code was, or even what attempt number the input code is currently on. Overall, this method is very computationally expensive as updated is called a lot, and in multiple situations I ended up in infinite loops after multiple input failures. Most attempts regarding adding watchers to the PIN or anything else involved the updated lifecycle component.
- you could also pass another prop to the component called shouldShake, and handle it from the parent, but this ends up being much more verbose than just using the refs method.

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,293 @@
...
+
+	&:focus,
+	&:valid,
+	&:hover {
+		border: 2px solid rgba(0, 0, 0, 0.9);
+	}
+}
+
+.shake {
+	animation: invalidcodeshake 500ms linear;
```
pretzelhammer, line 266: please update this to `invalidpinshake`

src/components/PinInput/src/PinInput.vue
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+.PinInput {
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	box-sizing: border-box;
+	width: 50px;
+	height: 50px;
```
pretzelhammer, line 234: just to double-check, the width and height are suppose to be `50px` and not `48px` right?

Ryguy11: yeah, it was 46px + 2px potential border on each side, solved using the box-sizing you suggested. Ive never used that before so that was good to know

src/components/PinInput/README.md
```diff
@@ -0,0 +1,312 @@
...
+		<m-dialog-content>
+			<m-heading :size="3">
+				Dialog heading
+			</m-heading>
+			<m-pin-input
+				ref="pinInput"
+				:class="$s.padding"
+				:invalid="invalidEntry"
+				variant="outline"
+				@complete="onCodeComplete"
```
pretzelhammer, line 220: can you rename the `onCodeComplete` function in the examples to `onPinComplete`?

pretzelhammer (approved):
just rename a few more things but it's otherwise good to go

PR #224, opened 2022-02-14
fix(dialog): Add a width:100% to the div wrapping the dialog content

JimRottinger:
[A previous PR of mine](https://github.com/square/maker/commit/f054fa01a8f86e96efe118b726c0998458d8f9ee) where I introduced `clickOnCloseOutside` to the dialog component causes a regression in which the dialog was no longer 100% width on mobile. This is because I wrapped the content in div in order to set up a vue `$ref` which broke the styling.

Pretty simple fix I am proposing here. Just set the width to 100% for mobile screens. 

Before:

image

After 

image

image

### DIFF
--- a/src/components/Dialog/src/DialogLayer.vue
+++ b/src/components/Dialog/src/DialogLayer.vue
@@ -18,7 +18,7 @@
 				/>
 				<div
 					ref="dialog"
-					:class="$s.DialogContent"
+					:class="$s.DialogContentWrapper"
 				>
 					<v :nodes="dialogApi.state.vnode" />
 				</div>
@@ -162,6 +162,12 @@ export default {
 @media screen and (--for-tablet-landscape-up) {
 	.DialogLayer {
 		align-items: center;
+	}
+}
+
+@media screen and (--for-tablet-landscape-down) {
+	.DialogContentWrapper {
+		width: 100%;
 	}
 }
 


src/components/Dialog/src/DialogLayer.vue
```diff
@@ -165,6 +165,12 @@ export default {
 	}
 }
 
+@media screen and (--for-tablet-landscape-down) {
```
landondurnan, line 168: We should flip this so it's mobile first. 100% should be the default and the tablet portrait up should be a set width.

landondurnan: See: https://github.com/square/maker/blob/master/src/components/Modal/src/Modal.vue#L118-L124

pretzelhammer, line 168: for CSS maintainability reasons we try to write all mobiles styles as the "default" styles and then progressively add the overrides for the tablet & up resolutions. is there a way we can re-write this style so that we have this at the root level:

```css
.DialogContentWrapper {
    width: 100%;
}
```

and then a media query which overrides it to something else for tablet & up like:

```css
@media screen and (--for-tablet-landscape-up) {
    .DialogContentWrapper {
        width: /* something here, like initial, or reset, or idk */;
     ]
}
```

landondurnan: Hmm, actually the ModalLayer example: https://github.com/square/maker/blob/master/src/components/Modal/src/ModalLayer.vue#L244-L259

landondurnan: We should actually look into internalizing a standard Layer component so we can offset this support into a single component rather than duplicating code and dealing with these issues this PR is looking to resolve. (But that will have to happen later)

pretzelhammer: it's our bad for having a `--for-tablet-landscape-down` in the first place, I'll look into ways we can get rid of that.

pretzelhammer (approved):

Issue #221, opened 2022-02-08
MButton isSingleChild overrides fullWidth prop


tatems:
### Bug description

If an MButton is given only a single node as it's child, _and_ the `full-width` prop is set to true, the button will not set it's width to `100%` as one would expect. As a workaround, I can set `width: 100% !important` on a button that falls in this category.

### Reproduction

https://codesandbox.io/s/trusting-hugle-utif0?file=/src/App.vue

### Environment

```shell
System:
    OS: macOS 12.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 136.85 MB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.18.2 - ~/.nvm/versions/node/v12.18.2/bin/node
    npm: 6.14.5 - ~/.nvm/versions/node/v12.18.2/bin/npm
  npmPackages:
    @square/maker: ^6.5.2 => 6.5.2 
    vue: ^2.6.12 => 2.6.12 
    webpack: ^5.1.3 => 5.1.3 
```
```


### Addressed by

_No response_

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

PR #225, opened 2022-02-15
fix(action bar buttons): allow wrapping text on action bar buttons

laurenhu:
This PR:
- Reduces action bar button text size from 16px to 14px, to align with medium buttons
- Allows action bar button text to wrap before truncating after 2 lines
- Fixes [issue](https://github.com/square/maker/issues/221) where the `isSingleChild` class was overriding the `full-width` prop.
Screen Shot 2022-02-15 at 3 44 41 PM

### DIFF
--- a/src/components/ActionBar/src/ActionBarButton.vue
+++ b/src/components/ActionBar/src/ActionBarButton.vue
@@ -1,12 +1,13 @@
 <template>
 	<button
+		ref="button"
 		:class="[
 			$s.Button,
 			$s[`align_${align}`],
 			$s[`shape_${shape}`],
 			{
 				[$s.fullWidth]: fullWidth,
-				[$s.iconButton]: isSingleChild(),
+				[$s.iconButton]: isSingleChild() && !fullWidth,
 				[$s.loading]: loading,
 			}
 		]"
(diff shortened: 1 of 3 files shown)


src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -155,15 +163,27 @@ export default {
...
 		},
 
+		hasMainAndLabelText() {
+			if (!this.$scopedSlots.information) {
+				return false;
+			}
+			const main = (this.$slots.default || []).filter(
+				(vnode) => vnode.tag || vnode.text.trim().length > 0,
+			);
+			const info = (this.$scopedSlots.information() || []).filter(
```
pretzelhammer, line 181: since we're using this pattern in 3 different places within this component can we refactor it out into its own function? e.g.
```
function getNumberOfContentfulVnodes(vnodes) {
    return (vnodes || []).filter( /* etc */);
}
```

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -317,17 +337,32 @@ export default {
...
+.hasMainAndLabelText {
+	text-align: right;
+
+	& .InformationText {
+		text-align: left;
+	}
 }
 
 .TruncateText {
+	display: -webkit-box;
```
pretzelhammer, line 352: I would put a comment here like "despite the -webkit prefix this style is actually supported by all modern browsers" because -webkit usually only means Safari (_and maybe_ Chrome & Opera) support it but in this case every browser supports this style, including Firefox and Edge. To save ourselves and others potential confusion later hopefully.

pretzelhammer (approved):
left a couple small comments, please address those before merging, but otherwise looks good to me!

PR #227, opened 2022-02-16
docs: Heading / Text Doc Revision

landondurnan:
### Documentation complexity
The documentation for Heading and Text was overly complex to convey what the basic components do on their own. 

### Missing or incorrect documentation
The generated docs for Heading and Text either cited incorrect information or was missing information.

### Prop names
We haven't been consistent with prop names particularly in cases that map directly to a CSS property.

### Simplified Documentation
Each component should document what it does on it's own in the simplest demo possible. Simple examples focused on the most typical output should be the goal for each component. Therefore, higher order concepts around Theming can be isolated and documented separately in the Theme component docs. All theming documentation was removed in favor of consolidation as recommended here: Theming documentation should be separate as recommended here: https://github.com/square/maker/pull/216#issuecomment-1040790471

### Updated documentation that was missing / incorrect
- Updated Heading reference from `h1` to `Heading Elements`
- Removed references to unrelated supported events
- Added `@values` for properties that weren't generated from ranges
- Added links for descriptions using JSDoc format [not yet supported in our doc generation]
- Moved properties so font-family / font-weight are together (similar to css linting property order)

### Updated prop names
Since we're adding some new props we should start to standardize how we name the options we support that provide styling closer to the CSS properties that are being used.

- Renamed `weight` to `font-weight`
- Renamed `text-color` to `color`

### DIFF
--- a/src/components/Text/README.md
+++ b/src/components/Text/README.md
@@ -1,229 +1,44 @@
 # Text
 
-Use the Text component for regular body text. There are 9 text sizes: -2 to 1. The font family, font weight, and text color are customizable.
+Use the Text component for regular body text. There are 10 text sizes: -2 to 7. The font family, font weight, and text color are customizable.
 
 ```vue
 <template>
 	<div>
-		font family
-		<br>
-		<select v-model="fontFamily">
-			<option value="inherit">
-				inherit
-			</option>
-			<option value="arial">
-				arial
-			</option>
-			<option value="serif">
-				serif
-			</option>
-			<option value="sans-serif">
-				sans-serif
-			</option>
-			<option value="monospace">
-				monospace
-			</option>
-		</select>
-		<br>
-		font weight
-		<br>
-		<input
-			v-model="fontWeight"
-			type="range"
-			min="100"
-			max="900"
-			step="100"
-		>
-		{{ fontWeight }}
-		<br>
-		text color
-		<br>
-		<input
-			v-model="textColor"
-			type="color"
-		>
-		<br>
 		<m-text
-			v-for="namedSize in namedSizes"
-			:key="namedSize.name"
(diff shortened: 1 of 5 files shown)


src/components/Text/README.md
```diff
@@ -1,238 +1,48 @@
 # Text
 
-Use the Text component for regular body text. There are 9 text sizes: -2 to 1. The font family, font weight, and text color are customizable.
+Use the Text component for regular body text. There are 10 text sizes: -2 to 7. The font family, font weight, and text color are customizable.
```
pretzelhammer, line 3: actually this was a mistake on my part, there are only 4 text sizes: -2 to 1.

src/components/Text/src/Text.vue
```diff
@@ -23,48 +22,52 @@ export default {
...
 		 */
 		element: {
 			type: String,
 			default: 'p',
 			validator: (element) => ['p', 'span'].includes(element),
 		},
 		/**
-		 * size of text
+		 * Size of text
+		 * @values 7, 6, 5, 4, 3, 2, 1, 0, -1, -2
```
pretzelhammer, line 34: this should be 1, 0, -1, -2

look at MIN_SIZE and MAX_SIZE consts defined above.

src/components/Theme/src/default-theme.js
```diff
@@ -48,8 +48,8 @@ export default function defaultTheme() {
 		text: {
 			fontFamily: 'inherit',
 			size: 0,
-			textColor: '@colors.text',
-			weight: 400,
+			color: '@colors.text',
+			fontWeight: 400,
```
pretzelhammer, line 52: you forgot to also update these prop names for the `heading` config below

PR #232, opened 2022-02-28
feat: text-transform prop for MText & MHeading to uppercase text

pretzelhammer:

no prop to uppercase all text in MText & MHeading (needed for website)

adds `textTransform` prop which supports the values: `inherit`, `none`, `uppercase`

## Other information

nope


landondurnan (commented):
For this and the font-style. Since inherit is the default set on the browser, should we make the default undefined? By making the default inherit we will increase the inline styles when it's likely not needed.

pretzelhammer:
@landondurnan pr updated

Issue #238, opened 2022-03-04
make TextButton more visible in styleguide


ashjtan:
### Feature request

Extract the TextButton documentation out of the Button page and give it its own page. (i.e. Alternative 1)

### Why?

The 'TextButton' documentation is currently under the ['Button' page](https://square.github.io/maker/styleguide/latest/#/Button) in the styleguide because of 
- TextButton currently falling under the Button dir
- the auto styleguide generation [(generate-api-tables)](https://github.com/square/maker/blob/master/build/generate-readme-api-tables/generate-api-tables.js)

making the component difficult to find and the documentation messier.

## 1. move the TextButton completely out of the Button dir, making it a separate component
i.e. a one-off fix 

## 2. change the formatting of the styleguide to make "subcomponents" more visible
i.e. a more holistic solution

### a) use subbullets/dropdowns in the left column to give each subcomponent its own page, rather than putting the documentation for multiple subcomponents in the one page
<img width="275" alt="Screen Shot 2022-03-04 at 12 46 17 PM" src="https://user-images.githubusercontent.com/20190589/156839102-b1effb22-9e3a-4e57-b441-e29af14831a2.png">

```
e.g.
- Button
   - TextButton
- ActionBar
  - ActionBarButton
  - ActionBarLayer
  - InlineActionBar
```

### <a href="#user-content-2b" id="2b">b)</a> add table-of-contents-like list of hash links to any "subcomponents" at the top of the page 
e.g. just above the "Button" title
<img width="1216" alt="Screen Shot 2022-03-04 at 12 54 10 PM" src="https://user-images.githubusercontent.com/20190589/156840098-dcd7dd05-7821-413e-b010-4d31a1acbd9f.png">

- this can be done manually in the readme
- one problem with generate-api-tables is that the tables are at the end after all the examples, so with [Button and TextButton](https://square.github.io/maker/styleguide/latest/#/Button), it's formatted like

```
 - Button examples
 - TextButton examples
 - Button props
 - Button slots
 - Button events
 - TextButton props
 - TextButton slots
 - TextButton events
 ```
so you can't actually hashlink to the "TextButton" section, since it's all split up. if we change it so that the page is ordered component-by-component rather than examples->tables, this is very feasible.

### c) change the UI design of the styleguide to more distinctly split up sections
[...]

PR #205, opened 2022-01-11
feat!: theming

pretzelhammer:
## Describe the problem this PR addresses

no theming

## Describe the changes in this PR

adds theming

## Other information

this PR will remain a draft until this PR is resolved https://github.com/square/maker/pull/177 but once that PR is resolved this PR will be ready for final review

## Migration Guide / Breaking Changes from 8.x to 9.x
- MTheme has changed dramatically, however nobody should have been using MTheme in Maker 8.x or lower since it didn't really do much, but if for some reason you were using it just know it has changed a lot
- MText's `textColor` prop has been renamed to `color`
- MHeading's `textColor` prop has been renamed to `color`
- MText & MHeading are now responsive to screen width, compared to 8.x the 9.x sizes are now: slightly larger from 300px - 600px and slightly smaller from 600px - 1200px

PR #249, opened 2022-03-11
refactor(text-button)!: extract TextButton out of Button

ashjtan:
See #238. The `TextButton` was in the same dir as `Button`, making it difficult to find, and muddying up the the docs for both.

- extracted `TextButton` out of `Button` and into its own dir (and consequently, styleguide page)
- added a link in `Button` docs to `TextButton` for better discoverability
Screen Shot 2022-03-10 at 4 29 33 PM

- cleaned up some of the TextButton docs


src/components/Button/README.md
```diff
@@ -1,4 +1,5 @@
 # Button
+For a simplified text-only button with no padding or focus/active/hover states, see: [TextButton](#TextButton).
```
ashjtan, line 2: note: this currently does not work; it changes the link correctly in the address bar, but it doesn't actually route you to the TextButton doc. i would like to open a new issue to implement the ability to link between pages in the styleguide; it would also apply particularly to the "table of contents" idea i mentioned in #238 "2b".

lmk what you think i should do with this for now, e.g.
- leave as-is (once clicked, you can refresh to get to the TextButton page)
- remove the link but maybe bold `TextButton` instead
- open an issue to "implement the ability to link between pages in the styleguide" as mentioned above
- etc. etc.

pretzelhammer: Change the link from `#TextButton` to `#/TextButton` and it works.

pretzelhammer: Change the link from #TextButton to #/TextButton and it works.

```suggestion
For a simplified text-only button with no padding or focus/active/hover states, see: [TextButton](#/TextButton).
```

pretzelhammer: i made the commit

pretzelhammer:
@ashjtan you originally worked on TextButton, right? Can you remind we why we originally put both components in the same directory? Or was it just an arbitrary decision?

pretzelhammer:
@ashjtan whenever making a breaking change you should write a Migration Guide at the bottom of your PR description like [shown in this PR here](https://github.com/square/maker/pull/205#issue-1099432918). I've added it for you this time but please review it and update it if I got anything wrong or missed anything.

src/components/Button/README.md
ashjtan: 🙏 🙏 TIL

pretzelhammer (approved):

laurenhu (approved):

PR #251, opened 2022-03-15
docs: add discouraging comments re explicit CSS values

bashunaimiroy:

@laurenhu raised a good point in [another PR](https://github.com/squareup/ecom-website/pull/11816/files/63e7bc363f6f924401dfa34a14c3c88c7cb21a67#r826349450) using the recently added `fontSize` prop:

`I don't want developers who are not aware that these are reasoned exceptions to start passing in css font size values in places where type scale and base font size should make a impact.`

This simply adds some documenting comments, to raise dev awareness that while you can use this to pass in CSS font size values, this essentially disables type scaling on the component.

Adds info to prop documentation comments for Heading and Text. Note, they are somewhat terse because ESLint is enforcing 100character max-length, and splitting the comment to multiple lines messes with the API Table generation. If it's too terse, we can disable the ESLint rule there.


bashunaimiroy:
thanks for PR title change @pretzelhammer

pretzelhammer (approved):

PR #253, opened 2022-03-16
fix(dialog): fix position bug on desktop dialog layer

laurenhu:
Fixes dialog layer bug on desktop. Typically, adding `position: fixed` causes the body to collapse if its contents are not as wide as the viewport. This `position: fixed` style is needed on mobile devices in order to use Dialog with the PinInput component but this resets the position on desktop.

### DIFF
--- a/src/components/Dialog/src/DialogLayer.vue
+++ b/src/components/Dialog/src/DialogLayer.vue
@@ -191,4 +191,11 @@ export default {
 	position: fixed;
 	overflow: hidden;
 }
+
+@media (--for-desktop-up) {
+	.disableScroll {
+		position: initial;
+	}
+}
+
 </style>


pretzelhammer (approved):

landondurnan:
@laurenhu What makes dialogs unique as compared to modals -- should we do the same thing to modals, or simple remove the position fixed on this class as we do here:
https://github.com/square/maker/blob/master/src/components/Modal/src/ModalLayer.vue#L240-L242

laurenhu:
@landondurnan `position: fixed` was not included for any of the takeover layers (Blade, Dialog, Modal) initially because it scrolls back to the top of the page when `disableScroll` is applied. However, as noted in the comment:
```
/**
 * Position fixed prevents unnecessary body scrolling and jumpiness on dialogs
 * when using the pin-input component inside of them on IOS devices
 *
 */
```

It was added to address a very specific issue in the Pin input + Dialog use case. @Ryguy11 explored many different solutions to the jumpiness on input focus issue and this was the best one. I also explored several different solutions after noticing the issue on `beta` and similarly didn't find a better solution.

PR #250, opened 2022-03-15
fix(button): makes button disabled while it's loading

pretzelhammer:
## Describe the problem this PR addresses

https://github.com/square/maker/issues/248
makes button disabled while loading

## Describe the changes in this PR

makes button disabled while loading

## Other information

nope

Issue #248, opened 2022-03-10
disable Button when loading


ashjtan:
### Feature request

Make Button unclickable when `loading` prop is true.

### Why?

Currently, when the `loading` prop is set, the button is still clickable. Event handlers that are fired from a button click can continue to be fired while the button is loading. I can't really think of instances where we would intend for the button to be clicked while it is shown as 'loading'. 

### Alternatives

No change; any instance where it is important to disable the button while loading can be handled individually by the users, i.e.
```
<m-button
   :loading="isLoading"
   :disabled="isDisabled || isLoading"
>
```

kevinlee11:
This also applies to the `m-text-button` (not sure if the logic comes from the same spot), noticed the same issue.

pretzelhammer:
Like you said, this can readily be handled by users by setting the disabled state if the loading state is active to make the button unclickable, e.g.:

```vue
<m-button
    :loading="isLoading"
    :disabled="isDisabled || isLoading"
>
```

Another alternative is to still allow clicks but to not perform any action if the loading state is active, e.g.:

```vue
<m-button
    @click="onClick"
    :loading="isLoading"
    :disabled="isDisabled"
>

onClick() {
    if (this.isLoading) return;
    // do click action stuff here
}
```

> I can't really think of instances where we would intend for the button to be clicked while it is shown as 'loading'.

Hmmm, what if clicking on a loading button cancels the loading action? Or what if clicking on a loading button opens a toast which shows the current loading progress using a progress bar?

These are uncommon scenarios for sure, but I'm not sure if I want to entirely rule them out by enforcing a disabled state when the loading state is active _at the library level_. As demonstrated above this is already trivially enforceable _at the user level_.

I guess my question is: is this really that common of a gotcha that trips up Maker users that it would be valuable for us to hardcode this behavior into the Button component itself?

kevinlee11:
I feel like the flip side ends up being worse in terms of sending API requests when there's still one in progress. Easy for a FE dev to forget to test the scenario of spamming a button and could result in some unintended behaviour.

Could expose functionality to allow disabling that functionality in case of any edge cases where we would want to allow clicking in the loaded state? Just seems to me we should be targeting the 90/10 use cases.

ashjtan:
@pretzelhammer thanks for pointing out some examples! given there are scenarios where you may want the loading button to be clickable, i agree that it's not functionality we want to _entirely_ rule out. 

[...]

PR #255, opened 2022-03-16
fix(action-bar-button): make ActionBarButton disabled when loading

ashjtan:
See #248.

- Makes ActionBarButton disabled when loading. This was already added to Button and TextButton in #250.
- also added a loading example in the ActionBar styleguide
Screen Shot 2022-03-16 at 2 30 08 PM

### DIFF
--- a/src/components/ActionBar/README.md
+++ b/src/components/ActionBar/README.md
@@ -332,6 +332,17 @@ export default {
 					</m-action-bar-button>
 				</m-inline-action-bar>
 			</div>
+			<div>
+				<m-inline-action-bar>
+					<m-action-bar-button
+						key="confirm"
+						loading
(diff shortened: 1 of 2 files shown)


src/components/ActionBar/README.md
```diff
@@ -332,6 +332,17 @@ export default {
 					</m-action-bar-button>
 				</m-inline-action-bar>
 			</div>
+			<div>
```
ashjtan, line 335: this can be removed if not needed

pretzelhammer (approved):

src/components/ActionBar/README.md
pretzelhammer: i think it's fine to keep

Issue #226, opened 2022-02-15
inputs missing focused state


shimaasq:
### Bug description

Input fields are designed to have 1px border on hover and 2px border when focused. in the styleguide currently they only have hover state. I would like it to be updated with a focused state. if borders change the size of the element, shadows are also possible solution. 

the benefit of having hover and focused states is that they give the impression that the experience is fluid and buyer is in control of it, which create trust in the system. I understand that hover states are desktop centric but we still have big % of users on desktop that we want to serve good exp to.

Link to [design](https://www.figma.com/file/SHY7Z7PbcVssjIJVuJ3IwU/%F0%9F%8D%81-Maker?node-id=3710%3A5816)

![image](https://user-images.githubusercontent.com/75693425/154153787-d2198245-0d73-4d92-b9f8-7115154280ca.png)

### Reproduction

[n/a](https://square.github.io/maker/styleguide/pin-input/#/Input)

### Environment

```shell
https://square.github.io/maker/styleguide/pin-input/#/Input
```

### Addressed by

not urgent

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

pretzelhammer:
Hmmm, I investigated this a bit and I think I've narrowed it down [to this line as part of this release]( https://github.com/square/maker/compare/v4.0.1...v4.1.0#diff-43302d1852d8010ee012aab4dda2d01abb664e11f73617fe625f2ba112a958edR3)

Focus states were implemented using `:focus-within` which worked for all supported browsers at the time, up until Maker v4.1.0 when `>0.5%` was added to our supported browsers list which added browsers like UC Browser and Opera Mini which do not support `:focus-within` which basically silently broke focus states _for all browsers_ because postcss refused to process `:focus-within` anymore since it wasn't part of all our supported browsers anymore.

I think we can fix this by adding `not UCAndroid 12.12` and `not OperaMini all` to our browserslist.

pretzelhammer:
input focus states are back: https://square.github.io/maker/styleguide/8.0.0/#/Input

PR #256, opened 2022-03-17
fix(input)!: focus states by dropping support for UC & Opera Mini

pretzelhammer:

https://github.com/square/maker/issues/226#issuecomment-1071061425

Drops support for UC Browser 12.12 and Opera Mini browsers so that our `:focus-within` CSS styles within the MInput component work again.

## Other information

This has been broken since v4.1.0 when [this line](https://github.com/square/maker/compare/v4.0.1...v4.1.0#diff-43302d1852d8010ee012aab4dda2d01abb664e11f73617fe625f2ba112a958edR3) was released 🤦

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,3 +1,12 @@
 # ECOM supported browsers from https://go/EcomBrowserSupport
- 
-> 0.5%, last 2 Edge versions, last 2 Firefox versions, last 2 Chrome versions, last 2 Safari versions, last 2 iOS versions, last 2 ChromeAndroid versions, not IE 11
+
+> 0.5%
+last 2 Chrome versions
+last 2 ChromeAndroid versions
+last 2 Edge versions
+last 2 Firefox versions
+last 2 iOS versions
+last 2 Safari versions
+not IE 11
+not UCAndroid 12.12
+not OperaMini all
--- a/package-lock.json
+++ b/package-lock.json
@@ -4722,9 +4722,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001248",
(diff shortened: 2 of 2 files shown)


landondurnan (approved):
Browserlist appears to work differently than I had expected. I didn't think it would remove unsupported code, rather I thought it'd simply warn of unsupported code.

It seems a shame to delist browsers because of a single CSS pseudo-element. We could target based on that pseudo-element, but then we're setting ourselves up to be far more granular about every new feature we want to use. 

It might be worth starting a discussion about how we could try to mitigate these types of issues in the future. I'll start one soon.

landondurnan:
@pretzelhammer Have we considered switching the `focus-within` to `focus`?

pretzelhammer:
@landondurnan we could maybe switch `focus-within` and use a combination of JS and `focus` to get a similar-ish effect but I'd rather just drop support for browsers that we never really meant to support in the first place.

According to our pubsite usage stats that Lokesh compiled a while back 0% of traffic comes from UC Browser and only 0.5% of traffic comes from Opera (and that's for _all_ Opera versions, Mini (for mobile devices) _and_ regular (for Desktop)). For reference, we dropped support for IE when it was 3% of our pubsite traffic.

Also, I was part of the discussion when we added the `> 0.5%` figure to our browserlist and the main reason it was added was to support older Chrome versions. We get significantly more pubsite traffic from older Chromes (more than 2 major versions old) than we get from UC and Opera combined. Older Chrome versions account for something like ~10% of traffic, and like I said before, UC and Opera combined are 0.5% (with 0% from UC and Opera usage going down every year).

PR #262, opened 2022-03-21
feat(row): row component

laurenhu:
Row component for use with Menu.

Screen Shot 2022-03-21 at 4 22 35 PM

### DIFF
--- a/src/components/Row/README.md
+++ b/src/components/Row/README.md
@@ -1,0 +1,123 @@
+# Row
+
+```vue
+<template>
+	<div class="wrapper">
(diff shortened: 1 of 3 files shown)


landondurnan (changes requested):
This all looks good, but I think we should keep this internal only until we've had the chance to test it in some other components. There's been interest in using it for the checkbox / radio as well as a new list component. There's a chance we decide to change the API more and that'll be easier if we hold off on releasing this as a public component.

You could move this to the util directory, remove the index.js, but still import directly in Menu.

PR #263, opened 2022-03-23
feat(text): add support for all text sizes

bashunaimiroy:

Website sometimes needs to render a text element with large font size, which semantically speaking is still a paragraph not a heading. Previously we were restricting MText to only accept sizes `-2` through `2`, and anything larger had to be rendered by MHeading, which [isn't best practice](https://webaim.org/techniques/semanticstructure/#:~:text=Use%20headings%20only%20when%20they%20represent%20following%20content.)

Let MText accept all sizes

## Other information

Thanks


pretzelhammer:
please add `required: true` to the `element` prop

pretzelhammer:
also, to support sizes up to 7 you have to add all of these CSS styles to MText which currently aren't present: https://github.com/square/maker/blob/master/src/components/Heading/src/Heading.vue#L207-L272 and https://github.com/square/maker/blob/master/src/components/Heading/src/Heading.vue#L334-L357

landondurnan:
I don't think there's anything inherently wrong with having the MText component support all sizes. In many ways I think we're getting closer and closer to simply merging the Heading and Text component together into a single component since the similarities are far more than they are different.

This is more "back of the house", but I'm surprised to hear that there's some existing cases where that range of sizes don't work. Is there an internal doc you can share that outlines those scenarios?

bashunaimiroy:
> This is more "back of the house", but I'm surprised to hear that there's some existing cases where that range of sizes don't work. Is there an internal doc you can share that outlines those scenarios?

Totally. This [PR description](https://github.com/squareup/ecom-website/pull/12006) explains it and outlines some cases. We can discuss more internally as well if you like

bashunaimiroy:
> please add `required: true` to the `element` prop

image

@pretzelhammer but on a more serious note, subsequent to our internal discussion: I'm adding a default value for `element` when it's `null` in the render method. This is necessary because `null` values don't get run through the Prop validator.

src/components/Text/README.md
```diff
@@ -31,18 +31,18 @@ export default {
...
-| text-transform | `string` | `'inherit'` | `inherit`, `none`, `uppercase`                                | text transform                                                                              |
-| text-align     | `string` | `'inherit'` | `inherit`, `left`, `right`, `center`                          | text align                                                                                  |
+| Prop           | Type     | Default            | Possible values                                               | Description                                                                                 |
+| -------------- | -------- | ------------------ | ------------------------------------------------------------- | ------------------------------------------------------------------------------------------- |
+| element        | `string` | `DEFAULT_TAG_NAME` | `p`, `span`, `div`, `li`                                      | HTML Element wrapper                                                                        |
```
bashunaimiroy, line 36: Not sure if this is preferable, or there is a way to get the auto-documenter to not do this. I extracted `p` into a const since we're using it in two places now and it should be consistent, but I can revert that if it's not worth this headache

src/components/Text/src/Text.vue
```diff
@@ -144,6 +145,9 @@ export default {
 		 * @slot text content
 		 */
 		const defaultSlot = this.$slots.default;
+
+		const element = this.element === null ? DEFAULT_TAG_NAME : this.element;
```
pretzelhammer, line 149:
```suggestion
		const element = this.element ? DEFAULT_TAG_NAME : this.element;
```

src/components/Text/README.md
pretzelhammer: idk how to make the auto-documenter to handle this correctly, let's just stick to using `'p'` literals instead of a const variable

pretzelhammer (approved):

Issue #257, opened 2022-03-17
Dialog on mobile width has double scrollbar sections


kevinlee11:
### Bug description

If you have a dialog open in a mobile width scenario (i.e. where it appears full width at the bottom) it ends up displaying double scrollbar sections.

![image](https://user-images.githubusercontent.com/18740390/158879694-32be5651-d154-44d6-a389-a4815698baef.png)
![image](https://user-images.githubusercontent.com/18740390/158879959-b0873504-56f2-4ed3-9a32-2e9cda353e0b.png)

### Reproduction

Open a dialog in a browser that has mobile width.

### Environment

```shell
System:
    OS: macOS 11.6.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 1.80 GB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.16.2 - ~/.nvm/versions/node/v12.16.2/bin/node
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.2/bin/npm
  npmPackages:
    @square/maker: ^7.0.0 => 7.0.0 
    vue: ^2.6.14 => 2.6.14 
    webpack: ^5.37.0 => 5.38.1
```

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

pretzelhammer:
Hmm, I feel like something might have changed with how Chrome handles `overflow: scroll;` recently because I'm seeing this bug even [back in Maker v4.3.0 when the Dialog component was first released](https://square.github.io/maker/styleguide/4.3.0/#/Dialog) and there's no way we would have released it with such an obvious bug.

pretzelhammer:
oof, the bug has even backported itself all the way to [Maker v2.0.1 to the Modal component](https://square.github.io/maker/styleguide/2.0.1/#/Modal) and back to [Maker v2.1.0 to the Blade component](https://square.github.io/maker/styleguide/2.1.0/#/Blade)

landondurnan:
I'm having trouble replicating the issue myself. Which browsers does this happen in and does it require different OS settings for scrollbar behavior?

pretzelhammer:
pr merged

PR #273, opened 2022-03-29
fix: double scrollbars on Dialog, Modal, Blade

pretzelhammer:

https://github.com/square/maker/issues/257

changes `overflow: scroll` to `overflow: auto` and does small refactor in Dialog to fix other small issues that crept in that we didn't notice earlier.

## Other information

nope

### DIFF
--- a/src/components/Blade/src/Blade.vue
+++ b/src/components/Blade/src/Blade.vue
@@ -17,7 +17,7 @@
 
 .Blade {
 	height: 100%;
-	overflow: scroll;
+	overflow: auto;
 	background: #f5f6f7;
 }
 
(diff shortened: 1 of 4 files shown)


landondurnan:
@pretzelhammer I was trying to test this against what's been deployed and I'm having trouble reproducing the issue unless I change my OS scrollbar behavior to always show scrollbars. Your code looks good -- but I'm not able to verify the original issue against this.

landondurnan (approved):
Not able to verify the original issue -- but tested this and it all works correctly. The code is also inline with expected styling for overflow and scroll behavior on our containers.

Issue #269, opened 2022-03-28
'text-style'/'variant' prop in Maker Text / Maker Heading


bashunaimiroy:
# Feature request

## What
A prop in the Maker Text / Maker Heading components, called `text-style`, `variant`, `text-type`, or any name that would be most clear (probably should avoid something conflated with a CSS property like `font-style`, let's discuss)

**Prop Values**: The value would refer to one of the four Text Styles([Figma link](https://www.figma.com/file/RvCNKalYNiOB0iyw6jYJM7/Typography-system-%E2%80%94-Shimaa-%E2%80%94-Q421?node-id=860%3A73614)) e.g. "headline", "title", "body", "label" that have a Font Weight and Font Family set via the Editor. 

**Behaviour**: If we pass one of these values in the `text-style` prop, e.g. 
```
<m-text text-style="headline"> headline here </m-text>
``` 
the MText component will internally resolve a Font Family and Font Weight for the passed value. In this case, Headline Font Family and Headline Font Weight.

These properties would be resolved within MText/MHeading using the MTheme object, similarly to how they're being resolved now ([github link](https://github.com/square/maker/blob/beta/src/components/Text/src/Text.vue#L111)). e.g. 

```
...resolveThemeableProps(this.textStyle, ['size', 'fontFamily', 'fontWeight', 'color']),
```
^ this is a draft, size and color will probably be resolved differently.

### Background
The ECOM Website app is moving towards offering sellers four Text Styles to customise. eg:
* Headline
* Title
* Body
* Label

The seller will customise these in the Global Font Settings panel ([figma link](https://www.figma.com/file/RvCNKalYNiOB0iyw6jYJM7/Typography-system-%E2%80%94-Shimaa-%E2%80%94-Q421?node-id=860%3A73614)) with a font family and font weight for each Text Style.

[...]

PR #278, opened 2022-03-31
feat!: user-defined theme variants for mtext & mbutton

pretzelhammer:

- https://github.com/square/maker/issues/269
- no user-defined theme-level variants possible

- removes MHeading and merges functionality into MText
    - mostly just makes it possible to create MTexts using `h1`-`h6` tags
- updates all labs / styleguide pages to use MText instead of MHeading
- adds support for user-defined theme-levels variants to theming system in Maker, initially implemented just for MText
    - the default variants for MText are `headline`, `title`, `body`, and `label`
- a bunch of MText refactors & improvements:
     - adds themeable `variant` prop
     - makes `element`, `fontStyle`, `textTransform`, and `textAlign` themeable
     - creates a new `cssValidator` util which is a factory function which produces validators for props which are just CSS styles, and updates the `fontFamily`, `fontWeight`, `fontSize`, `lineHeight`, `color`, `fontStyle`, `textTransform`, and `textAlign` props to use `cssValidator`. changes `fontWeight` and `lineHeight` from Numbers to Strings to accommodate wider range of now supported CSS values.
     - improves docs, adds more links to MDN
     - minimizes use of inline styles by only setting non-default/non-`inherit` styles

Bulk of the work in done in these key files:
- [Theme/src/default-theme.js](https://github.com/square/maker/pull/278/files#diff-b34aee85b50fd2933be427ceddd7edcdb6c5b6eec78d8ae56cbe88838d98df68)
- [Theme/src/utils.js](https://github.com/square/maker/pull/278/files#diff-d37ced2f50eb2fb5c3076ec544ad5f7d854de83bf4e61b754858e7e326039f5c)
- [Text/src/Text.vue](https://github.com/square/maker/pull/278/files#diff-39c2254ebba624c8563c6104f0c04eb3ccc5cdc7ca13ff72d0ca7bc2f9abb82a)

Theming MText (& MHeading) before this PR:
```js
text: {
	// default theme for texts
	fontFamily: 'inherit',
	size: 0,
	color: '@colors.text',
	fontWeight: 400,
},
heading: {
	// default theme for headings
	fontFamily: 'inherit',
	size: 2,
	color: '@colors.heading',
	fontWeight: 700,
},
```

Theming MText after this PR:
```js
text: {
	// default theme for all texts
	size: 0,
	fontFamily: 'inherit',
	color: '@colors.text',
	element: 'p',
	fontWeight: 'inherit',
	fontStyle: 'inherit',
	textTransform: 'inherit',
	textAlign: 'inherit',
	// default variant for all texts is body
	variant: 'body',
	variants: {
		headline: {
			// default theme for headline texts
			size: 4,
			fontWeight: '700',
			element: 'h1',
			color: '@colors.heading',
		},
		title: {
			// default theme for title texts
			size: 2,
			fontWeight: '700',
			element: 'h2',
			color: '@colors.heading',
		},
		body: {
			// default theme for body texts
			size: 0,
			fontWeight: '400',
			element: 'p',
		},
		label: {
			// default theme for label texts
			size: -1,
			fontWeight: '700',
			element: 'h6',
			textTransform: 'uppercase',
		},
	},
},
```

Making Headlines, Titles, Body text, and Label text before this PR:
```vue
<template>
	<m-heading :size="4">headline</m-heading>
	<m-heading>title</m-heading>
	<m-text>body</m-text>
	<m-heading element="h6" :size="-1">label</m-heading>
</template>
```

Making Headlines, Titles, Body text, and Label text after this PR:
```vue
<template>
	<m-text variant="headline">headline</m-text>
	<m-text variant="title">title</m-text>
	<m-text>body</m-text>
	<m-text variant="label">label</m-text>
</template>
```

## Migration Guide

- refactor all `<m-heading>`s in your code to `<m-text variant="title">`
- margin by default is now zero'd on all mtexts, which only affects mtexts that use the default `p` element, so you may have to add margin back to those components in your app


landondurnan:
Before merging this breaking change we should understand that there's existing gaps with our theme output when it comes to our default font-family and weights. Every other component we have that will leverage the font family, weights, and sizes will also need to be updated. Buttons, forms, etc all should reflect the default output as well and it doesn't seem likely that we will want to or in some cases even can embed a text component into every other component that needs to have the same styling.

While this PR resolves the larger issue we have with regular heading and text elements -- we need to think more holistically here. If a breaking change is inevitable we need to at the very least do it as one major release since each one of these will increasingly create more disruptions as we work on other non-breaking changes.

src/components/Theme/src/utils.js
```diff
@@ -48,24 +48,76 @@ function capitalizeFirstLetter(string) {
...
+				// default theme value
+				const themeValue = this.theme[componentKeyInTheme][propName];
+				if (themeValue) {
+					valueOrPath = themeValue;
+				}
+
+				// variant value
+				const hasUserDefinedVariants = !this.$vnode.componentOptions
+					.Ctor.extendOptions.props.variant?.validator;
+				if (this.resolvedVariant && hasUserDefinedVariants) {
```
bashunaimiroy, line 100: This part I don't understand- It's my understanding that the `validator` is a method that should be called. It looks here like we're just checking for the existence of the method?  Why is that

bashunaimiroy:
I took another look and overall this would be super helpful, especially the behaviour where passing `inherit` for a style-related prop (e.g. `color="inherit"`) will prevent it from getting added to inline styles. This addresses my concern [here (slack link)](https://square.slack.com/archives/C02NYVDM0JU/p1648662238913339), although mostly I see why we'd prefer the "friction" factor of using the `color` prop explicitly and I'll be favouring that. 

So that specific improvement in this PR would make my Text to MText Conversion work easier ([internal github link](https://github.com/squareup/ecom-website/pull/12057/files)), and the text variant improvements in general would make everything better. I don't have an opinion on the breaking change of removing MHeading, but it won't prevent my work. I endorse this. I added a minor code question, otherwise I'm ready to approve this PR.

src/components/Text/src/Text.vue
```diff
@@ -169,7 +207,15 @@ export default {
...
 <style module="$s">
-.Paragraph {
+.Text {
+	margin: 0;
+	color: inherit;
+	font-weight: inherit;
+	font-family: inherit;
+	font-style: inherit;
+	text-align: inherit;
+	text-transform: inherit;
```
landondurnan, line 217: Setting the margin to zero may be unexpected. Since we're agnostic to a normalize existing -- the expectation here is that additional external style could be styling text output, particularly with margins. But changing this to margin: 0 now could create unintended consequences.

Also why do all these need to be explicitly defined with inherit?

landondurnan (changes requested):
Most of this looks good. The variant system looks like it'll provide a lot more flexibility and control, particularly down the road.

I had some explicit questions and feedback inline.

As of right now, releasing major changes can be pretty disruptive since there's a gap between when we know these changes can be released and the impact on consumers. Since we don't have a way right now to release non-breaking bug fixes to multiple major releases with our semantic release jobs, I'd recommend one of two things to get this merged so others can start using it.

1. Don't delete MHeading yet. We can keep the component as-is and not add additional features.
2. Release this as a beta until we get closer to a point we know it won't be disruptive with other ongoing work.

Either way we should do a branch or pre-release of this work so we can test it in some more places.

landondurnan:
Looks like there is an issue with the markdown table generation:

CleanShot 2022-03-31 at 14 19 32

The component also looks to have default element and sizes, but those don't show up in the markdown table as defaults.

laurenhu (approved):
This is great and solves for the issue of how to deal with theming for button variants. Right now in Website all solid (primary) Maker buttons are themed but all outline (secondary) buttons need a custom props passed to them. This would also allow us to remove all the ugly code in ActionbarButton for supporting the default style of the close button (by assuming all icon only buttons are close buttons) and adding an actual `close` or `cancel` variant for actionbar buttons.

src/components/Theme/src/utils.js
pretzelhammer: this is an internal maker implementation detail, it doesn't really affect the public api, but if you're curious the reason it was written this way is because some components resolve variants internally and others resolve it from the theme, and the validation method can be different for each scenario, i updated this code in my most recent commit anyway

src/components/Text/src/Text.vue
pretzelhammer: I imported this style from MHeading, [which has had `margin: 0;` since 2020, although the earliest commit in this repo is from jan 2021](https://github.com/square/maker/blob/458fd80b45541f6ba4598cecb10f7b213f20cf4a/src/components/Heading/src/Heading.vue#L110).

I tested the behavior trying out all of the different tags and spans and divs don't have any margin by default, neither do lis, and we explicitly and intentionally removed it from h1s - h6s, so it makes ps the only exception. While this is technically a breaking change, I think it makes the behavior of all of the different element tags a user can select on a MText more consistent (i.e. none of them provide their own margin) which is better.

I've removed the explicit inherits.

pretzelhammer:
@landondurnan 

> Since we don't have a way right now to release non-breaking bug fixes to multiple major releases with our semantic release jobs.

I think we might have a way to do this. Semantic release has [documentation](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/release-workflow/maintenance-releases.md) on how to do this. I'll experiment backporting a v9 bugfix to v8.

> 2. Release this as a beta until we get closer to a point we know it won't be disruptive with other ongoing work.

Okay sure, I changed the base of this PR from `master` to `beta`.

src/components/Text/src/Text.vue
```diff
@@ -169,7 +203,9 @@ export default {
 </script>
 
 <style module="$s">
-.Paragraph {
+.Text {
+	margin: 0;
```
landondurnan, line 207: This will have unintended consequences. I'd recommend changing the classes so the margin: 0 only happens with the headline, title (and maybe label) variants, but regular text has had margins and this would remove a lot of default spacing on those elements.

landondurnan:
I have concerns that we keep changing the way we're passing theme data. Variants will be useful, but we should support a global definition with the fonts.

Our current data looks something like this:
```
"heading": {
    "fontFamily": "Karla",
    "fontWeight": "600"
  },
  "text": {
    "fontFamily": "Domine",
    "fontWeight": "400"
  },
  "fonts": {
    "baseSize": "18",
    "sizeScale": "1.327"
  }
```

But in this PR its changing again after we just defined this spec in the v9 release that just came out a little over a week ago.

I think this approach by defining the component + variant as the way to style specific components is powerful for advanced used cases, but limiting in that it only styles those components.

We have numerous components that would benefit from a more generic global definition. Most of our components should inherit what's currently defined as the text.fontFamily + text.fontWeight. Then we also want the fonts.baseSize on many components or the ability to leverage one of the generated step sizes that are being generated in the Text component.

So my concern that I originally brought up about how we're changing the data structure for theming is still relevant because the way things are going is going to require every component to have props to change simple defaults for the font family, weight, and size.

What I have been recommending prior to our v9 release and again before we're looking at introducing yet another breaking change to the theme data structure is a definition based on this approach: https://theme-ui.com/theme-spec/

By leveraging this spec we can have simple global definitions (which would be sufficient for our current use cases) and support the variants for advanced use cases later.

```
fonts: {
    body: 'system-ui, sans-serif',
    title: 'system-ui, sans-serif',
    headline: 'system-ui, sans-serif',
    label: 'system-ui, sans-serif'
	baseSize: 16,
    sizeScale: 1.15
},
fontWeights: {
    body: 400,
    title: 700,
    headline: 400,
    label: 400,
    bold: 700
},
```

Using this data could be additive and non-breaking. We can still do the variant approach as a way to do additional customizations in unique scenarios, but it would alleviate the need to define every prop on every components to pass a few CSS styles.

src/components/Text/src/Text.vue
pretzelhammer: i think it only impacts texts rendered with a `p` element which inherited some margin from the default browser stylesheet. hmmm, i'll chat about this with Lauren and Bashu and see what they say

src/components/Text/src/Text.vue
landondurnan: Removing the margin on default paragraphs does have an impact, it's unexpected and I think it should and can be avoided.

src/components/Text/src/Text.vue
bashunaimiroy: `p` elements already are styled with `margin: 0` in Website's CSS Reset ([github link](https://github.com/squareup/ecom-website/blob/text-to-mtext-conversion/resources/assets/js/prime/src/css/reset.css#L15-L22)). This should be redundant for `p` elements and therefore fine. 

My only concern is if somebody's adding `margin` to their `p` tag through a class, and then we convert it into an MText and override it- but that's mostly happening via `<style scoped>` tags at the Vue component level, and therefore the specificity will override this one, so this works as a default.

Here's an example of the specificity I just discussed. I'm using a `p` tag I converted to MText in EventStatus ([github link](https://github.com/squareup/ecom-website/pull/12057/files#diff-ab9e31e774b1d7332b68b373e05b54b801918d6d570760c698b208ee83bc0a32R55)), using a mock class with the same specificity as this `Text` class we're looking at:
image

src/components/Text/src/Text.vue
laurenhu: FWIW, Website's renderer reset.css currently resets all `p` (among other text tags) to `margin: 0`. I think as long as we are not inlining the `margin: 0` style it should be fine. MHeading currently applies class-based margins from scoped styles over the Maker `margin: 0` style.

src/components/Text/src/Text.vue
laurenhu: I responded in an different thread but I think it was resolved/collapsed. +1 to Bashu's explanation - `p` tags are currently reset anyway. This just prevents engineers from defining global styles that apply margin, but it doesn't prevent margin overrides by scoped classes. I think that's fine - the whole reason we have so many global styles in website is because it was the only way to share styles between different components, a problem Maker and Maker Theming is meant to directly address.

landondurnan: All this context is helping to establish that the impact of removing margins from the the text component would have minimum to no impact for our largest consumer right now. It would still be considered a breaking change based on the code we've already shipped, I still think we could avoid that right now until we can provide a clearer direction or ability to change it.

Some additional discussion and perhaps some documentation around our philosophy about how styling should be done globally when using Maker. For example, we can be opinionated about a reset vs normalize solution, but we should document that and communicate how to manage that output more clearly.

Tailwind takes the reset approach and then communicates where and how it's happening: https://tailwindcss.com/docs/preflight

Bootstrap starts with Normalize and then makes more opinionated targets styles: https://getbootstrap.com/docs/5.1/content/reboot/

Either way, they're clearly documenting their approach and allows consumers to subscribe or alter that default behavior. Personally -- if I was a consumer of Maker I'd want to be able to specifically target and style the margins of certain tags. One way we could do that with our Theme is having a global default with the ability to customize with a `styles` data option like so: https://theme-ui.com/theme-spec/#styles

We don't have to do any of this right now -- but I thought it worthwhile to queue up some of these approaches so we can dig in when it's appropriate. Reducing the shared styles would be ideal with our Website consumer -- but I just want to be cautious about scoped styles being the only way to get the desired output as that gets incredibly repetitive and error prone to do something as simple as adding a margin on paragraphs.

landondurnan (approved):
Approving for beta so we can start testing. I still have concerns about the theme data structure changing, but this is likely inevitable even with some alternative approaches. Let's move forward and we can evaluate the details through more testing.

PR #279, opened 2022-04-01
feat(theme)!: global theme data with css properties

landondurnan:

CleanShot 2022-03-31 at 09 26 00

* Each component requires an object merge between the default theme and any locally set props. This ends up evaluating all theming with a great deal of JS before rendering which impacts performance and increases the weight of the DOM. This also can break CSS specificity since even theme options that are considered global or internal to a section has to be rendered inline on that component.
* Several components that should inherit the default body font and weight remain unstyled since the maker theme only sets explicit styles on components. The Heading and Text component get the intended styles, but all other components don't get font updates.
* There's code duplication for the fluid type scale that should be more DRY.

* Adds a small dependency that generates CSS properties on the `:root` from our injected theme JSON.
* Leverages those same CSS properties as the default output for Theme and Heading (this could be expanded in more places -- this work is more a representative proof-of-concept). But simply applying default font styles on the Theme component allows every component inside of the theme to inherit and be reactive to those styles with just CSS.
* Creates a global and typography css that allows moving the shared styles to one place (needs more work to scope these styles), but attempts to address this issue: https://github.com/square/maker/issues/39. Updated the styleguide to leverage the global styles, so the heading / text component should reflect the fluid scale.
* Implements support for the four text variants we've discussed adding to the Text component: body (aka text), heading (aka title), headline, label. The heading and headline share styles with the ability to customize further for headline. The body and label share styles with the ability to override for label. [Updated theme lab demo to show this in practice.](https://square.github.io/maker/lab/theme-stitches-dev/#/Themes/)

CleanShot 2022-04-04 at 08 59 04
CleanShot 2022-04-04 at 08 59 26

## Other information

Updated this advanced theme demo to show impact of changes: https://square.github.io/maker/lab/theme-stitches-dev/#/Themes/

Primarily this is a proof-of concept, but there are some solutions here for an easier way for us to globally style Maker components. This currently is limited to just Typography, but could be further expanded for colors and profiles including generating `on` colors that reduce our contrast checks. In the future a similar solution for shape could be leveraged across components.

Another thing that would need to be addressed are default CSS output for each component so that if Theme data isn't present it would fall back to what we consider to be the default in that component. It does appear that right now since the font scale styles were moved that more work is needed to either include those style in the styleguide or import them directly into the component, not sure which is best.

### Proposed theme data changes
[Started to discuss this in a comment in this PR](https://github.com/square/maker/pull/278#issuecomment-1086105434)

Our current data structure looks like this for typography:
```
fonts: {
    baseSize: 16,
    sizeScale: 1.17,
},
text: {
    fontFamily: 'inherit',
    size: 0,
    color: '@colors.text',
    fontWeight: 400,
},
heading: {
    fontFamily: 'inherit',
    size: 2,
    color: '@colors.heading',
    fontWeight: 700,
},
```

While this supports the ability to default and have custom overrides for these components, there's a massive gap with how our default typography styles apply to all components (see image at top). We need a set of font definitions that work all text output. By altering our default data to be more generic we can document and leverage those styles logically based on our primary support for typography, even with adding additional typographic options beyond heading and text components.

I've implemented that in the theme lab demo, but this is what it looks like when passing theme data to the theme component:
```
fonts: {
	body: 'Domine',
	heading: 'Karla',
	baseSize: 19,
	sizeScale: 1.217,
},
fontWeights: {
	body: '400',
	heading: '600',
},
```

So out of the box the body font family, weight, and base size is applied globally on the theme.  And then at the component level the text component was updated to support additional variants, namely `body`, `title`, `headline`, and `label`. (Sidenote -- I used the names we've defined in website, but to keep consistent with the data perhaps `title` should be `heading`)  The component has different fallbacks based on the logic we prefer where body is the default, but can be customized further like so:

```
"fonts": {
    "body": "Open Sans",
    "heading": "Open Sans",
    "label": "Playfair Display",
    "baseSize": 16,
    "sizeScale": 1.15
},
"fontWeights": {
    "body": "400",
    "heading": "600",
    "label": "800"
}
```

In this case we can set a label option and the label variant in the text component will use those new styles instead. This could easily be expanded to every other component that we want to support a label output -- things like nav, forms, buttons. But we don't need to add variant options to those components. We can simply style the defaults so the CSS fallbacks work based on label > body > system-font.


landondurnan:
Sample output of the css properties being generated from the theme data:
```
:root, .maker-t-jFMsbb {
    --maker-colors-background: #31353f;
    --maker-colors-heading: #ffffff;
    --maker-colors-text: #ffffff;
    --maker-colors-primary: #d5e2fb;
    --maker-colors-neutral-0: #000000;
    --maker-colors-neutral-10: #60636b;
    --maker-colors-neutral-20: #777981;
    --maker-colors-neutral-80: #9c9ea3;
    --maker-colors-neutral-90: #f4f4f4;
    --maker-colors-neutral-100: #ffffff;
    --maker-colors-color-elevation: #777981;
    --maker-colors-color-overlay: rgba(255, 255, 255, 0.32);
    --maker-heading-fontFamily: Domine;
    --maker-heading-fontWeight: 400;
    --maker-text-fontFamily: Karla;
    --maker-text-fontWeight: 400;
    --maker-fonts-baseSize: 19;
    --maker-fonts-sizeScale: 1.217;
  ```

File: src/components/Theme/src/Theme.vue [hunk 1]
```diff
@@ -54,35 +44,47 @@ export default {
...
 		resolveTheme(this.$data, this.parentTheme, this.theme, this.profile);
+		this.applyTheme();
+	},
+	beforeMount() {
+		this.applyTheme();
+	},
+	methods: {
+		applyTheme() {
+			const { theme } = createStitches({ theme: this.theme, prefix: 'maker' });
+			this.defaultTheme = theme;
```
pretzelhammer, line 57: i don't think this line is doing anything?

src/components/Theme/src/Theme.vue
```diff
@@ -54,35 +44,47 @@ export default {
...
-		// update theme on prop changes
 		resolveTheme(this.$data, this.parentTheme, this.theme, this.profile);
+		this.applyTheme();
+	},
+	beforeMount() {
+		this.applyTheme();
+	},
+	methods: {
+		applyTheme() {
+			const { theme } = createStitches({ theme: this.theme, prefix: 'maker' });
```
pretzelhammer, line 56: I'm assuming this takes the JS theme object and creates a bunch of CSS custom properties from it and adding them to the `:root` scope of the page. is that all it's doing or is it also doing anything else? I read [the docs](https://stitches.dev/docs/api#createstitches) for this function but they're not really clear about what it's actually doing.

src/components/Text/src/Text.vue
```diff
@@ -110,34 +110,40 @@ export default {
...
 			validator: (textAlign) => ['inherit', 'left', 'right', 'center'].includes(textAlign),
 		},
+		/**
+		 * Variant - allows four custom font styles through the Theme component
+		 * @values body, title, headline, label
+		 */
+		variant: {
+			type: String,
+			default: 'body',
+			validator: (variant) => ['body', 'title', 'headline', 'label'].includes(variant),
```
pretzelhammer, line 120: this seems to hardcode 4 variants into the component, and from looking at the CSS classes below the only customizable properties of each variant are color, fontFamily, and fontWeight.

can you generalize this implementation to allow for arbitrarily many user-defined variants which allow customizing any of the themeable properties of the component? this approach has to be able to do at least that much to reach feature-parity with the current beta release.

landondurnan: We only have plans to support four variants and the only thing that can be changed are those three options.

src/components/Theme/src/Theme.vue
landondurnan: https://stitches.dev/docs/api#theme-1 this provides more details around this function, the only difference is the `core` library does the same thing being agnostic to the framework.

src/components/Theme/src/Theme.vue
landondurnan: Yeah, you're right. I pulled this over from an older implementation. Originally, I was passing this as the the provide data to all the components, but in this iteration it's not necessary. I'll remove it in this PR.

This would set us up to completely remove the default-theme.js. The default styles would be set either globally in a css file or respectively in each component. The theme injection would then provide the ability for external consumers to pass that data.

src/components/Theme/src/Theme.vue
pretzelhammer: that link isn't that helpful, it doesn't say anything about appending styles to the `:root` or why it does that, i wanna understand the rationale behind throwing stuff into `:root`

src/components/Text/src/Text.vue
pretzelhammer: we have user-defined variants in beta

pretzelhammer (changes requested):
To summarize some convos we've had on Slack:
1. Please update this PR so that generated CSS variables do not pollute the global scope, but are scoped only to where MTheme is mounted & rendered within the DOM (as it currently works in `master`).
2. If you're only using stitches to do is convert a JS object into CSS variables then please remove it as a dependency and replace it with a function that does the same thing. We do not need stitches for this functionality.
3. Please rebase this PR from the `master` branch to the `beta` branch so that we can see your changes in context with the latest updates to the theming system and the Text component.
4. Please show how your approach handles nesting Theme components. A nested Theme component should be able to do both: 1) Reference a profile (a full theme definition) that's defined in the parent Theme's `profiles` prop and 2) Define a new theme that inherits any properties it doesn't specify from the parent theme. The solution should be holistic and not piecemeal, i.e. nested themes need to support 100% of the functionality of unnested themes (to match functionality parity with what's currently in `master`). A working Website branch would be ideal.
5. Please show how your approach handles user-defined variants in the Theme (to match functionality parity with what's currently in `beta`). A working Website branch would be ideal.
6. Please show how your approach handles validating "mixed" props, where mixed means: a combination of Vue props and CSS custom props (aka CSS vars). For example, currently the Button components receives `color` and `bgColor` as Vue props (regardless of whether they are set as local overrides or passed via a Theme component) and can use chroma to contrast check the values. In the approach this PR proposes, the Button component would be refactored to inherit `color` and `bgColor` as CSS custom props but the user would still be able to set either as a local override on the Button component and then how would that component perform a contrast check? Also, even if both are inherited as CSS custom props, does the Button component still try to do a contrast check then or would it just take the values as-is?

src/components/Theme/src/Theme.vue
landondurnan: that's the best way to define global css variables: https://developer.mozilla.org/en-US/docs/Web/CSS/:root

laurenhu:
I agree that this PR should be rebased against `beta`. There's currently a need for variant support of the type that it provides which we haven't figured out an adequate solution for purely with predefined variants and css.

We will also need `resolveThemeableProps` restored for MText/MHeading `color` at the very least but in general I'm wary of excluding any props before we are even completely done with typography. We can add css variables that are usable by non-Maker components and set styles that cascade down from the Site Theme without this change.

landondurnan:
Just going to call out that I'm not going to rebase off of beta, because this is meant to be a completely different alternative approach to what is currently in the beta. I will instead document the primary differences between what exists now in production and each approach as two different alternatives.

src/components/Theme/src/Theme.vue
pretzelhammer: that link also isn't helpful, it doesn't explain why defining css vars on `:root` is "the best" or better than defining them anywhere else, also as i explained and requested earlier i don't think we should be polluting the global css scope with maker's css vars

landondurnan:
Closing this in favor a more incremental approach -- will keep linking other PRs here for reference, but this PR won't ever get merged as-is.

PR #280, opened 2022-04-01
ci: run semantic release on maintanence releases

pretzelhammer:

once we make a major release there's no automated way for backporting features or bugfixes to prior major releases

adds `'[1-9][0-9]*.x'` to the `branches` config of our `release` action which should now run semantic-release on any pushes to branches that match that regex pattern, e.g. `8.x`, `9.x`, `10.x` etc

## Other information

- github docs on branch filtering syntax (which is not regex but is close to it): https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
- semantic release docs on publishing maintenance releases: https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/release-workflow/maintenance-releases.md

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
-    branches: [master, next, next-major, beta, alpha]
+    branches: [master, next, next-major, beta, alpha, '[1-9][0-9]*.x']
 
 jobs:
   release:


privatenumber (approved):

landondurnan:
This is really cool @pretzelhammer, I didn't think it would be this easy. Thanks for working on this

PR #281, opened 2022-04-04
fix: double scrollbars on Dialog, Modal, Blade (#273)

pretzelhammer:

Experimenting with backporting bug fixes to older major versions. This code has already been merged and released for v9 but I'm backporting it to v8.

fixes double scrollbars for v8

## Other information

nope

### DIFF
--- a/src/components/Blade/src/Blade.vue
+++ b/src/components/Blade/src/Blade.vue
@@ -17,7 +17,7 @@
 
 .Blade {
 	height: 100%;
-	overflow: scroll;
+	overflow: auto;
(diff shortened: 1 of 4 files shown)


pretzelhammer:
okay... this did not work like how i expected it to work... the release job never ran... i think it's because i wrote the branch filtering syntax as `'[1-9][0-9]*.x'` which i EXPECTED would match `8.x` but it doesn't because github doesn't treat `*` the same as regex does. i need to update this to `'[0-9]+.x'`.

PR #282, opened 2022-04-04
ci: fix github branch filtering syntax for release job

pretzelhammer:

I tried backporting a bugfix from v9 to v8 but the release action never ran: https://github.com/square/maker/pull/281#issuecomment-1087768768

i think it's because i didn't read github docs closely enough. I thought `*` meant _"match zero or more of the **preceeding** character"_ which is what it means in regex... but how github interprets it as is _"match zero or more of **any** character"_ 🤦 .

fixes the github branch filtering syntax so that it hopefully works this time

## Other information

nope

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


landondurnan (approved):
🤞 

what bug are you backporting?

PR #288, opened 2022-04-06
feat(menu): menu component

laurenhu:
Menu component including
`MMenu.vue` - Initializes and toggles a Popover with a toggle that can be either a select-style dropdown or a button
`MMenuOption.vue` - Wraps a menu option or action

Additional utility components
`utils/Row/src/Row.vue` - A Row component

Screen Shot 2022-04-06 at 6 44 17 PM
Screen Shot 2022-04-06 at 6 44 28 PM
Screen Shot 2022-04-06 at 6 44 35 PM


landondurnan (commented):
Overall the code is really clean and the row utility and interactions are all nice. 

I had some ideas about the template markup that will require a bit more time to thing through and write-up. These are likely things you already considered: 

One suggestion is to replace the toggle prop with a slot that a user could insert their own component. This essentially would be identical to the existing popover component. We could simply suggest a button / select component where appropriate. This would allow reuse of our own components and keep the styles associated with that output within those components. I'm not sure what issues there could be with a select control -- but we'd likely need to prevent the default behavior?

src/utils/Row/src/Row.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+	<div
+		:class="$s.Row"
+		v-bind="$attrs"
+		v-on="$listeners"
+	>
+		<div
```
landondurnan, line 7: Since we have a "prefix" and "suffix" on inputs, it might make sense to make the icons slots a bit more generic and named the same way.  That could open up the ability also to pass other atomic controls in like checkbox, radio, toggle, etc.

landondurnan: Icon = prefix

src/utils/Row/src/Row.vue
```diff
@@ -0,0 +1,98 @@
...
+			</div>
+			<div
+				v-if="$slots['side-secondary']"
+				:class="$s.Secondary"
+			>
+				<slot name="side-secondary" />
+			</div>
+		</div>
+		<div
+			v-if="$slots['side-icon']"
```
landondurnan, line 42: side-icon = suffix

src/components/Select/src/SelectControl.vue
```diff
@@ -12,7 +12,23 @@
 			<!-- @slot Select prefix -->
 			<slot name="prefix" />
 		</span>
+		<button
+			v-if="$slots.default"
+			:class="[
+				$s.Select,
```
landondurnan, line 18: Just to be a bit safer, could we create a separate class for this like `.SelectButton` you can then share the CSS with the `.Select` and more safely target the new styling just to the button.

landondurnan (commented):
Had one suggestion about the SelectControl button style. Besides that everything is looking good.

While I don't think we need to do this now, we should consider accessibility as the menu currently doesn't have aria / keyboard controls. One way we could try to resolve that is reworking the select component to support custom menu output. 

The primary reason so that we could support both a native / nonNative (i.e. highly styled) output and improve our accessibility with aria labels and keyboard controls to navigate through the menu options.

There's a good article about doing something like that here: https://css-tricks.com/striking-a-balance-between-native-and-custom-select-elements/

That would likely be a larger refactor to the Select component, but thankfully we have a good chunk of this work in this feature that we could continue to build on top. I think the only consideration long-term is how might that impact the Menu component if you could effectively make a Menu dropdown with a Select component? When would you use one over the other if we went did those refactors? Nothing we need to answer now as I don't think any of it has a big impact on our APIs, but if anyone has some thoughts it might benefit to have a bit of that discussion in here for posterity.

landondurnan:
On second-thought: there are some things we could attempt to do now based on WAI recommendations for a component like this: https://www.w3.org/WAI/tutorials/menus/application-menus/

landondurnan (approved):
Awesome work! Excited to try the row component in some more places and the accessibility that you added to the menu works great.

PR #293, opened 2022-04-12
fix!: renames global 'variants' to 'patterns' to avoid name collision with local 'variant' props

pretzelhammer:

allows customizing existing "patterns" and creating new "patterns" for buttons, which was not possible previously while they were still called "variants" and there was a name collision with button's local "variant" prop

just did

## Migration Guide

- change all `<m-text variant="etc">` to `<m-text pattern="etc">`
- change all `<m-button variant="etc">` to `<m-button pattern="etc">`
- change any theme data from
```js
{
  button: {
    variants: {
       // etc
    },
  },
  text: {
    variants: {
      // etc
    },
  },
},
```
- to this
```js
{
  button: {
    patterns: {
       // etc
    },
  },
  text: {
    patterns: {
      // etc
    },
  },
},
```

### DIFF
--- a/src/components/Button/README.md
+++ b/src/components/Button/README.md
@@ -2,38 +2,6 @@
 
 For a simplified text-only button with no padding or focus/active/hover states, see: [TextButton](#/TextButton).
 
-```vue
-<template>
-	<div>
-		<m-button
-			variant="primary"
-		>
-			primary
-		</m-button>
-		<m-button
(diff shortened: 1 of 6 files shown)


laurenhu (approved):
I really like the nomenclature of `pattern` to describe groups of props that are reused and how it separates those groupings from a component's variant since a pattern can encompass a variant as well.

PR #294, opened 2022-04-12
feat(theme): Global font theme data

landondurnan:

image

The theme data only allowed passing font family / weight to the Heading and Text component.

This PR allows for passing global font data for the family and weight to theme. The theme applies the body family / weight as a default as a style for overall theme. Any components in the theme will inherit that as a default.

The data is compliant with the [Theme UI Spec](https://theme-ui.com/theme-spec):
```
fonts: {
	baseSize: 16, // existing data
	sizeScale: 1.17, // existing data
	body: 'Open Sans',
	heading: 'Open Sans',
},
fontWeights: {
	body: 400,
	heading: 600,
},
```

This is also backwards compatible with the existing component data we're passing for the heading and text component. If users want to keep it the same they could reference the font data the same way they reference global colors like so:

```
text: {
	color: '@colors.text',
	fontWeight: '@fontWeights.body',
	fontFamily: '@fonts.body',
},
heading: {
	color: '@colors.heading',
	fontWeight: '@fontWeights.heading',
	fontFamily: '@fonts.heading',
},
```

## Other information

This is PR is part of a larger effort to support all the typography options that are needed. Originally, the bulk of this work was done here (https://github.com/square/maker/pull/279) but this PR was modified to address feedback and implement changes in a non-breaking way.

You can test the changes in a branched version of the Theme Lab here: https://square.github.io/maker/lab/font-theme-data/#/Themes/

### DIFF
--- a/lab/experiments/Themes/index.vue
+++ b/lab/experiments/Themes/index.vue
@@ -276,10 +276,10 @@ export default {
 		background: (store) => store.theme.colors.background,
 		fontLoad() {
 			const fonts = [];
-			const fontHeading = themeStore.$state.theme.heading.fontFamily;
-			const fontWeightsHeading = themeStore.$state.theme.heading.fontWeight;
-			const fontText = themeStore.$state.theme.text.fontFamily;
-			const fontWeightsText = themeStore.$state.theme.text.fontWeight;
+			const fontHeading = themeStore.$state.theme.fonts.heading;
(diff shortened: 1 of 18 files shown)


landondurnan:
Since we're planning on breaking changes anyway to support upcoming requirements, going to close this out and rework on top of the existing beta.

PR #296, opened 2022-04-12
feat: Additional Theme font options

landondurnan:

https://github.com/square/maker/issues/269

We currently have the ability to pass font data related to Heading and Text, but there's a desire to support more font options to allow for a greater range of customization of typography output.

CleanShot 2022-04-12 at 15 46 29

- Theme: Adds data and properties for two new font options: `headline` and `label`
- Text: 
	- Updates component to support leveraging any of the four supported font options through a variant prop: `headline`, `label`, `heading`, and `body`. Body is the default font style.
	- Expands the supported `element` options so that the Text component could be used in place of where the Heading component is used. (We're planning on consolidating the Heading and Text component into Text and remove Heading in a future release)
- Buttons: Updates default to leverage the label font styles with a fallback to the body if not set.
- Theme Lab: Adds controls and preview output for Headline and Label font styles.

## Other information

This is PR is part of a larger effort to support all the typography options that are needed. Originally, the bulk of this work was done here (https://github.com/square/maker/pull/279) but this PR was modified to address feedback and implement changes in a non-breaking way.

This is based on existing work in this PR: https://github.com/square/maker/pull/294

You can test the feature additions in this theme lab: https://square.github.io/maker/lab/new-font-data/#/Themes/

### DIFF
--- a/src/components/Text/src/Text.vue
+++ b/src/components/Text/src/Text.vue
@@ -39,12 +39,12 @@ export default {
 		},
 		/**
 		 * Variant, allows four custom font styles through the Theme component
-		 * @values body, title, headline, label
+		 * @values body, heading, headline, label
 		 */
 		variant: {
 			type: String,
 			default: 'body',
-			validator: (variant) => ['body', 'title', 'headline', 'label'].includes(variant),
+			validator: (variant) => ['body', 'heading', 'headline', 'label'].includes(variant),
 		},
 		/**
 		 * Font family
@@ -122,7 +122,7 @@ export default {
 	},
 
 	computed: {
-		...resolveThemeableProps('text', ['size', 'fontFamily', 'fontWeight', 'color']),
+		...resolveThemeableProps('text', ['size', 'fontFamily', 'fontWeight']),
 		sizeClass() {
 			const minNonNegativeSize = 0;
 			if (this.resolvedSize >= minNonNegativeSize) {
@@ -136,7 +136,7 @@ export default {
 			return {
 				fontFamily: this.useThemeComponentData ? this.resolvedFontFamily : this.fontFamily,
 				fontWeight: this.useThemeComponentData ? this.resolvedFontWeight : this.fontWeight,
-				color: this.resolvedColor,
+				color: this.color,
 				fontSize: this.fontSize,
 				lineHeight: this.lineHeight,
 				'--mobile-base-font-size': fonts.baseSize,
@@ -280,14 +280,18 @@ export default {
 	font-family: var(--fonts-body, inherit);
 }
 
-.text_title {
+.text_heading,
+.text_headline {
 	margin: 0;
+	color: var(--color-heading, #000);
+}
+
+.text_heading {
 	font-weight: var(--fontWeights-heading, 600);
 	font-family: var(--fonts-heading, inherit);
 }
 
 .text_headline {
-	margin: 0;
 	font-weight: var(--fontWeights-headline, var(--fontWeights-heading, 600));
 	font-family: var(--fonts-headline, var(--fonts-heading, inherit));
 }
--- a/src/components/Text/src/Text.vue
+++ b/src/components/Text/src/Text.vue
@@ -26,7 +26,7 @@ export default {
 		element: {
 			type: String,
 			default: 'p',
-			validator: (element) => ['p', 'span', 'div', 'li'].includes(element),
+			validator: (element) => ['p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'span', 'div', 'li'].includes(element),
 		},
 		/**
 		 * Size of text
@@ -36,6 +36,15 @@ export default {
 			type: Number,
 			default: undefined,
 			validator: (size) => size >= MIN_SIZE && size <= MAX_SIZE,
+		},
+		/**
+		 * Variant, allows four custom font styles through the Theme component
+		 * @values body, title, headline, label
+		 */
+		variant: {
+			type: String,
+			default: 'body',
+			validator: (variant) => ['body', 'title', 'headline', 'label'].includes(variant),
 		},
 		/**
 		 * Font family
@@ -123,9 +132,11 @@ export default {
 		},
 		inlineStyles() {
 			const { fonts } = this.theme;
+			const resolveProps = this.resolveVariant();
+
 			return {
-				fontFamily: this.resolvedFontFamily,
-				fontWeight: this.resolvedFontWeight,
+				fontFamily: resolveProps ? this.resolvedFontFamily : this.fontFamily,
+				fontWeight: resolveProps ? this.resolvedFontWeight : this.fontWeight,
 				color: this.resolvedColor,
 				fontSize: this.fontSize,
 				lineHeight: this.lineHeight,
@@ -135,9 +146,16 @@ export default {
 		},
 	},
 
+	methods: {
+		resolveVariant() {
+			return !(this.variant === 'headline' || this.variant === 'label');
+		},
+	},
+
 	render(createElement) {
 		const {
 			$s,
+			variant,
 			sizeClass,
 			inlineStyles,
 			fontStyle,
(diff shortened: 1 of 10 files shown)


src/components/Text/src/Text.vue
```diff
@@ -37,6 +37,15 @@ export default {
...
 			validator: (size) => size >= MIN_SIZE && size <= MAX_SIZE,
 		},
+		/**
+		 * Variant, allows four custom font styles through the Theme component
+		 * @values body, title, headline, label
+		 */
+		variant: {
+			type: String,
+			default: 'body',
+			validator: (variant) => ['body', 'title', 'headline', 'label'].includes(variant),
```
pretzelhammer, line 47: this should be `paragraph` instead of `body`

src/components/Text/src/Text.vue
```diff
@@ -249,6 +268,30 @@ export default {
 	--lh-7: calc(var(--lh-6) * var(--line-height-scale));
 }
 
+/* Variant defaults */
+.text_body,
+.text_label {
+	font-weight: var(--fontWeights-body, 400);
+	font-family: var(--fonts-body, sans-serif);
```
pretzelhammer, line 275: why is the fallback `sans-serif` instead of `inherit`? please make it the latter

src/components/Text/src/Text.vue
```diff
@@ -135,9 +146,16 @@ export default {
 		},
 	},
 
+	methods: {
+		resolveVariant() {
```
pretzelhammer, line 150: this should be a computed prop

src/components/Text/src/Text.vue
```diff
@@ -123,9 +132,11 @@ export default {
 		},
 		inlineStyles() {
 			const { fonts } = this.theme;
+			const resolveProps = this.resolveVariant();
+
 			return {
-				fontFamily: this.resolvedFontFamily,
-				fontWeight: this.resolvedFontWeight,
+				fontFamily: resolveProps ? this.resolvedFontFamily : this.fontFamily,
+				fontWeight: resolveProps ? this.resolvedFontWeight : this.fontWeight,
```
pretzelhammer, line 139: This code is very confusing. If `this.fontFamily` is ever anything other than `undefined` then that's what `this.resolvedFontFamily` will be set to, same for the font weights. The only time they are different are if `this.fontFamily` is `undefined` in which case `this.resolvedFontFamily` will get the value `inherit` from the default theme so why not just check that? Or why not just check for the value `inherit` instead of using this convoluted logic? Also why is the logic depend on whether or not it's a `paragraph`/`title` variant vs a `headline`/`label` variant?

pretzelhammer:
this pr should be closed because it's subsumed by this pr: https://github.com/square/maker/pull/289

the latter pr updates more components (for both typography & color) and works on top of the latest code in beta (which includes breaking changes like getting rid of the MHeading component and merging it into MText), and also has more features (like user-defined theme patterns (formerly known as user-defined theme variants))

src/components/Theme/src/default-theme.js
```diff
@@ -19,6 +19,16 @@ export default function defaultTheme() {
 		fonts: {
 			baseSize: 16,
 			sizeScale: 1.17,
+			body: 'inherit',
+			heading: 'inherit',
+			headline: 'inherit',
+			label: 'inherit',
+		},
+		fontWeights: {
+			body: 400,
```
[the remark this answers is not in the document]
landondurnan: Answered in previous comment.

src/components/Text/src/Text.vue
pretzelhammer: `paragraph` is what Website uses and we should stay aligned with Website where possible given they're our only and biggest user

src/components/Text/src/Text.vue
```diff
@@ -26,7 +26,7 @@ export default {
 		element: {
 			type: String,
 			default: 'p',
-			validator: (element) => ['p', 'span', 'div', 'li'].includes(element),
+			validator: (element) => ['p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'span', 'div', 'li'].includes(element),
```
pretzelhammer, line 29: adding this makes the mheading component irrelevant/useless. it should be removed.

src/components/Text/src/Text.vue
```diff
@@ -123,21 +132,35 @@ export default {
 		},
 		inlineStyles() {
 			const { fonts } = this.theme;
+
 			return {
-				fontFamily: this.resolvedFontFamily,
-				fontWeight: this.resolvedFontWeight,
+				fontFamily: this.useThemeComponentData ? this.resolvedFontFamily : this.fontFamily,
+				fontWeight: this.useThemeComponentData ? this.resolvedFontWeight : this.fontWeight,
 				color: this.resolvedColor,
```
laurenhu, line 139: This no longer applies the correct resolved color. We would need to replace it with a computed property to handle mapping from `theme.colors.heading` and `theme.colors.text` to the different relevant variants.

OR the preferable approach would be to just change this to `color: this.color` and adding `--color-heading` and `--color-text` to the correct places in the CSS classes below.

I do think if we do this, we might as well remove `text` as a theme property - it wouldn't do anything with a useful level of specificity that Text.vue would't be getting from `theme.colors` and `theme.fonts` instead.

This is a pretty important distinction between this and the alternate approach which builds on our current theming system.

src/components/Text/src/Text.vue
```diff
@@ -123,21 +132,35 @@ export default {
...
 				fontSize: this.fontSize,
 				lineHeight: this.lineHeight,
 				'--mobile-base-font-size': fonts.baseSize,
 				'--mobile-font-size-scale': fonts.sizeScale,
 			};
 		},
+		/*
```
laurenhu, line 146: I don't think this does anything. Once we remove `theme.text` none of the theme component data will be applied here. They'll all be applied on MTheme as css and referenced in Text.vue

src/components/Text/src/Text.vue
landondurnan: It does, I agree. I figured the follow-up work from this PR would be the breaking changes we're planning.

src/components/Text/src/Text.vue
landondurnan: I'd really prefer to keep somewhat agnostic to UI in one product. It used to be body, now it's paragraph, next week it could be text -- the point is Website should feel free to change their label and output however they see fit. We should come up with something that doesn't need to change. The primary goal of `body` is this is the default styles applied to the entire body of the document -- this font and weight is used for more than just paragraphs so it's a more inclusive and descriptive name of how it will be applied in this and every other component.

src/components/Text/src/Text.vue
landondurnan: @laurenhu You're right, I was working on some code that did that, but I was hesitant to change too much when changing how color applies didn't seem like the primary goal. But I think as-is it may be a regression or not work as expected. I'll do some more testing in the lab perhaps try both.

src/components/Text/src/Text.vue
landondurnan: I didn't want to imply that we should remove `theme.text`. I think we should continue to support the ability to pass data to components the same way we do now and expand on that with variants.

This is more meant to be a short-term approach to provide that data that shouldn't change the public API. I think both ways of passing data would work, but I'd like to spend more time to evaluate how reusable css is generated and how the variants would work together.

src/components/Text/src/Text.vue
landondurnan: I updated to leverage the color data in the style so the output maps as expected. It's also still possible to pass a color directly inline:
CleanShot 2022-04-14 at 15 08 45

src/components/Text/src/Text.vue
landondurnan: I realized that the variant `title` was following an old name and didn't conform to our existing internal data, the theme spec, or the new name. So updated `title` to `heading`

landondurnan:
Since we're planning on breaking changes anyway to support upcoming requirements, going to close this out and rework on top of the existing beta.

PR #297, opened 2022-04-13
fix(choice): fix reactivity of choice => choice option

tatems:
When introducing the `disabled` field on the choice provider, reactivity was broken and prevented UI updates from selections/prop changes. This PR adds a small refactor to move selection/disabled handling to the provider.

### DIFF
--- a/src/components/Choice/src/Choice.vue
+++ b/src/components/Choice/src/Choice.vue
@@ -10,6 +10,7 @@
 </template>
 
 <script>
+
 import { MThemeKey, defaultTheme, resolveThemeableProps } from '@square/maker/components/Theme';
 import assert from '@square/maker/utils/assert';
 import getContrast from '@square/maker/utils/get-contrast';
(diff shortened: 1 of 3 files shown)


src/components/Choice/src/Choice.vue
```diff
@@ -18,10 +19,35 @@ import key from './key';
...
 	provide() {
+		/**
+		 * Unfortunately, it's difficult to pass reactive data via a provider in Vue 2
+		 * Vue 3 will bring the `computed` helper to make (most of) this cleaner
+		 */
 		return {
 			[key]: {
-				...this.$data,
-				disabled: this.disabled,
+				isSelected: (value) => {
```
landondurnan, line 28: I think we could likely simplify all this by referencing another similar component we've written in the past: https://github.com/square/orbit/blob/master/src/components/ButtonGroup/src/ButtonGroup.vue

All the data should be passed down from the parent and the child elements can leverage that data or call a method to update it.

tatems: Fair enough!

landondurnan (approved):
Code looks good and looks to all be working as expected now. Thanks for the quick fix!

PR #298, opened 2022-04-13
fix(popover): update popover color styles

laurenhu:
Changes to Popover color styles as well as lab to demo them.

Screen Shot 2022-04-13 at 4 35 01 PM
Screen Shot 2022-04-13 at 4 35 16 PM

### DIFF
--- a/lab/experiments/Popover.vue
+++ b/lab/experiments/Popover.vue
@@ -1,0 +1,196 @@
+<template>
+	<div
+		:class="$s.Wrapper"
+	>
+		<m-theme
+			:class="$s.Preview"
+			:theme="theme(backgroundColor)"
+		>
+			<m-heading
+				:size="0"
+			>
+				Parent Site MTheme
+			</m-heading>
+			<input
+				v-model="backgroundColor"
+				type="color"
+			>
+			<m-popover>
+				<template #action="popover">
+					<m-button
+						:class="$s.Button"
+						@click="popover.toggle()"
+					>
+						Popover Toggle
+					</m-button>
+				</template>
+				<template #content>
+					<m-popover-content
+						:class="$s.Popover"
+					>
+						<m-heading>
+							Popover content
+						</m-heading>
+						<m-text>
+							Content for a basic popover
+						</m-text>
+						<m-divider />
+						<m-text>
+							Content for a basic popover
+						</m-text>
+						<m-input
+							variant="outline"
+							placeholder="Delivery address"
+						/>
+					</m-popover-content>
+				</template>
+			</m-popover>
+			<m-text>
+				Lorem ipsum dolor sit amet, consectetur adipiscing elit,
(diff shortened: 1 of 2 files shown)


lab/experiments/Popover.vue
```diff
@@ -0,0 +1,196 @@
...
+				sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+				Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris
+				nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in
+				reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+				pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa
+				qui officia deserunt mollit anim id est laborum.
+			</m-text>
+			<m-theme
+				:class="$s.Block"
+				:theme="theme(blockColor)"
```
landondurnan, line 59: I tested this with a full theme data sample and it also works by just setting the profile, but either work.

lab/experiments/Popover.vue
```diff
@@ -0,0 +1,196 @@
...
+				<m-heading
+					:size="0"
+				>
+					Nested Block MTheme
+				</m-heading>
+				<input
+					v-model="blockColor"
+					type="color"
+				>
+				<m-popover>
```
landondurnan, line 70: One thought I had was we could make the popover more flexible by passing a profile prop to the component, this is something I thought we could do with sections too -- but for now it makes sense for it to inherit whatever the defaults from it's parent.

src/components/Popover/src/PopoverContent.vue
```diff
@@ -33,9 +41,19 @@ export default {
 
 	computed: {
 		computedStyles() {
+			// PopoverContent can be rendered outside the current Mtheme wrapper
+			// depending on placement of MPopoverLayer so we need to make sure
+			// it receives the correct Mtheme color variables
+			const themeColorVars = {};
+			Object.entries(this.theme.colors).forEach(([color, hex]) => {
```
landondurnan, line 48: It's unfortunate that we have to reprocess the values again here into CSS, but this is the best solution for now.

Ideally we can simplify this later by generating the CSS at the top of the theme for the default colors and profiles -- at that point we could remove this and optimize with setting a class on the popover.

lab/experiments/Popover.vue
landondurnan: CleanShot 2022-04-13 at 15 40 57

screenshot from that test. No issues, just wanted to show it works as expected.

landondurnan (approved):
I think this looks pretty good and the output is what we were hoping to get. We can likely optimize some of the internals later or add the ability to pass profile names, but for now this is a great improvement.

pretzelhammer (approved):

PR #300, opened 2022-04-18
fix(pininput): on safari one-time-code input there is no event.data

Ryguy11:
In safari the one-time-code autofill option was not correctly filling out all the inputs in the pin-input. Instead the autocomplete code was being filled completely in the first input.
square-pay-code

The issue was the event.data of the autofill did not contain any data on Safari. Not exactly sure why to be honest, but we are able to work around this by reaching into `event.target.value` for the input.

### DIFF
--- a/src/components/PinInput/src/PinInput.vue
+++ b/src/components/PinInput/src/PinInput.vue
@@ -126,7 +126,7 @@ export default {
 		},
 
 		onInputPin(event, index) {
-			const input = event?.data || event?.target?.value;
+			const input = event?.target?.value;
 			if (!input) {
 				return;
 			}
--- a/src/components/PinInput/src/PinInput.vue
+++ b/src/components/PinInput/src/PinInput.vue
@@ -126,13 +126,13 @@ export default {
 		},
 
 		onInputPin(event, index) {
-			if (!event.data) {
+			const input = event?.data || event?.target?.value;
+			if (!input) {
 				return;
 			}
-
 			// Only allow integers in input
 			const BASE_TEN = 10;
-			const inputValue = Number.isInteger(Number.parseInt(event.data, BASE_TEN)) ? event.data : '';
+			const inputValue = Number.isInteger(Number.parseInt(input, BASE_TEN)) ? input : '';
 
 			// One-time-code autofill is treated as an input, not a paste
 			if (this.attemptSplitPinIntoInputs(inputValue, index)) {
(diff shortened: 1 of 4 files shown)


tatems (approved):

src/components/PinInput/src/PinInput.vue
```diff
@@ -126,13 +126,13 @@ export default {
 		},
 
 		onInputPin(event, index) {
-			if (!event.data) {
+			const input = event?.data || event?.target?.value;
```
landondurnan, line 129: Could you add some comments here about the data values and the reason for the conditional? Also any external references regarding why safari works a bit differently if they exist?

pretzelhammer, line 129: Can you help me understand why `event.data` was used in the first place? I'm use to seeing `event.target.value` everywhere, which seems to be the reliable cross-browser standard for accessing an input's value, so if you're gonna be using that you can probably just get rid of the `event.data`.

pretzelhammer: Oh, does it have something to do with people trying to paste the entire code into the first input cell? 🤔

Ryguy11: https://developer.mozilla.org/en-US/docs/Web/API/InputEvent

Youre right that it makes sense to just get rid of the event.data in this case

mattmewton (approved):

src/components/PinInput/src/PinInput.vue
Ryguy11: No - its just that the input event has a `.data` attribute that seems to have cross-browser support which is why its originally what I was using. On Safari it doesnt seem to be there for some reason, even though the docs say it supports  it. I also tried using `input.native` to see if for some reason that was messing with anything but it had the same effect

Ryguy11: Im just going to remove the conditional as per Kirills suggestion

Ryguy11: re: pasting the entire code into the first input cell - the component already supports that functionality 👍

pretzelhammer (approved):

src/components/PinInput/src/PinInput.vue
pretzelhammer: great, thanks for the context! i've resolved the open threads and approved the pr

PR #304, opened 2022-04-20
refactor!: theme color mapping and output

landondurnan:

There's some changes in the v10 Beta theme configuration that make modifications to the theme data that are different from our plans and long-term goals.

This PR updates the theme configuration data so that it's more compatible with existing data and theme data we plan to support in the future.

### Details
The current theme data has a set of global design tokens primarily focused on Color, with v10 adding additional support for theming Typography. I've outlined the specific data changes in each version with more explanation on what we should add, change, or leave alone.

#### Colors

```
// SCROLL THIS CODE
// v9 Theme Color Data and CSS Properties
// Neutrals stay the same
'--color-background': colors.background,
'--color-text': colors.text,
'--color-heading': colors.heading,
'--color-elevation': colors['color-elevation'],
'--color-overlay': colors['color-overlay'],

// v10 Beta Theme Color Data and CSS Properties
// Neutrals stay the same
'--color-background': colors.background,
'--color-paragraph': colors.paragraph,
'--color-title': colors.title,
'--color-primary': colors.primary,
'--color-elevation': colors['color-elevation'],
'--color-overlay': colors['color-overlay'],

// v10 Recommended Theme Color Data CSS Properties
// Neutrals stay the same
'--color-background': colors.background,
'--color-text': colors.text,
'--color-heading': colors.heading,
'--color-primary': colors.primary,
'--color-secondary': colors.secondary,
'--color-elevation': colors.elevation,
'--color-overlay': colors.overlay,
```

* Leave alone: `colors.text` and `colors.heading` as generic global tokens are intended to apply to all headings and text. Changing these names to names like `paragraph` and `title` insinuates that the colors apply to less of our output. The original names were not selected because of labels in our largest consuming app, but primarily to define a spec that is more inline with definitions in other systems and agnostic to ever-changing product opinions. Our design tokens and data shouldn't need to change due to a single product UI.
* Add: `colors.primary` and `colors.secondary` is a great addition that provides a reusable mapping to our components, but also subsequently custom components or styling based on globally defined color profiles. We discovered that one piece of data we're setting globally called `outlineButtonColor` is missing in the global data which should directly map to our secondary button variant. So setting / passing both primary and secondary colors would be helpful for buttons and likely future scenarios.
* Change: Our theme-default sets data for `colors.elevation` and `colors.overlay` however, the data we've been passing into the theme object has explicitly included a hyphenated version of the token, despite being part of the same colors object. We should change the theme to accept the same color tokens that are set in the theme-default and be consistent with how all the other colors are named.

#### Fonts (and fontWeights)
```
// SCROLL THIS CODE
// v9 Theme Fonts Data and CSS Properties
// None

// v10 Beta Theme Fonts Data and CSS Properties
fonts: {
    baseSize: 16,
    sizeScale: 1.17,
    title: {
        fontFamily: 'inherit',
        fontWeight: '500',
    },
    paragraph: {
        fontFamily: 'inherit',
        fontWeight: '400',
    },
    label: {
        fontFamily: 'inherit',
        fontWeight: '500',
    },
},

'--font-family-title': fonts.title.fontFamily,
'--font-weight-title': fonts.title.fontWeight,
'--font-family-paragraph': fonts.paragraph.fontFamily,
'--font-weight-paragraph': fonts.paragraph.fontWeight,
'--font-family-label': fonts.label.fontFamily,
'--font-weight-label': fonts.label.fontWeight,

// v10 Recommended Theme Fonts Data and CSS Properties
fonts: {
    baseSize: 16,
    sizeScale: 1.17,
    body: 'system-ui',
    heading: 'system-ui',
    headline: 'system-ui',
    label: 'system-ui',
},
fontWeights: {
    body: 400,
    heading: 600,
    headline: 400,
    label: 500,
},

'--font-body': fonts.body,
'--fontWeight-body': fontWeights.body,
'--font-heading': fonts.heading,
'--fontWeight-heading': fontWeights.heading,
'--font-label': fonts.label,
'--fontWeight-label': fontWeights.label,
```

* Change: (Worth debating) Simplify the fonts object and split into `fonts` and `fontWeights`, makes our tokens simpler and easier to type check and automate processing in the future. This data format is also compatible with the Theme UI spec which is becoming more ubiquitous in more systems. But the primary benefit is that this format is easier to type and remember: `fonts.body` is simpler than `fonts.paragraph.fontFamily`
* Change: `body` instead of `paragraph`. The default font family and weight is intended to apply to everything in the theme, not just paragraphs. Defining this as the `paragraph` increases confusion as where this style will apply, where `body` is a far more common way of defining this default as seen in numerous other design systems and type tools [[1](https://m3.material.io/styles/typography/tokens#d74b73c2-ac5d-43c5-93b3-088a2f67723d)][[2](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/)][[3](https://www.carbondesignsystem.com/guidelines/typography/type-sets#body-styles)][[4](https://chakra-ui.com/docs/styled-system/theming/theme?scroll=true#typography)][[5](https://spectrum.adobe.com/page/typography/)][[6](https://kyleamathews.github.io/typography.js/)]
* Change: `heading` instead of `title`. Over the years we've found ourselves coming up with different terms for essentially the same thing. From an editorial perspective there's subheading, subtitle, headline, headers, titles, etc. But from a semantic HTML perspective, these are all just different labels for essentially setting a h1-h6 (i.e. [Heading](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)) tags in a document. Heading is more generic and ubiquitous as an all encompassing term. It's also a far more common way of defining this token in numerous other design systems [[1](https://www.lightningdesignsystem.com/design-tokens/#category-font)][[2](https://spectrum.adobe.com/page/typography/)][[3](https://www.carbondesignsystem.com/guidelines/typography/type-sets#fluid-heading-styles)][[4](https://chakra-ui.com/docs/styled-system/theming/theme?scroll=true#typography)] It's worth noting that systems like Material / Apple when providing multiple typographic styles for "headings" they have made distinctions of `Display || Headline || Title`, so perhaps this is worth some debate. But I still think looking at the full landscape, that `heading` is more common (and why we had a Heading component, instead of a Title component, to begin with)
*  Change: CSS property naming conventions. As we defined with the `colors` we've begun to establish a good pattern of `[tokenCategory].[token].[value]` and making our CSS properties consistent with those names. We've begun to establish the singular form of the token category. So `colors.primary` easily translates to `--color-primary`. As we look at our fonts, we should do the same thing. `fonts.body` easily translates to `--font-body`. Depending on some decisions make around the font vs fontWeight data structure -- this translation to the CSS property should be maintained. In the short-term for consistency, but in the long-term so we could automatically generate these global tokens into CSS more easily. It's worth some additional discussion about how we feel about a mix of pascal + hyphens in CSS properties. A lot of systems leveraging tools like styled components, emotion, stitches all translate design tokens wholesale where pascal token names like `lineHeights` in the CSS properties. I'm not as strongly opinionated about this since we can currently manually translate these object keys to hyphen and later automate that, but it's likely more common in other systems to just use hyphens.

### Additional considerations and feedback from testing beta + labs
* It appears that setting the button patterns data in the default theme doesn't work as expected. In this PR, I'm setting default colors for the primary / secondary variants, but even [testing these changes manually on the beta](https://github.com/square/maker/compare/beta...beta-pattern-color-test) they aren't working.
* In the Theme lab, I recently started testing `beta` with a nested theme. I was attempting to test theme reactivity and mimic our internal `ThemeBlock` component.  I'm not sure if I have it setup correctly, but it's not reactive to changes to the parent theme. (I'll push up a lab deploy that shows the issue in isolation outside of this PR)
* Setting the default font families to `inherit` makes things work a bit differently than expected. While the text component resolves the theme data and inlines the font family, that's not happening everywhere -- even with the theme setting a default font. I figured we should be a bit more opinionated and set a default to `system-ui`, it might be masking the issue, but it appears to work better. (I'll push up a lab deploy that also shows this)
* I think exposing a `pattern` concept as an alternative to both global design tokens and `variants` is premature, particularly since this concept is limited to two components where one is already using variants. While I appreciate it's solving a problem, I think we should keep the concept internal until we can refactor our code to have a more robust theming system with variants. `variants` are by far the most common way of documenting how to get alternative styles on a component in every system. We can add variants to the text component the same way we are doing it for buttons. The public API can remain consistent and meets expectations for the prop to allow alternative styling options.
* Since we know that others are using what was previously considered internal CSS properties and that changes to those properties now constitute breaking changes -- we should consider prefixing CSS properties (and eventually classes) we're generating to prevent collisions. `maker-[token]` for example.


pretzelhammer:
I've updated the PR title from `refactor` to `refactor!` because this contains breaking changes

pretzelhammer:
> It appears that setting the button patterns data in the default theme doesn't work as expected. In this PR, I'm setting default colors for the primary / secondary variants, but even [testing these changes manually on the beta](https://github.com/square/maker/compare/beta...beta-pattern-color-test) they aren't working.

That branch isn't working because the Buttons styleguide doesn't use patterns yet, it's still on variants. I've made a [new branch which updates the Button styleguide to use patterns](https://github.com/square/maker/compare/beta-pattern-color-test...beta-pattern-color-test-fix) so it's now responsive to changes in the default theme.

pretzelhammer:
> Since we know that others are using what was previously considered internal CSS properties and that changes to those properties now constitute breaking changes -- we should consider prefixing CSS properties (and eventually classes) we're generating to prevent collisions. maker-[token] for example.

This is reasonable and should be its own PR. I've made a PR for it here: https://github.com/square/maker/pull/305

src/components/Theme/src/Theme.vue
```diff
@@ -56,27 +56,28 @@ export default {
...
 		styles() {
-			const { colors, fonts } = this;
+			const { colors, fonts, fontWeights } = this;
 
 			return {
+				'--color-background': colors.background,
+				'--color-text': colors.text,
+				'--color-heading': colors.heading,
+				'--color-primary': colors.primary,
+				'--color-secondary': colors.secondary,
```
laurenhu, line 66: Since we're erring on the side of not adding necessary variables, I would not add a secondary color. There's no consistent use case for it throughout Website, in fact it doesn't define one. More often than not, outline button color and text or title color are the same. It's also a very specific definition - the color for secondary/outline buttons, not a bigger pattern.

src/components/Theme/src/default-theme.js
```diff
@@ -48,9 +47,11 @@ export default function defaultTheme() {
 			patterns: {
 				primary: {
 					variant: 'primary',
+					color: '@colors.primary',
 				},
 				secondary: {
 					variant: 'secondary',
+					color: '@colors.secondary',
```
laurenhu, line 54: I would default this to paragraph color for the default theme and let website override it if necessary.

src/components/Theme/src/Theme.vue
landondurnan: We could immediately use it for things like buttons and links. While we label the colors in website currently as solid and outline buttons -- there's a benefit to abstract those colors to be useable for more than buttons. There's also precedent as it's a pretty common convention to have a secondary color in the core set of colors in many systems that allow color customizations:

* https://getbootstrap.com/docs/5.1/customize/color/
* https://m3.material.io/styles/color/the-color-system/key-colors-tones

We currently aren't able to leverage the secondary button color we allow people to customize in the profile without a lookup and passing it in. I don't think at this point we should be exposing explicit and inconsistent methods for styling the components through a combination of generic tokens, props, variants, and now patterns as it creates additional confusion about how the theming system works while it's still in a state flux. 

I believe the safest way for us to expand our theming options is through the global / core tokens for colors, fonts, and shape. From there we can internally change and refine how they are used and applied to the components until we're at a point of exposing maximum flexibility for styling the components in a consistent way.

pretzelhammer:
> Change: Our theme-default sets data for colors.elevation and colors.overlay however, the data we've been passing into the theme object has explicitly included a hyphenated version of the token, despite being part of the same colors object. We should change the theme to accept the same color tokens that are set in the theme-default and be consistent with how all the other colors are named.

I think this is probably an oversight or a bug. I think this is reasonable on its own standalone task so I've made a PR for it: https://github.com/square/maker/pull/307

src/components/Theme/src/Theme.vue
laurenhu: How does the abstraction it line up? Is `colors.primary` always the solid button color?

If so, having a `primary` and `secondary` Maker color could cause some confusion on Website, where `--primary-color` always means brand color and `--secondary-color` is deprecated on older sites. Whereas `--maker-primary-color` means the color of the primary CTA which could be different than the brand color.

pretzelhammer:
> Leave alone: `colors.text` and `colors.heading` as generic global tokens are intended to apply to all headings and text. Changing these names to names like paragraph and title insinuates that the colors apply to less of our output. The original names were not selected because of labels in our largest consuming app, but primarily to define a spec that is more inline with definitions in other systems and agnostic to ever-changing product opinions. Our design tokens and data shouldn't need to change due to a single product UI.

We have `title` and `paragraph` text patterns, and the colors apply to those patterns, so it makes sense that the `title` color applies to the `title` text pattern. I think it's reasonable to keep the `beta` names as they are. Also, our design and data shouldn't need to change to fall in line with some arbitrary 3rd-party open source spec.

> Add: `colors.primary` and `colors.secondary` is a great addition that provides a reusable mapping to our components, but also subsequently custom components or styling based on globally defined color profiles. We discovered that one piece of data we're setting globally called outlineButtonColor is missing in the global data which should directly map to our secondary button variant. So setting / passing both primary and secondary colors would be helpful for buttons and likely future scenarios.

We have `colors.primary`, and as Lauren said we don't really need a `colors.secondary`. The convention at the moment is to use `colors.paragraph` for any secondary/outline buttons, both internally in Maker and within Website. If Website needs to customize this they can override it within the button secondary pattern.

> Change: (Worth debating) Simplify the fonts object and split into fonts and fontWeights, makes our tokens simpler and easier to type check and automate processing in the future. 

I disagree. I find this harder to write and harder to cognitively parse:

```js
fonts: { // is really only font family data, not general font data
    headline: 'etc',
    title: 'etc',
    paragraph: 'etc',
    label: 'etc',
},
fontWeights: {
    headline: '700',
    title: '500',
    paragraph: '400',
    label: '500',
},
fontColors: {
   headline: 'etc',
   title: 'etc',
   paragraph: 'etc',
   label: 'etc',
}
```

Than this:

```js
fonts: {
    headline: {
        family: 'etc',
        color: 'etc',
        weight: '700',
    },
    title: {
        family: 'etc',
        color: 'etc',
        weight: '500',
    },
    paragraph: {
        family: 'etc',
        color: 'etc',
        weight: '400',
    },
    label: {
        family: 'etc',
        color: 'etc',
        weight: '500',
    },
},
```

In the former example, if I want to understand how a headline will be rendered, I have to look at n-different sub-objects stitch it together in my head. In the latter example I just have to look at one place and all the headline data is there. The latter is much simpler imo.

> - Change: `body` instead of `paragraph`.
> - Change: `heading` instead of `title`.

Since we have `title` and `paragraph` text patterns I think it's nice to keep the global names consistent with the individual pattern names.

> Change: CSS property naming conventions. As we defined with the colors we've begun to establish a good pattern of `[tokenCategory].[token].[value]` and making our CSS properties consistent with those names. We've begun to establish the singular form of the token category. So `colors.primary` easily translates to `--color-primary`. As we look at our fonts, we should do the same thing. `fonts.body` easily translates to `--font-body`. Depending on some decisions make around the font vs fontWeight data structure -- this translation to the CSS property should be maintained. 

No. We want to maintain a layer of abstraction between the theme spec & the public CSS API. We should never make any promises about how one is turned into the other. This is especially important if we ever want to make changes to the theme spec or the public CSS API without affecting the other. It's further important if we want to add design tokens to the theme spec that do not have a 1:1 relationship to some existing CSS property. A good example of this is shape. We may add `shape: 'squared' | 'rounded' | 'pill'` to the theme spec and it makes no sense to expose a `--shape: 'squared' | 'rounded' | 'pill'` to the public CSS API because that is not a valid CSS property.

> I think exposing a `pattern` concept as an alternative to both global design tokens and `variants` is premature, particularly since this concept is limited to two components where one is already using variants. While I appreciate it's solving a problem, I think we should keep the concept internal until we can refactor our code to have a more robust theming system with `variants`.

`patterns` === `variants`. They're the same thing, just have a different name.

landondurnan:
Much of this PR has been addressed in other smaller separate PRs.

PR #306, opened 2022-04-20
fix(notice): background color for variant block

landondurnan:

CleanShot 2022-04-20 at 15 50 42

When notice component was used in a theme with the block variant it was always gray.

The notice background color has the incorrect CSS property fallback, setting to a neutral color (when used in the theme) as the default.

Removes the neutral color as the primary css property, instead relying on the color-bg variable set in each of the notice types. It's always there so don't need to have a fallback.

## Other information

This wasn't caught in doc dev testing for the notice since there's no theme to set the neutral colors. We should consider setting up some testing environments with doc dev to easily switch between a theme context and no theme context.

### DIFF
--- a/src/components/Notice/src/Notice.vue
+++ b/src/components/Notice/src/Notice.vue
@@ -185,6 +185,6 @@ export default {
 
 .variant_block {
 	padding: 16px;
-	background-color: var(--neutral-10, var(--color-bg));
+	background-color: var(--color-bg);
 }
 </style>


landondurnan:
Because of the simplicity of this bug fix and it's blocking another dev, I'm going to use admin privileges get the patch out.

kiasquare (approved):
Thank you so much Landon!!

PR #310, opened 2022-04-27
feat(shapes): theming support for shapes

laurenhu:
Adds theming support for shape across multiple maker components.

- `shapes` prop on default theme where default component shape and theme border radius sizes are defined
```
		shapes: {
			default: 'rounded',
			radii: {
				small: '8px',
				medium: '16px',
				large: '32px',
			},
		},
```
- Add global maker border radius variables to `Theme`, `--maker-border-radius` and `--maker-border-radius-button` for all shape components that don't have a local `shape` override.
- Update component styles to reflect new shape updates.
- Update `ThemeTest` lab to demo changes in shape controls

### DIFF
--- a/src/components/SegmentedControl/src/SegmentedControl.vue
+++ b/src/components/SegmentedControl/src/SegmentedControl.vue
@@ -2,7 +2,6 @@
 	<div
 		:class="[
 			$s.Container,
-			$s[`shape_${shapeInner}`],
 			$s[`size_${sizeInner}`],
 		]"
 	>
@@ -32,14 +31,6 @@ export default {
 			required: true,
 		},
 		/**
-		 * Shape of Control & Segments
-		 */
-		shape: {
-			type: String,
-			default: 'rounded',
-			validator: (shape) => ['squared', 'rounded', 'pill'].includes(shape),
-		},
-		/**
 		 * Size of Control & Segments
 		 */
 		size: {
@@ -51,7 +42,6 @@ export default {
 	data() {
 		return {
 			currentValue: this.selected,
-			shapeInner: this.shape,
 			sizeInner: this.size,
 		};
 	},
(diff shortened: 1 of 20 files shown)


pretzelhammer (approved):
I think all these:
```vue
// in computed prop
'--border-radius-small': this.theme.shapes.radii.small,
'--border-radius-medium': this.theme.shapes.radii.medium,
'--border-radius-large': this.theme.shapes.radii.large,
// in style
border-radius: var(--border-radius-small, 8px);
```

can be refactored to plain CSS vars, e.g.
```vue
// in Theme component
'--maker-border-radius-small': shapes.radii.small,
'--maker-border-radius-medium': shapes.radii.medium,
'--maker-border-radius-large': shapes.radii.large,
```

and then in the individual components:
```vue
// no additional work in computed prop
// then in style
border-radius: var(--maker-border-radius-small, 8px);
```

the only reason i suggest this is because these radii values don't seem to be associated with any component local props and the components don't seem to do any additional processing/validating of these radii values when they receive them, other than turning them into CSS vars. generally when some theming element meets these requirements we would normally create a theme-global CSS var for it, but this code is fine to merge as-is if you have a good reason for doing it the way you did or if you're on a tight schedule.

src/components/SegmentedControl/src/SegmentedControl.vue
```diff
@@ -31,14 +30,6 @@ export default {
 			type: undefined,
 			required: true,
 		},
-		/**
-		 * Shape of Control & Segments
-		 */
-		shape: {
```
landondurnan, line 37: Removing the shape props on components will constitute a breaking change. While our largest consuming product won't have these options available as product controls, we could leave these props alone as other consumers could benefit from them.

But even if we are set on removing them, which is reasonable, we'll need to plan for a major release and document those changes accordingly.

PR #311, opened 2022-04-28
refactor!: some font-related config renames

pretzelhammer:

reaching team alignment, using more generic names in our global theme config for typography related stuff

renames `title` to `heading` and `paragraph` to `body`, also updates some docs

## Other information

https://square.github.io/maker/styleguide/some-font-color-renames/#/Theme

### DIFF
--- a/src/components/Theme/README.md
+++ b/src/components/Theme/README.md
@@ -30,17 +30,17 @@ Any components within the theme component will inherit the tokens provided. We r
 			</label>
 			<label class="color-option">
 				<input
-					v-model="theme.colors.title"
+					v-model="theme.colors.heading"
 					type="color"
 				>
-				Title
+				Heading
 			</label>
 			<label class="color-option">
 				<input
-					v-model="theme.colors.paragraph"
+					v-model="theme.colors.body"
 					type="color"
 				>
-				Paragraph
+				Body
 			</label>
 			<m-text
 				pattern="title"
@@ -84,16 +84,16 @@ Any components within the theme component will inherit the tokens provided. We r
 				pattern="title"
 				:size="0"
 			>
-				Title
+				Heading
 			</m-text>
 			<div class="font-choice">
 				<m-select
-					v-model="theme.fonts.title.fontFamily"
+					v-model="theme.fonts.heading.fontFamily"
 					class="family-choice"
 					:options="fontOptions"
 				/>
 				<m-select
-					v-model="theme.fonts.title.fontWeight"
+					v-model="theme.fonts.heading.fontWeight"
 					:options="defaultWeights"
 				/>
 			</div>
@@ -101,16 +101,16 @@ Any components within the theme component will inherit the tokens provided. We r
 				pattern="title"
 				:size="0"
 			>
-				Paragraph
+				Body
 			</m-text>
 			<div class="font-choice">
 				<m-select
-					v-model="theme.fonts.paragraph.fontFamily"
+					v-model="theme.fonts.body.fontFamily"
 					class="family-choice"
 					:options="fontOptions"
 				/>
 				<m-select
-					v-model="theme.fonts.paragraph.fontWeight"
+					v-model="theme.fonts.body.fontWeight"
 					:options="defaultWeights"
 				/>
 			</div>
@@ -335,8 +335,8 @@ export default {
 			theme: {
 				colors: {
 					background: '#ffffff',
-					paragraph: '#000000',
-					title: '#000000',
+					body: '#000000',
+					heading: '#000000',
 					primary: '#000000',
 					'neutral-0': '#ffffff',
 					'neutral-10': '#f1f1f1',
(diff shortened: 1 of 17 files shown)


src/components/Theme/README.md
```diff
@@ -348,14 +348,18 @@ export default {
 				fonts: {
 					baseSize: 16,
 					sizeScale: 1.15,
-					title: {
+					heading: {
```
landondurnan, line 351: How are we supporting the fourth option for `headline`?

src/components/Theme/src/Theme.vue
```diff
@@ -67,14 +67,14 @@ export default {
...
 				'--maker-color-elevation': colors.elevation,
 				'--maker-color-overlay': colors.overlay,
-				'--maker-font-family-title': fonts.title.fontFamily,
-				'--maker-font-weight-title': fonts.title.fontWeight,
-				'--maker-font-family-paragraph': fonts.paragraph.fontFamily,
-				'--maker-font-weight-paragraph': fonts.paragraph.fontWeight,
+				'--maker-font-family-heading': fonts.heading.fontFamily,
+				'--maker-font-weight-heading': fonts.heading.fontWeight,
+				'--maker-font-family-body': fonts.body.fontFamily,
+				'--maker-font-weight-body': fonts.body.fontWeight,
```
landondurnan, line 77: I think we should flip the names to be consistent with the data structure, like colors above.

`colors.heading` == `--maker-color-heading`
`fonts.heading.fontFamily` should be `--maker-font-heading-font-family`

I also think the `fontFamily` and `fontWeight` could drop one of the words since it's part of that object. So shortening it further:

`fonts.heading.family` && `fonts.heading.weight` == `--maker-font-heading-family` && `--maker-font-heading-weight` 

this would require changing the theme object as well to shorten it -- but both I think would simplify the data / properties.

src/components/Theme/README.md
pretzelhammer: maker does not use any headlines in any components

src/components/Theme/src/Theme.vue
pretzelhammer: it's already consistent: `[prefix]-[category/property]-[name]`. `prefix` = `maker`, `category/property` = `color|font-family|font-weight`, `name` = some descriptive name for the variable. all of the variables follow this format.

src/components/Theme/README.md
pretzelhammer: heading applies to both title and headline patterns

pretzelhammer: https://github.com/square/maker/pull/311/files#diff-b34aee85b50fd2933be427ceddd7edcdb6c5b6eec78d8ae56cbe88838d98df68R83-R84

and

https://github.com/square/maker/pull/311/files#diff-b34aee85b50fd2933be427ceddd7edcdb6c5b6eec78d8ae56cbe88838d98df68R90-R91

src/components/Theme/src/Theme.vue
pretzelhammer: i don't think we should ever tightly-couple the shape of the theme data to how the public CSS variables are named, otherwise it will turn any tiny changes to either into breaking changes. we should have a layer of abstraction/indirection between the two.

landondurnan: I might be overthinking this a bit. I do agree that at face-value one CSS property is easier to read than the other. But I don't understand the motivation around not tightly coupling the data? My understanding is that would make both our input and output more consistent.

I wouldn't necessarily describe keeping the JSON / CSS conversion as "tightly-coupled" -- but more of a direct translation between the theme data and how that gets converted into reusable data, in our case with the CSS properties.

This is a pattern we can look at in tools like [Style Dictionary](https://amzn.github.io/style-dictionary/#/tokens) and [Adobe DSP](https://github.com/demianborba/spectrum-dsp), where the data coming in has a direct correlation to what's coming out in the CSS. In fact, I think we could in the future leverage the same tools to automatically store, validate, process, and document our theme tokens for usage inside of Maker components and when creating custom components along-side Maker.

So for example with the DSP reference with Spectrum design system: 
* https://github.com/demianborba/spectrum-dsp/blob/master/data/fonts.json (Input)
* https://github.com/demianborba/spectrum-dsp/blob/master/dist/styledictionary/css/variables.css#L94-L105 (CSS property output)

Or Material 3's new tokens: 
* https://github.com/material-foundation/material-tokens/blob/main/dsp/data/fonts.json
* https://github.com/material-foundation/material-tokens/blob/main/dsp/dist/styledictionary/css/variables.css#L171-L205

There's definitely some tooling and conversions that we're not using and isn't totally relevant to us right now, but it could be in the future and that's why basing some of our decisions on these solutions could be helpful now so that our principles about adding these and any future values are consistent. I think the input and output should be the same, the only difference is the format of JSON to CSS.  Here's another good reference about the naming conventions and how different systems look at how that converts to tokens: https://medium.com/eightshapes-llc/naming-tokens-in-design-systems-9e86c7444676

But as a quick reference, here's similar CSS properties from Spectrum and Material 3 that is being generated from DSP / Style dictionary:

```
--spectrum-global-font-body-family: Adobe Clean;
--spectrum-global-font-body-style: Regular;

--md-sys-typescale-body-medium-family: Roboto;
--md-sys-typescale-body-medium-weight: 400;
```

I think the potential mixup here is that our category `font` also sounds like a property. And because the property also has the word `font` in it.  So we could simplify the data to something like `fonts.body.family` which would logically make sense as a property to be `--maker-font-body-family`

pretzelhammer: i still think that `--maker-font-family-heading` and `--maker-font-weight-heading` is just easier and simpler to read than `--maker-font-heading-family` and `--maker-font-heading-weight`

pretzelhammer: resolved in slack huddle

PR #313, opened 2022-05-02
refactor(pininput): uses blockformcontrollayout internally now

pretzelhammer:

makes pininput code more standard compared to the other block form items in maker

refactors pininput to use blockformcontrollayout internally

## Other information

nope

### DIFF
--- a/src/components/PinInput/README.md
+++ b/src/components/PinInput/README.md
@@ -305,9 +305,10 @@ export default {
 
 ## Slots
 
-| Slot  | Description      |
-| ----- | ---------------- |
-| error | Input error slot |
+| Slot    | Description             |
+| ------- | ----------------------- |
+| default | —                       |
+| error   | slot for error messages |
 
 
 ## Events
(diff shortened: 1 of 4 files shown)


laurenhu (commented):
@Ryguy11 could you review and confirm all the events/methods you need to access on the old `PinInput` component are still available on the new PinInputFormItem wrapper? I see an `this.$emit('complete', pin);` event on the now nested PinInputControl that we might want to listen for and emit from PinInputFormItem. Please advise if there are any others.

pretzelhammer:
@laurenhu the `complete` event is covered because the wrapper component proxies all event listeners to the inner component: https://github.com/square/maker/pull/313/files#diff-f1d6cc8686e2e62c3cc6cf3fa48593e0000861d9e8655916047fe206ff22dee0R8

I manually tested all of the docs examples myself and have confirmed they work as expected after the refactor: https://square.github.io/maker/styleguide/refactor-pininput/#/PinInput

laurenhu (approved):

PR #316, opened 2022-05-06
feat(stepper): allow manual number entry for stepper

### LINKED ISSUE #317: Allow manual entry for stepper
### Feature request

Expose a number input in between the stepper buttons to allow manual entry.

### Why?

When users need to change the stepper value by a large amount, it can be cumbersome/annoying to have to hit the buttons many many times. Exposing an input would allow a shortcut to increase/decrease by larger amounts.

### Alternatives

_No response_

### Additional context

_No response_

tatems:
Setting a stepper input to a much higher or lower number requires tapping the `+`/`-` buttons many times to get to the correct value.

Closes #317

This PR adds the ability to manually type in a value for a stepper in order to side step having to tap the increment/decrement buttons many times.
Clicking/tapping the inner number will create an input that will autofocus and (for mobile) pop open the numeric keyboard.
Manual inputs are validated to be an actual number and be within the min/max range (if defined).
The updated value will not be emitted until the user blurs/hits enter.

https://user-images.githubusercontent.com/3402466/167197148-ad631c88-7ff0-4418-ab46-7a3a991bfeac.mov

### DIFF
--- a/src/components/Stepper/README.md
+++ b/src/components/Stepper/README.md
@@ -7,7 +7,7 @@
 	>
 		<m-stepper
 			v-model="number"
-			min="0"
+			min="-10"
 			max="10"
 		/>
 		v-model: {{ number }}
@@ -46,7 +46,7 @@ export default {
 
 ## Events
 
-| Event          | Type     | Description           |
-| -------------- | -------- | --------------------- |
-| stepper:update | `number` | updated stepper value |
+| Event          | Type | Description |
+| -------------- | ---- | ----------- |
+| stepper:update | -    | —           |
 <!-- api-tables:end -->
--- a/src/components/Stepper/README.md
+++ b/src/components/Stepper/README.md
@@ -7,7 +7,7 @@
 	>
 		<m-stepper
 			v-model="number"
-			min="-10"
+			min="0"
 			max="10"
 		/>
 		v-model: {{ number }}
--- a/src/components/Stepper/src/Stepper.vue
+++ b/src/components/Stepper/src/Stepper.vue
@@ -152,10 +152,6 @@ export default {
 				return;
 			}
 			const incrementBy = 1;
-			/**
-			 * updated stepper value
-			 * @property {number}
-			 */
 			this.emitUpdate(this.value + incrementBy);
 		},
 		decrement() {
@@ -163,10 +159,6 @@ export default {
 				return;
 			}
 			const decrementBy = 1;
-			/**
-			 * updated stepper value
-			 * @property {number}
-			 */
 			this.emitUpdate(this.value - decrementBy);
 		},
 
@@ -183,17 +175,33 @@ export default {
 			event?.preventDefault?.();
 			event?.stopPropagation?.();
 
-			const newValue = Number(this.manualValue);
+			const newValue = Number.parseInt(this.manualValue, 10);
 			this.isSettingManualValue = false;
 
 			if (Number.isNaN(newValue) || newValue === null || newValue === undefined) {
 				return;
 			}
 
-			this.emitUpdate(newValue);
+			this.emitUpdate(this.validateManualValueInRange(newValue));
+		},
+
+		validateManualValueInRange(newValue) {
+			if (!Number.isNaN(this.minVal) && newValue < this.minVal) {
+				return this.minVal;
+			}
+
+			if (!Number.isNaN(this.maxVal) && newValue > this.maxVal) {
+				return this.maxVal;
+			}
+
+			return newValue;
 		},
 
 		emitUpdate(newValue) {
+			/**
+			 * updated stepper value
+			 * @property {number}
+			 */
 			this.$emit('stepper:update', newValue);
 		},
 	},
--- a/src/components/Stepper/src/Stepper.vue
+++ b/src/components/Stepper/src/Stepper.vue
@@ -13,9 +13,30 @@
 		>
 			<minus :class="$s.Icon" />
 		</m-button>
-		<span :class="$s.Quantity">
-			{{ value }}
-		</span>
+		<div
+			:class="$s.Quantity"
+		>
+			<input
+				v-show="isSettingManualValue"
+				ref="manualInput"
+				v-model="manualValue"
+				:class="$s.QuantityManualInput"
+				:min="min"
+				:max="max"
+				type="number"
+				inputmode="numeric"
+				align="center"
+				@blur="commitManualValue"
+				@keyup.enter="commitManualValue"
+			>
+			<span
+				:class="$s.QuantityReadonly"
+				@click="triggerManualInput"
+			>
+				<!-- This allows us to auto-resize the input as users type -->
+				{{ isSettingManualValue ? manualValue : value }}
+			</span>
+		</div>
 		<m-button
 			variant="primary"
 			size="small"
@@ -106,6 +127,13 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			manualValue: 0,
+			isSettingManualValue: false,
+		};
+	},
+
 	computed: {
 		...resolveThemeableProps('stepper', ['color', 'textColor', 'shape']),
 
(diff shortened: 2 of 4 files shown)


src/components/Stepper/src/Stepper.vue
```diff
@@ -13,9 +13,30 @@
 		>
 			<minus :class="$s.Icon" />
 		</m-button>
-		<span :class="$s.Quantity">
-			{{ value }}
-		</span>
+		<div
+			:class="$s.Quantity"
+		>
+			<input
```
landondurnan, line 19: We should run this functionality past design. 

I'm making some assumptions, but I think we'd get better styling / states / and interaction behavior by using our Maker input internally here.

landondurnan (changes requested):
Before adding new features to Maker we should have a request submitted: https://github.com/square/maker/issues/new?assignees=&labels=feature+request%2Cpending+triage&template=FEATURE_REQUEST.yml

Even if you're immediately planning on doing the work, it can help give the team a heads up so we can discuss some of the details and give some recommendations. 

The functionality looks good and it's a nice feature, but we should get aligned with design on the behavior. I added a suggestion regarding using the Maker input, but design will want to weigh in.

tatems:
Whoops, my bad on not creating the issue, I'll add it now for posterity

src/components/Stepper/src/Stepper.vue
tatems: That was my initial approach, but based on the designs I've been given, it's not _currently_ possible with `MInput` in it's current state (eg too big, can't remove border _and_ background). I'll confirm with design if this is still the direction we want to go with.

tatems: Just confirmed, design wants it (visually) how I've implemented.

src/components/Stepper/src/Stepper.vue
```diff
@@ -124,22 +155,57 @@ export default {
...
+			this.isSettingManualValue = true;
+			this.$nextTick(() => {
+				this.$refs.manualInput.focus();
+				this.$refs.manualInput.select();
+			});
+		},
+
+		commitManualValue(event) {
+			event?.preventDefault?.();
+			event?.stopPropagation?.();
```
landondurnan, line 179: I haven't seen optional chaining like this on event handling before. Events should always exist and have the same properties, so is this necessary?

src/components/Stepper/README.md
```diff
@@ -46,7 +46,7 @@ export default {
 
 ## Events
 
-| Event          | Type     | Description           |
-| -------------- | -------- | --------------------- |
-| stepper:update | `number` | updated stepper value |
+| Event          | Type | Description |
+| -------------- | ---- | ----------- |
+| stepper:update | -    | —           |
```
landondurnan, line 51: Unfortunate this got removed, but since they're autogenerated I'm not sure how we can put these back in. @pretzelhammer do you know how we should handle event documentation?

src/components/Stepper/src/Stepper.vue
```diff
@@ -124,22 +155,57 @@ export default {
...
+
+			const newValue = Number.parseInt(this.manualValue, 10);
+			this.isSettingManualValue = false;
+
+			if (Number.isNaN(newValue) || newValue === null || newValue === undefined) {
```
landondurnan, line 184: Not sure if / what we could do about this, but there's a few characters that get through:

CleanShot 2022-05-06 at 12 27 38

src/components/Stepper/src/Stepper.vue
```diff
@@ -13,9 +13,33 @@
...
+				ref="manualInput"
+				v-model="manualValue"
+				:class="$s.QuantityManualInput"
+				:min="min"
+				:max="max"
+				type="number"
+				inputmode="numeric"
+				align="center"
+				@blur="commitManualValue"
+				@keyup.enter="commitManualValue"
```
landondurnan, line 30: Pro / cons either way, but removing the `enter` on this would catch invalid data on each keyup:
CleanShot 2022-05-06 at 12 29 27

landondurnan (commented):
Tested on a few browsers both mobile / desktop and everything is working as expected. Good work. 

I through some curve balls at it and there might be an opportunity to make some minor improvements noted below.

src/components/Stepper/README.md
pretzelhammer: hmmm, this is already properly documented here, not sure why the auto-generator isn't catching it: https://github.com/square/maker/pull/316/files#diff-1844fa4f25a196f09e947190a8f5d759a44d6a6e0889f6f2905d915e1218e46fR204-R208

src/components/Stepper/src/Stepper.vue
```diff
@@ -13,9 +13,33 @@
...
+			>
+			<span
```
pretzelhammer, line 32: instead of having `<input v-if="editing"><span v-else>` you can have `<input :disabled="!editing">` and then you don't need the `span` (since you already styled the `input` to look like a `span` anyway). i think this should work, if not lemme know.

src/components/Stepper/README.md
pretzelhammer: anyway, fixing this is not a blocker for this PR, so I'm gonna resolve this thread

src/components/Stepper/src/Stepper.vue
pretzelhammer: `e` is a valid number value, used for writing exponents in scientific notation, e.g. `1e2` = 100. A small issue is that the `Number.parseInt` doesn't correctly handle values with `.` or `e` characters in them :/

image

src/components/Stepper/src/Stepper.vue
```diff
@@ -124,22 +155,57 @@ export default {
...
+				this.$refs.manualInput.focus();
+				this.$refs.manualInput.select();
+			});
+		},
+
+		commitManualValue(event) {
+			event?.preventDefault?.();
+			event?.stopPropagation?.();
+
+			const newValue = Number.parseInt(this.manualValue, 10);
```
pretzelhammer, line 181: since `<input type="number">` lets characters like `.` and `e` get entered we should change this to:
```
const newValue = Math.round(Number.parseFloat(this.manualValue, 10));
```
which correctly handles those inputs, unlike `Number.parseInt`. An alternative solution would be to prevent people from entering `.` or `e` into the input by adding an `input` or `change` event listener onto the `input` element and do something like:
```js
allowCharacter(event) {
  return /\d/.test(event.key); // return false if non-digit character
}
```

src/components/Stepper/src/Stepper.vue
laurenhu: I agree - do we need to have a separate span? Seems like it'd be simpler to just have the input.

tatems: Yeah you're right, I definitely can!

tatems: Actually, scratch that. I was using the input to auto-resize the input based on the value.

src/components/Stepper/src/Stepper.vue
tatems: Good suggestion, I'll do that.

src/components/Stepper/src/Stepper.vue
tatems: I think I'm going to switch over the `blur` and `keyup.enter` since it seems to mimic this behaviour anyway.

src/components/Stepper/src/Stepper.vue
tatems: I was being overly cautious here, I'll update

src/components/Stepper/src/Stepper.vue
pretzelhammer: We're suggesting to get rid of the `span` and leave only the `input`, is that not possible? 🤔

tatems: To clarify, the reason I've left this is to allow the input to be resized.
Basically, the `span` will adjust to the width of the entered content as the user updates the value. The `input` width is tied to the containers width, and so it will resize with those changes.
This seemed like the least hacky solution to get the input to automatically resize itself, but I'm open to other suggestions to get the same effect. Otherwise, the input will have to be a set width, which might result in a jankier experience.

pretzelhammer (approved):

PR #322, opened 2022-05-11
feat(notice): more themable notices

pretzelhammer:

contextual colors for notices / text buttons

read diff

## Other information

https://square.github.io/maker/styleguide/theme-notices/#/Notice


laurenhu (approved):
When we eventually move the color generation utils back into Maker, I'd like to update the README for this as well as other contextual color components to include a color picker so adaptation to bg can be previewed in real time.

But that can be part of that follow up task. This looks great!

src/components/Theme/src/default-theme.js
```diff
@@ -15,6 +15,21 @@ export default function defaultTheme() {
 			elevation: '#ffffff',
 			overlay: 'rgba(0, 0, 0, 0.3)',
 			primary: '#000000',
+			critical: {
+				fill: '#cd2026',
+				text: '#a82826',
+				subtle: '#f6eceb',
```
landondurnan, line 21: It's not clear in the code or documentation how fill or subtle are used. We're directly referencing text, but how do the others work?

landondurnan, line 21: We should create more documentation around these colors and how they are used. I understand text, but the term fill and subtle are less common names.

It might be good to get aligned with design on nomenclature and consider how we're documenting the terms and values for both design and engineering. I can't find a reference in our design documentation.

landondurnan (commented):
We may want to consider how we could define default contextual colors for what we consider a light background and dark background. There's also some internal discussion around moving our neutral colors to be generated inside Maker (or offering some utilities), it feels very similar to how we may want to try to automate the definition of contextual colors.

The flexibility to define all the contextual colors is great, but there could be opportunity to simplify the input based on a single color. I'm not sure the logic design used to define the variations on the error colors, but if it's a formula we could automate it internally.

Similar to Material:
image

src/components/Notice/README.md
```diff
@@ -249,11 +253,14 @@ export default {
...
 
-| Prop    | Type     | Default    | Possible values                       | Description    |
-| ------- | -------- | ---------- | ------------------------------------- | -------------- |
-| type    | `string` | `'info'`   | `error`, `success`, `warning`, `info` | type of notice |
-| variant | `string` | `'inline'` | `inline`, `block`                     | notice variant |
-| color   | `string` | —          | —                                     | notice color   |
+| Prop       | Type     | Default    | Possible values                       | Description                        |
+| ---------- | -------- | ---------- | ------------------------------------- | ---------------------------------- |
+| pattern    | `string` | —          | —                                     | pattern defined at theme level     |
+| type       | `string` | —          | `error`, `success`, `warning`, `info` | type of notice                     |
```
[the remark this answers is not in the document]
pretzelhammer: `type` is still necessary to [select the correct icon](https://github.com/square/maker/pull/322/files#diff-072cc9e1481b797412eb201a1a5520c6ffbfb6e2d79de3924bde71255cd2ef50R123-R133) for the notice. if we add a `icon` prop to notice we can remove `type` and replace it entirely with patterns, but doing so would be a breaking change.

src/components/Theme/src/default-theme.js
pretzelhammer: already reached alignment with Design on these, these are the terms Design uses

landondurnan:
The dark theme preview and showing how notices work with different backgrounds is a really nice touch. Showing how we support light / dark backgrounds and the full range of different background colors is a very unique aspect to Maker.

Since we're also talking about moving our contrast methods internally for neutral colors generated, it's very similar to checks and output being done in the notice demo. While we can support manually passing different values, likely the ideal state for all generated colors is to get Maker to do it for the consumer. That way a Maker theme / user can pass the least amount of data possible.

Would it be possible to build in the contrast logic and color generation inside of the Notice component as a start to that optimization? We could have the same demo of how Notices work with different background colors, but we internalize some of the JS code we're documenting as an external requirement to automatically happen internally.

pretzelhammer:
@landondurnan both Lauren and I would like to move more of the color generation code from Website to Maker. We're thinking for starters to expose more util functions from Maker which would generate stuff like neutrals and contextual colors given a background color (similarly to [how we do so in the ThemeTest lab](https://github.com/square/maker/blob/master/lab/experiments/ThemeTest.vue#L576-L606)). That's not within the scope of this PR though. I think the sequencing would likely be:

1. Merge & release this PR
1. Update Maker in Website to use the new themeable contextual MNotices
1. Implement MVP Pill component
1. Update Maker in Website to use new MPill component (which will also support contextual colors)
1. Move Website color generation code into Maker, export the logic as util functions
1. Update Maker in Website, remove all of Website's color generation code and replace it with Maker's

PR #328, opened 2022-05-18
fix(image): fix image skeleton and lazy load

laurenhu:
- Image skeletons transition out while the image transitions in so they both appear at the same time
- The image lazy load causes images to disappear and reappear when they're scrolled out of viewport

- This PR fixes the image skeleton transition and prevents images from disappearing when they're scrolled offscreen.
- It also adds a `lazyload` property to Image and defaults to loading on mount by default


pretzelhammer:
@laurenhu like we briefly discussed on Slack, please remove `isIntersecting` from the local data, and also remove the `onIntersection` method and replace this line:
```js
observer.watch(this.$el, this.onIntersection);
```
with something like:
```js
observer.watch(this.$el, ({ isIntersecting }) => if (isIntersecting) this.load());
```

pretzelhammer (approved):

PR #337, opened 2022-05-23
ci: silence remove worflow error

pretzelhammer:
it's okay to silence the error because it's okay if it fails


pretzelhammer:
using admin privileges to merge since there's not anything to really review here

PR #343, opened 2022-05-25
feat(image)!: update shape property for image

laurenhu:
The current and upcoming shapes conversion work needs Images to support a `shape` prop that supports either `square`, `circle` or `arch` as values.

Updates the local `shape` property on Image to support the following values:
- `square` - this resets the border-radius set by the theme's `shapes.imageBorderRadius` value to `0`
- `circle` - this applies a border radius equal to the image element's height (per designs)
- `arch`- this applies a border radius equal to the image element's height to the top left/right corners of the image

Screen Shot 2022-05-25 at 3 22 19 PM

Screen Shot 2022-05-25 at 3 22 00 PM

Screen Shot 2022-05-25 at 3 22 09 PM

## Migration Guide from 11.x to 12.x

- MImage's `shape` prop has changed:
  - grep your codebase for `<m-image` and/or `shape="` to find all instances that may need to be migrated and then:
  - `squared` value has been renamed to `square` (and should be migrated to `square`)
  - `pill` has been removed (and can be migrated to `square` or new `circle` value)
  - `rounded` has been removed (and can be migrated to `square` or new `circle` value)

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -1,5 +1,8 @@
 <template>
-	<div :class="$s.ImageWrapper">
+	<div
+		ref="image-wrapper"
+		:class="$s.ImageWrapper"
+	>
 		<m-skeleton-block
 			v-if="!loaded"
 			:class="[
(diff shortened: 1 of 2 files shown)


src/components/Image/src/Image.vue
```diff
@@ -81,7 +91,7 @@ export default {
 		shape: {
 			type: String,
 			default: undefined,
-			validator: (shape) => ['squared', 'rounded', 'pill'].includes(shape),
```
pretzelhammer, line 84: renaming `squared` to `square` and removing `rounded` and `pill` counts as a breaking change (even if nobody was using them), we should update the PR title from `feat(image):` to `feat(image)!:`

pretzelhammer (approved):

pretzelhammer:
i added a Migration Guide to the PR description since this PR contains breaking changes, please feel free to update it or change it if I got anything wrong, but otherwise everything else looks good to me, so I approved

PR #347, opened 2022-05-27
fix(image): add original shape option for images

laurenhu:
Adds an `original` image shape option to reset an image within a theme.

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -91,7 +91,7 @@ export default {
 		shape: {
 			type: String,
 			default: undefined,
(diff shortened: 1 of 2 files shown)


src/components/Image/src/Image.vue
```diff
@@ -91,7 +91,7 @@ export default {
 		shape: {
```
pretzelhammer, line 91: please add a JSDoc comment above this prop describing what the different values do and it'll automatically get added to the component's styleguide page via the automated API table generation... it's not immediately obvious why `original` exists or what it does...

pretzelhammer (approved):
approved, but please address [my small comment](https://github.com/square/maker/pull/347/files#r883760062) before merging, thank you!

PR #351, opened 2022-06-06
fix!: improve internal color generation

landondurnan:
BREAKING CHANGE: Dependency changes constitute major release

- We have a dependency on a rather large library that provides a set of color functions we use throughout our components. It's possible to replace these methods with something more lightweight.
- Buttons have rudimentary state changes that could be improved. More specifically, our hover / active states only make the color darker. With a dark color, the state changes aren't visible.

The majority of this work replaces one of our dev / peerDependencies `chroma-js` with an alternative library `colord` that offers the same color methods in a smaller package.

While working on replacing those methods, I noticed an opportunity to cleanup how colors were generated in a few components, namely the Button and ActionBarButton. The hover and active states were updated to provide better output. 

- On primary, depending on the background color will generate either a lighter or darker color. Darker was the only output prior. The same was done on the ActionBarButton.
- On secondary and tertiary rather than a black with different opacity, the background is generated from the color provided at different opacities for hover and active.

See examples:

Before:
CleanShot 2022-06-24 at 17 42 04

After:
CleanShot 2022-06-24 at 17 37 19

## Other information

While our package size remains largely unchanged, this will improve performance for consumers since the peerDependency can be replaced with a smaller library. Accordingly, this is considered a breaking change: https://github.com/semver/semver/issues/502


pretzelhammer:
I'd like to hold off on giving this a full review until [this other PR is merged first](https://github.com/square/maker/pull/352) and this one is updated to accommodate those changes. I think it'll increase the number of color utilities we'd need to support.

A few small comments though:
- I'm not sure if I'm comfortable about us maintaining our own color functions... they're easy to get wrong and none of really understand any of the math behind any of them... which makes it even easier for us to get them wrong.
- Did you consider [the alternative color library Hiroki suggested last year](https://github.com/square/maker/discussions/190)? [colord](https://github.com/omgovich/colord) seems to be significantly smaller and faster than chroma, and probably still suits all of our needs.
- While this PR technically makes Maker smaller as a standalone library... it makes it larger for Website, who is already using chroma for other things. I think we should consider a solution that would make both Maker and Website smaller. A good solution would be to refactor both Website and Maker from chroma to colord. That way both become smaller AND the Maker team doesn't have to assume ownership/maintenance of the color functions.

And one minor nitpick:
- it's not safe to swap out `chroma.valid` for `cssValidator` because the latter is way more permissive. The former allows color value formats like `#rrggbb`, `rgb(r, g, b)`, `hsl(h, s, l)`, and `red` where the latter allows all of those formats plus `hwb(h, w, b)` and `var(--css-custom-property)`. Before this PR the color props would only validate the first 4 formats, and they would all be correctly handled by chroma. After this PR the color prop will now validate all 6 formats but the color logic now only correctly handles colors in `#rrggbb` and `rgb(r, g, b)` format and will break for the remaining 4. If this regression is intentional we would need to mark this PR as a breaking change, if not then it would need to be fixed before this PR is merged.

landondurnan:
colord looks great, I agree maintaining our own color utilities can be a bit nerve-racking. I thought it was worth the trade-off for how little we actually needed. But I'll explore swapping out chroma with colord and see what that looks like.

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -49,27 +49,28 @@
...
-import chroma from 'chroma-js';
+import { colord } from 'colord';
 import PseudoWindow from 'vue-pseudo-window';
 import { MLoading } from '@square/maker/components/Loading';
 import { MThemeKey, defaultTheme, resolveThemeableProps } from '@square/maker/components/Theme';
 import { getContrast } from '@square/maker/utils/get-contrast';
 
-// TODO: refactor the code below so it's shared with Button component
+function setColorVariables(tokens) {
+	const textColor = getContrast(tokens.color, tokens.textColor);
```
laurenhu, line 59: Minor nitpick - can we group the constants and the colors together in separate blocks to be more readable?

src/components/ActionBar/src/ActionBarButton.vue
```diff
@@ -49,27 +49,28 @@
...
 <script>
-import chroma from 'chroma-js';
+import { colord } from 'colord';
 import PseudoWindow from 'vue-pseudo-window';
 import { MLoading } from '@square/maker/components/Loading';
 import { MThemeKey, defaultTheme, resolveThemeableProps } from '@square/maker/components/Theme';
 import { getContrast } from '@square/maker/utils/get-contrast';
 
-// TODO: refactor the code below so it's shared with Button component
+function setColorVariables(tokens) {
```
laurenhu, line 58: Would be great to have this as a util maybe `buttonColorVariables` on maker-colors.js so share between this, the Action Bar Button and potentially Choice and Segmented Control? Maybe taking `stateAdjust` as a prop?

laurenhu (approved):
This is great! Thanks for all the cleanup (and interaction states) on Button. Minor notes - but I think this looks find now and moving `setColorVariables ` to a reusable util can also be done as a follow-up pass along with adding interaction states for Choice and Segmented Control.

src/components/ActionBar/src/ActionBarButton.vue
landondurnan: I thought about that -- but the logic is slightly different right now with the Button since it supports transparent states. I wasn't sure how to support a totally universal utility across all components right now based on the fact that we're conditionally generating different colors based on some internal logic in each component. 

It is definitely possible to generate a more universal utility that could do the same thing for all components. When I started considering a more idealized state perhaps leveraging something like HSLA variables and adjusting output with more CSS, I realized that's a much bigger change than what this task originally set out to do. 

It'd likely be better to tackle that utility separately and look at this PR as a step working us towards that.

src/components/ActionBar/src/ActionBarButton.vue
landondurnan: Certainly, but what were you thinking specifically? I sort of approached it as:

1. base colors (color / background color)
2. state adjustments constants
3. state colors (with their conditionals)

I needed to declare the state adjustments before the state colors so they could be used there. Would you mind pseudo-code what you're thinking? I'd be happy to update in both components.

laurenhu: Maybe just renaming the color constants to match their variable names? `const colorHover`, `const colorText` etc? Something scannable so it's more obvious at a glance which is a color generation constant and which are the colors that will be mapped to variables and returned.

Issue #350, opened 2022-06-06
Theme bug: profiles definitions in nested themes merged incorrectly


laurenhu:
### Bug description

When the MTheme `resolveTheme` function is merging its `theme` prop with parent themes, profiles are being merged in index order without regard to their `id`. This could also result in profiles with unexpected properties as well as `defaultProfile` being overwritten if a nested `<m-theme :theme="defaultTheme()">` is used.

In the following situation:

```
<m-theme :theme="{ profiles: [{ id: 'profile-one', colors: { background: 'red' } }] }>
	<m-theme :theme="{ profiles: [{ id: 'defaultProfile' }] }"></mtheme>
</m-theme>
```

this is the data on the inner MTheme:

```
profiles: [
    {
        id: 'defaultProfile',
        colors:  { background: 'red' },
    }
]
```

### Reproduction

Nest a Mtheme component with `theme.profiles` definitions inside another Mtheme component with different `theme.profiles` definitions.

### Environment

```shell
All
```

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

pretzelhammer:
assuming we're matching profiles by their ids, how should they be merged? should they be merged together recursively OR should the latter defined profile completely override the formerly defined profile?

pretzelhammer:
notes from our slack huddle:
- we don't need a `defaultProfile` profile in the default-theme `profiles` array
- we can merge profile definitions in the array by id (non-breaking)
- we can change the profiles array into a profiles object (breaking)
- no hard deadline

PR #360, opened 2022-06-10
fix(theme): merges nested profiles by id field

pretzelhammer:

closes https://github.com/square/maker/issues/350

updates merge strategy to catch keyed arrays

## Other information

nope

### DIFF
--- a/src/components/Theme/src/Theme.vue
+++ b/src/components/Theme/src/Theme.vue
@@ -8,8 +8,9 @@
 </template>
 
 <script>
-import { mergeWith, find } from 'lodash';
+import { merge, mergeWith, find } from 'lodash';
 import { BASE_TEN } from '@square/maker/utils/constants';
+import { showWarning } from '@square/maker/utils/debug';
 import key from './key';
 import defaultTheme from './default-theme';
 import { resolve, getPath } from './utils';
@@ -20,11 +21,46 @@ function isVueComponent(value) {
 	return value && value.render;
 }
 
+// returns true if value is an array
+// of objects with id fields
+function isKeyedArray(value) {
+	return value?.[0]?.id;
+}
+
+function keyedArrayIntoObject(arrayValue) {
+	const objectValue = {};
+	for (const item of arrayValue) {
+		objectValue[item.id] = item;
+	}
+	return objectValue;
+}
+
+function objectIntoKeyedArray(objectValue) {
+	const arrayValue = [];
+	for (const [id, item] of Object.entries(objectValue)) {
+		item.id = id;
+		arrayValue.push(item);
+	}
+	return arrayValue;
+}
+
+function mergeKeyedArrays(value, mergeValue) {
+	const valueObject = keyedArrayIntoObject(value);
+	const mergeValueObject = keyedArrayIntoObject(mergeValue);
+	const mergedObject = merge(valueObject, mergeValueObject);
+	const mergedKeyedArray = objectIntoKeyedArray(mergedObject);
+	return mergedKeyedArray;
+}
+
 // recursively merge all objects
 // EXCEPT for Vue components
-function mergeStrategy(_value, mergeValue) {
+function mergeStrategy(value, mergeValue) {
 	if (isVueComponent(mergeValue)) {
 		return mergeValue;
+	}
+
+	if (isKeyedArray(value) && isKeyedArray(mergeValue)) {
+		return mergeKeyedArrays(value, mergeValue);
 	}
 
 	// returning undefined means "merge values recursively"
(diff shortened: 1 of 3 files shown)


src/components/Theme/src/Theme.vue
```diff
@@ -20,21 +21,66 @@ function isVueComponent(value) {
...
 	// returning undefined means "merge values recursively"
 	// https://lodash.com/docs/4.17.15#mergeWith
 	return undefined;
 }
 
 function resolveTheme(data, parentTheme, theme, profileId) {
 	mergeWith(data, parentTheme, theme, mergeStrategy);
-	mergeWith(data, find(data.profiles, { id: profileId }), mergeStrategy);
+	if (profileId) {
+		const foundProfile = find(data.profiles, { id: profileId });
```
landondurnan, line 74: We can likely use the native `find` over the lodash one. https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_find

src/components/Theme/src/Theme.vue
```diff
@@ -20,21 +21,66 @@ function isVueComponent(value) {
...
+
+function keyedArrayIntoObject(arrayValue) {
+	const objectValue = {};
+	for (const item of arrayValue) {
+		objectValue[item.id] = item;
+	}
+	return objectValue;
+}
+
+function objectIntoKeyedArray(objectValue) {
```
landondurnan, line 38: There was an open question regarding why we chose to define profiles in an array. I can't recall that reason and we likely need to continue to support that to avoid breaking changes, but I could imagine all this could get much simpler without having to iterate through each.

Should we document how / why we support this data format? Should we consider changing it?

src/components/Theme/src/Theme.vue
```diff
@@ -20,21 +21,66 @@ function isVueComponent(value) {
...
+	const arrayValue = [];
+	for (const [id, item] of Object.entries(objectValue)) {
+		item.id = id;
+		arrayValue.push(item);
+	}
+	return arrayValue;
+}
+
+function mergeKeyedArrays(value, mergeValue) {
```
landondurnan, line 47: We should consider adding some tests, the data manipulation here is sound but pretty complex. Having some tests to ensure we're always getting the right data with these methods would be ideal.

landondurnan (commented):
It'd be ideal to demonstrate the bug fix with some sample code before / after. High-level all the code looks good, but it's challenging to know how to verify without sample theme inputs or tests.

src/components/Theme/src/Theme.vue
```diff
@@ -20,21 +21,66 @@ function isVueComponent(value) {
 	return value && value.render;
 }
 
+// returns true if value is an array
+// of objects with id fields
+function isKeyedArray(value) {
```
laurenhu, line 26: Should we just name these functions `isProfilesArray`, `profilesArrayIntoObject`, etc? Since this fixes an issue with the `theme.profiles` property and we don't have any other arrays in the theme data?

Unless we expect to add more arrays to the theme data. But if the idea is to eventually change `theme.profiles` to an object in the future, making it clear these functions are specifically to convert the profiles array might make cleanup easier down the line.

src/components/Theme/src/Theme.vue
laurenhu: I am guessing it was set up as an array to match Website. Had to dig pretty deep to find the reason profiles are saved as an array in Website instead of an object:
https://coda.io/d/ECOM-Website_dM7hmP2aSPk/Color-schema-updates_su4Up#_luUPn

```
Previous colorProfiles structure: Object instead of array
Faster, easier for lookup, but if we allow customer driven creation of new custom profiles in the future, we’ll need a way to track their order so they display consistently when displayed as options in the UI, and the object approach displayed below would need to be iterated on.
```
Screen Shot 2022-06-15 at 10 16 04 AM

Obviously, this is not a consideration for Maker themes at all so I think we're fine to convert `theme.profiles` to an object down the line.

laurenhu (approved):
This works well as a fix for the current issue with merging the profiles array.

Added some context on why Website uses array (and probably we mirrored that setting up the MTheme profiles). It seems the main reason is not applicable for MTheme so we could add a task to convert profiles to an object eventually. It would be faster and cleaner.

src/components/Theme/src/Theme.vue
pretzelhammer: i think this is the key reason, and really the only reason, why an array was chosen over an object:

> we’ll need a way to **track their order** so they **display consistently when displayed as options in the UI**

that's the only thing an array can do that an object can't, unless we generate ascii-betical ids, e.g. `--profile-01`, `--profile-02` and then sort by their ids to generate a consistent ordering.

src/components/Theme/src/Theme.vue
pretzelhammer: Had a sync with Bashu the other day and he showed me some theme data and they're using keyed arrays for the secondary colors also... e.g.,

```
secondaryColors: [
  { id: '--whatever', color: 'whatever' },
  { id: --something-else', color: 'something-else' },
]
```

I think we should leave the name and behavior as-is, since it solves the problem in general and isn't overfitted specifically for profiles.

src/components/Theme/src/Theme.vue
pretzelhammer: can't write tests, we don't even have a unit test lib yet, and i think adding that would be outside the scope of this PR, so i created an issue for it: https://github.com/square/maker/issues/363

src/components/Theme/src/Theme.vue
pretzelhammer: I added a comment above the keyedArray function to state that it's primarily used for `profiles`

src/components/Theme/src/Theme.vue
pretzelhammer: anyway, this discussion is outside the scope of this PR, since if we do migrate profiles from a keyedArray into an object that would be a breaking change and force refactors in website (and possible checkout) which are both fine things to do but something we can do in a follow-up PR

PR #366, opened 2022-06-15
fix(menu): fix auto focus on menu options, remove custom toggle style

laurenhu:

The combination of `MPopoverLayer` sitting at the end of a page and the `.focus()` event on `MMenuOption` auto-scrolling to bring the element into frame caused Menus on long pages to jump when opened. It also prevented the menu option from actually being auto-focused when the menu opens.

This also removes the `display: inline-block` style from the custom menu toggle.

BUG

This PR adds the `preventScroll: true` option to the `MMenuOption` focus events to prevent the page from scrolling when the menu content is mounted.

FIX

### DIFF
--- a/src/components/Menu/src/Menu.vue
+++ b/src/components/Menu/src/Menu.vue
@@ -24,7 +24,6 @@
 				</select-control>
 				<div
 					v-else
-					:class="$s.CustomToggle"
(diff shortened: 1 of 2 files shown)


landondurnan (approved):
Nice fix!

PR #375, opened 2022-06-29
fix(image): fix loading proper size image

kevinlee11:
Fixes images being loaded at full size instead of the corresponding sizes attribute passed in.

Add the sizes attribute when creating the image element for loading.

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -171,6 +171,12 @@ export default {
 				img.srcset = this.srcset;
 			}
 
+			// Needed to not load the full size image
+			// and match the size loaded in the UI
+			if (this.$attrs?.sizes) {
+				img.sizes = this.$attrs.sizes;
+			}
+
 			img.addEventListener('load', () => {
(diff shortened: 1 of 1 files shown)


sohnj877 (approved):
Thanks!

src/components/Image/src/Image.vue
```diff
@@ -171,6 +171,12 @@ export default {
 				img.srcset = this.srcset;
 			}
 
+			// Needed to not load the full size image
+			// and match the size loaded in the UI
+			if (this.$attrs?.sizes) {
```
pretzelhammer, line 176: i feel like we probably should just make `sizes` a prop if we're expecting it and using it inside the component

landondurnan:
this is already merged so it doesn't really matter, but this should have been considered a feature rather than a bug fix. While the feature is fixing a bug in the consuming app, Maker didn't have the feature before and this is an addition rather than a fix.

kevinlee11:
So Maker did have this feature before, it was just being passed implicitly through `v-bind="$attrs"`

PR #380, opened 2022-07-06
feat(dialog): add afterCloseHook and closeData to dialogs

### LINKED ISSUE #378: Provide A Method for a Dialog to Surface How It Was Closed + Other Info
### Feature request

Currently, there is no way for a dialog to provide any information to the component when it is being closed. 

Eg:

```typescript
// in the dialog/modal
export default {
  ...
  methods: {
    handleClose() {
      this.dialogApi.close(this.someInternalState);
    },
  },
};

// in the opening parent component
export default {
  ...
  methods: {
    this.dialogApi.open(SomeDialogComponent, {
      beforeCloseHook(closeData) {
        return closeData !== SOME_FLAG_TO_BLOCK_CLOSING;
      },
    } 
  }
};
```

### Why?

Currently, passing information from the dialog to its parent requires going through something like Vuex/Pinia, which causes a disconnect and can lead to hacky workarounds.

### Alternatives

_No response_

### Additional context

I am willing to write the needed code to get this working (I have a POC branch ready).

### LINKED ISSUE #379: Add an afterCloseHook to Dialogs
### Feature request

Similar to the current `beforeCloseHook` option, this would fire when the dialog/modal has been fully dismissed.

As an extension of #378, it would also accept the same `closeData`.

```typescript
// in the dialog/modal
export default {
  ...
  methods: {
    handleClose() {
      this.dialogApi.close(this.someInternalState);
    },
  },
};

// in the opening parent component
export default {
  methods: {
		openDialog() {
			this.dialogApi.open(SomeDialogComponent, {
				beforeCloseHook(closeData) {
					return closeData !== SOME_FLAG_TO_BLOCK_CLOSING;
				},
	
				afterCloseHook(closeData) {
					if (closeData === SOME_FLAG_A) {
						this.performActionA();
					} else if (closeData === SOME_FLAG_B) {
						this.performActionB();
					}
				},
			});
		}
	}
};
```

### Why?

It would be handy to have a callback function that does not block the actual closing of a dialog/modal. This would allow consumers to kick off async functions without causing the dialog to stay open.

### Alternatives

_No response_

### Additional context

I am willing to open a PR with this (I have a POC ready with the one for #378)

tatems:
Closes #378
Closes #379

This PR addresses the above issues to provide more avenues to interact with modals/dialogs when they are being closed.

- Adds another option to the `open` function in `dialogApi` called `afterCloseHook`
  - This function does not block the respective component from closing
  - Can be used in conjunction with `beforeCloseHook`
- Add an optional argument for the `close` function in `dialogApi` to allow the component to directly pass data when it closes itself
  - both `beforeCloseHook` and `afterCloseHook` will receive these same arguments

### DIFF
--- a/src/components/Modal/src/Modal.vue
+++ b/src/components/Modal/src/Modal.vue
@@ -94,10 +94,7 @@ export default {
 		beforeClose: {
 			immediate: true,
 			handler(hook) {
-				if (hook) {
-					// parentModal refers to the layer that this modal is actually in
-					this.modalApi.state.parentModal.state.options.beforeCloseHook = hook;
-				}
+				this.modalApi.state.options.beforeCloseHook = hook;
 			},
 		},
 	},
--- a/src/components/Modal/src/Modal.vue
+++ b/src/components/Modal/src/Modal.vue
@@ -94,7 +94,8 @@ export default {
 		beforeClose: {
 			immediate: true,
 			handler(hook) {
-				this.modalApi.state.options.beforeCloseHook = hook;
+				// parentModal refers to the layer that this modal is actually in
+				this.modalApi.state.parentModal.state.options.beforeCloseHook = hook;
 			},
 		},
 	},
--- a/src/components/Modal/src/Modal.vue
+++ b/src/components/Modal/src/Modal.vue
@@ -94,8 +94,10 @@ export default {
 		beforeClose: {
 			immediate: true,
 			handler(hook) {
-				// parentModal refers to the layer that this modal is actually in
-				this.modalApi.state.parentModal.state.options.beforeCloseHook = hook;
+				if (hook) {
+					// parentModal refers to the layer that this modal is actually in
+					this.modalApi.state.parentModal.state.options.beforeCloseHook = hook;
+				}
 			},
 		},
 	},
--- a/src/components/Button/src/Button.vue
+++ b/src/components/Button/src/Button.vue
@@ -353,6 +353,7 @@ export default {
 	&.loading {
 		/* don't inherit color in loading state on hover/active */
 		color: transparent !important;
+		opacity: 1;
 	}
 }
 
--- a/src/components/Dialog/README.md
+++ b/src/components/Dialog/README.md
@@ -349,7 +349,7 @@ _CloseDataAndHooksDemoDialog.vue_
 				v-model="selectedOption"
 				value="option-1"
 			>
-				Option 1 (blocks the dialog)
+				Option 1 (blocks the dialog from closing)
 			</m-radio>
 			<m-radio
 				v-model="selectedOption"
--- a/src/components/Dialog/README.md
+++ b/src/components/Dialog/README.md
@@ -236,7 +236,11 @@ export default {
 ```
 
 ### Configurable options
+
+#### `dialog.open`
+
 The `dialogApi.open()` function has a second optional object parameter that offers configurable options. Current available options are:
+
 
 ```ts
 {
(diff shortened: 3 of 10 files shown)


pretzelhammer:
this seems reasonable to me, my only request is that you add a little more to the docs examples, instead of `typescript` code blocks use `vue` code blocks (so they'll be transformed into interactive demos) and show some basic use-case for the closeData and the afterCloseHook.

tatems:
@pretzelhammer will do!

pretzelhammer:
the docs look good, thanks, but there seems to be a small bug in new modal docs demo, the modal is still close-able on this page even if option-1 is selected: https://square.github.io/maker/styleguide/dialog-modal-close-data/#/Modal

can you fix it? then i'm down to approve the pr

landondurnan:
Just tested locally and the demos and code updates look good. The only thing I noticed was the same thing @pretzelhammer had already mentioned with the modal not actually being blocked.

tatems:
@pretzelhammer @landondurnan should be fixed now! The prop would overwrite the open option even if it wasn't defined. Does it maybe make sense to deprecate that method of setting the hook?

src/components/Modal/src/Modal.vue
```diff
@@ -94,7 +94,10 @@ export default {
 		beforeClose: {
 			immediate: true,
 			handler(hook) {
-				this.modalApi.state.options.beforeCloseHook = hook;
+				if (hook) {
+					// parentModal refers to the layer that this modal is actually in
+					this.modalApi.state.parentModal.state.options.beforeCloseHook = hook;
```
landondurnan, line 99: Thinking about this some more -- `parentModal` isn't something that is always intended to exist. We dissuade modal stacking for a reason and this will change the original behavior of the beforeClose hook.

I think this needs to be rethought a bit, perhaps having your own custom watcher and setting the data some data with the original hook?

tatems: Yeah I'm not really a fan of this either, but I tried to do it so the existing functionality would not break, but here's essentially what was happening (which imo was not correct but worked anyway)

- Component opens modal
- Layer `0` creates the modal component and the next modal layer (`1`)
- Modal component attaches the `beforeCloseHook` option to layer `1`
- `modalApi.close` is called, it checks for hooks in layer `1` and continues if valid/no hook
- Layer `1` is torn down

But now, this is how it works:
- Component opens modal with before/after hooks
- Layer `0` creates the modal component, attaches the hooks, and then creates the next layer (`1`)
- The modal component also defines its own hook, and reaches in to set it on layer `0` (this is the `parentModal` stuff that this thread is on).
- `modalApi.close` is called, it now checks for hooks in layer `0`
- Layer `1` is torn down

I think the fundamental complication here is that the `modalApi` within the context of an open modal is actually attached to one layer up (hence the internal changes to reference the hooks from `this.state.options.[before/after]CloseHook` to `vm.currentLayer.state.options.[before/after]CloseHook`).

landondurnan: @tatems Really appreciate your write up and considerations around this. I was starting to suspect that there may have been some unintended behavior in the original implementation that "worked" but perhaps in a confusing way. However, I've always interpreted the current behavior as correct due to how we construct the injection of the ModalLayer and Modal having one API injected between both meant the modal is intentionally a child element and provides some level of control over stacking contexts between other modals or other components.

Perhaps there is some rethinking we need to do around all this, but it's probably better for us to internally structure that behavior consistently whether there is modal stacking or not. We made some quick changes to support stacking and somewhat reluctantly. But it does appear that this creates some challenges with the before / after behavior you're trying to achieve.

I'm not exactly sure what that change, if any, that should be. If it's not an immediate need could we dig deeper into some real-world scenarios and we can try to better account for them in both contexts (single modal + stacking modals) -- perhaps planning ahead as another example that we need to consider automated testing to ensure this behavior works consistently with the new hooks.

pretzelhammer: i think i almost got it... and then i lost it... but i know the modal stacking code is very confusing to begin with, so even code that works correctly and as-intended sometimes looks like a confusing hacky mess.

@tatems if you could please add your explanation as a code comment i would appreciate it and will be willing to approve this pr. it's okay if the code comment and is huge and takes up a bunch of lines, the code is confusing to begin with i don't want people to have to dig through old pr comments to have to understand it, would be much nicer to have the explanation and reasoning inline.

tatems: I think the most confusing part is that modals work just slightly differently enough from popovers and dialogs that it's hard to keep the behaviours straight.

I agree that we should maybe rethink this. And since it's not blocking our immediate project (we only need the dialog functionality), I can drop the modal changes from this and re-open as a separate PR/issue/discussion?

pretzelhammer (approved):
since you moved the modal stuff out and simplified the code i'm approving this

PR #384, opened 2022-07-07
fix(text-button): use button color for loading, and keep opacity at 1

kevinlee11:
Updates the loading component to use the same color as the text-button and fix the opacity to be 1.

Text button:
image

Loading state before fix:
image

Loading state after fix:
image


landondurnan (approved):
Nice catch.

We should make this change in the Button component as well. Just realized the designs don't change the opacity and use the current color there as well.

kevinlee11:
> Nice catch.
> 
> We should make this change in the Button component as well. Just realized the designs don't change the opacity and use the current color there as well.

I'll update that as well

Issue #356, opened 2022-06-07
Accordion


landondurnan:
### Figma link

https://www.figma.com/file/SHY7Z7PbcVssjIJVuJ3IwU/%F0%9F%8D%81-Maker?node-id=5400%3A7158

### Description

We'd like to add a standard accordion component to Maker with flexibility to support passing in our existing text component to control the visual output / styling.

- Accordions should be able to work with a single item or multiple items. 
- We need the ability to opt-in to different behavior for the accordion items to have one open at a time or multiple at a time
- We need slots for secondary information as part of the selectable item (or header)
- We need hover / active states to help communicate interaction and accessibility
- We should support all recommended ARIA labels and other HTML considerations: https://a11y-style-guide.com/style-guide/section-navigation.html
- We need support for an inline accordion with a different visual style, implementation as a separate component or variant on the accordion needs some additional discovery.

The current design still needs a few interaction state updates. Disabled is likely premature, but we do need hover / active / focus states -- likely similar to what we do with buttons.

### API

The Accordion component from Orbit may be a good starting point: https://github.com/square/orbit/blob/master/src/components/Accordion/doc/index.md

However, with the additional flexibility for the accordion text we will likely need a slot approach over passing a string in as a prop.

Internally we should consider leveraging the row component (currently a `util` with the idea we'd move it to a public component): https://github.com/square/maker/tree/master/src/utils/Row. (see usage in Menu: https://github.com/square/maker/blob/master/src/components/Menu/src/MenuOption.vue)

Consumer template API
```
<m-accordion>
  <m-accordion-item>
    <template slot="title">Item title</template>
    <template slot="secondary">Secondary information</template>
    <template slot="content">Lorem ipsum</template>
  </m-accordion-item>
  <m-accordion-item>
    <template slot="title">Item title</template>
    <template slot="secondary">Secondary information</template>
    <template slot="content">Lorem ipsum</template>
  </m-accordion-item>
</m-accordion>
```
props:
- v-model (optional) - control which element is open
- variant - TBD depending on "inline" option

slots:
- title
- secondary
- content

events:
- accordion item change open / close, pass the v-model

### Theme support

- The accordion component should be transparent, but inherit the text colors and fonts automatically from the theme. If using the Text component in the available slots, this should happen already. If just using normal text then we'll just use the body text in the theme.
- There may be some generated interaction colors, potentially leveraging existing neutrals or something similar to what we do with button states.
- The inline accordion should inherit the primary color

PR #385, opened 2022-07-08
feat(accordion): mvp accordion component

pretzelhammer:

closes https://github.com/square/maker/issues/356

implements accordion component

## Other information

https://square.github.io/maker/styleguide/mvp-accordion/#/Accordion

### DIFF
--- a/src/components/Accordion/src/Accordion.vue
+++ b/src/components/Accordion/src/Accordion.vue
@@ -11,6 +11,7 @@
 			@click="toggleExpanded"
 		>
 			<template #default>
+				<!-- @slot title of accordion -->
 				<slot name="title">
 					<m-text
 						pattern="title"
@@ -22,6 +23,7 @@
 				</slot>
 			</template>
 			<template #secondary>
+				<!-- @slot secondary info, goes under title -->
 				<slot name="secondary">
 					<m-text
 						pattern="paragraph"
@@ -32,6 +34,7 @@
 				</slot>
 			</template>
 			<template #suffix>
+				<!-- @slot open & close icon -->
 				<slot name="icon">
 					<m-icon :name="iconName" />
 				</slot>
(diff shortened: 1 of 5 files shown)


src/components/Accordion/src/Accordion.vue
```diff
@@ -0,0 +1,204 @@
...
+			type: String,
+			required: false,
+			default: '',
+		},
+	},
+
+	data() {
+		return {
+			// empty string values are truthy for accordion
+			internalExpanded: !!this.expanded || this.expanded === '',
```
laurenhu: See above re: the `|| this.expanded === ''` and prop type for `expanded` and `expandKey`.

laurenhu (approved):
Looks good. Great examples for the doc. Question over whether we want to use `Row` for the Accordion header to maintain consistency with spacing, but we can always swap in in later.

src/components/Accordion/src/Accordion.vue
```diff
@@ -0,0 +1,202 @@
+<template>
+	<div
```
landondurnan, line 2: Should consider all the recommendations from W3C on the markup, attributes and controls. https://www.w3.org/WAI/ARIA/apg/patterns/accordion/

src/components/Accordion/src/Accordion.vue
```diff
@@ -0,0 +1,204 @@
...
+		:class="$s.Accordion"
+		v-bind="$attrs"
+		v-on="$listeners"
+	>
+		<div
+			:class="$s.AccordionHeader"
+			@click="toggleExpanded"
+		>
+			<div
+				:class="$s.TitleIconAligner"
```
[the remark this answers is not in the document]
pretzelhammer: I tried to use the Row component initially but it didn't work for me because the icon/suffix slot is center-aligned between the title slot and the secondary slot, but for the Accordion designs the icon slot needs to be aligned with just the title slot and the secondary slot hangs below on its own.

Issue #362, opened 2022-06-14
Right-aligned toggles/checkboxes


tatems:
### Feature request

To support some of the design work on SO/ESDK, we'd like the ability to flip the layout of the checkbox and toggle components so that the input lives on the right side of the label instead of the left side.

### Why?

Currently there isn't much flexibility with the toggle and checkbox component layouts.

### Alternatives

![image](https://user-images.githubusercontent.com/3402466/173579476-c1f65f13-01b2-45b2-8e19-6ccb5d27e1b0.png)

tatems:
We're willing to open a PR with this as part of our internal project. cc @sohnj877

pretzelhammer:
I'm thinking Maker should just expose the underlying control components, e.g. RadioControl, CheckboxControl, etc so that users can use them in their own custom layouts. What do you think? @tatems

tatems:
Yup that would make sense to me!

pretzelhammer:
Discussed this with Design and they removed the right-aligned controls from the designs. I'm closing this for now but feel free to re-open if right-aligned form controls pops up again.

PR #396, opened 2022-07-21
feat: export standalone controls without form layout wrapper

pretzelhammer:

closes https://github.com/square/maker/issues/362

exports standalone controls without form layout wrapper

## Other information

nope


tatems (approved):

laurenhu:
I'm worried this approach will be confusing - at least when it's applied to all the components. For instance, it's not clear at a glance from the styleguide which Input component to use, especially since they're now referenced as InputControl vs. InputFormItem in the docs, with the latter coming first. Appearance-wise, there's little difference between the two components, besides the error slot. I could see this making sense for radio, checkbox and toggle but not as much for Input, Select and PinInput. I would limit standalone controls to just the former three. Alternatively, we could also introduce an `align` prop to them, which we already have on Input to style the alignment of text. This would also allow us to standardize spacing for both configurations.

laurenhu (commented):
I'm worried this approach will be confusing - at least when it's applied to all the components. For instance, it's not clear at a glance from the styleguide which Input component to use, especially since they're now referenced as InputControl vs. InputFormItem in the docs, with the latter coming first. Appearance-wise, there's little difference between the two components, besides the error slot. I could see this making sense for radio, checkbox and toggle but not as much for Input, Select and PinInput. I would limit standalone controls to just the former three. Alternatively, we could also introduce an align prop to them, which we already have on Input to style the alignment of text. This would also allow us to standardize spacing for both configurations.

mattmewton:
Agree with @laurenhu 's comments, I think makes sense to keep the API as clean and simple as possible for consumers of Maker and the approach outlined in the above comment seems to align with that.

pretzelhammer:
closing this PR because the associated issue has been closed, but if it's ever re-opened we can also re-open this pr

PR #403, opened 2022-08-04
feat: mvp badge component

pretzelhammer:

no badge component

adds badge component

## Other information

Badge styleguide link: https://square.github.io/maker/styleguide/mvp-badge/#/Badge


src/components/Badge/src/Badge.vue
```diff
@@ -0,0 +1,209 @@
...
+			if (this.pseudoTarget === 'before') {
+				classes.push(this.$s.BeforeBadge);
+			} else {
+				classes.push(this.$s.AfterBadge);
+			}
+			return classes;
+		},
+		styles() {
+			const bgColor = this.resolvedBgColor || '#000';
+			const textColor = this.resolvedTextColor || getContrast(bgColor);
```
laurenhu, line 129: Should we test text color against the bg the way we do for button?

laurenhu (approved):
This looks great and you've really considered different use cases with the `noRelative` and `pseudoTarget` props. Only feedback is that we should consider checking contrast on the textColor the way we do for button - though it looks like we don't do it for Pill.

PR #409, opened 2022-08-23
feat(image): add image:loaded event

laurenhu:

Currently, there is no way of knowing when the `` element in an `MImage` component is not only loaded but has completed it's fade-in transition.

This PR adds a `image:loaded` event that is emitted when the image src is loaded and the `@after-enter` hook on the image transition is called.

### DIFF
--- a/src/components/Image/README.md
+++ b/src/components/Image/README.md
@@ -8,7 +8,7 @@
 			class="image"
 			src="https://source.unsplash.com/random/400x600"
 			:shape="shape"
-			@image:loaded="imageLoaded"
+			@image:visible="imageVisible"
 		/>
 		<m-image
 			class="image image-tall"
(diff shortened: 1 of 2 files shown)


src/components/Image/README.md
```diff
@@ -100,4 +110,8 @@ Supports attributes from [`<img>`](https://developer.mozilla.org/en-US/docs/Web/
 ## Events
 
 Supports events from [`<img>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img).
+
+| Event        | Type | Description                                        |
+| ------------ | ---- | -------------------------------------------------- |
+| image:loaded | -    | Image is loaded and fade in transition is complete |
```
bashunaimiroy, line 116: maybe it would be more accurate to call it `image:ready` or `image:visible`? But it's not a blocker issue imo

pretzelhammer (approved):
+1 on Bashu's comment, i think it would be more semantic to call it `image:visible` or something

PR #418, opened 2022-09-07
fix: makes modal, dialog, & blade responsive to theme bg & body colors

pretzelhammer:

as written in pr title

makes color & bgColor of blades themeable (similar to how it is for modals & dialogs), also updates the default theme to set the color & bgColor values to `@colors.body` and `@colors.background` for all components

## Other information

nope

### DIFF
--- a/src/components/Blade/src/Blade.vue
+++ b/src/components/Blade/src/Blade.vue
@@ -1,5 +1,10 @@
 <template>
-	<div :class="$s.BladeContainer">
+	<div
+		:class="$s.BladeContainer"
+		:style="style"
+		v-bind="$attrs"
+		v-on="$listeners"
+	>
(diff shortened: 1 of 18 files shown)


src/components/Blade/src/Blade.vue
```diff
@@ -1,24 +1,78 @@
...
+</script>
+
 <style module="$s">
 .BladeContainer {
 	position: relative;
 	width: 100%;
 	height: 100%;
 	overflow: hidden;
+	color: var(--color);
+	background-color: var(--bg-color);
```
laurenhu, line 70: Can we include fallback values for these like we do with Modal and Dialog?

laurenhu (approved):
LGTM! Minor comment for fallback css values.

PR #428, opened 2022-09-14
ci: fail loudly when deploys push fails

pretzelhammer:
if pushing a styleguide or lab deploy failed during CI it would fail silently which makes it hard to find in the github UI, this updates the script to fail loudly when a push fails

### DIFF
--- a/build/deploys/push.js
+++ b/build/deploys/push.js
@@ -1,6 +1,10 @@
+/* eslint-disable no-console */
+
 const { promisify } = require('util');
 const exec = promisify(require('child_process').exec);
 const ensureDeployDirectory = require('./ensure');
+
+const EXIT_ERROR_CODE = 1;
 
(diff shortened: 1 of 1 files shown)


build/deploys/push.js
```diff
@@ -18,10 +22,15 @@ const ensureDeployDirectory = require('./ensure');
...
-	await exec('git push');
-	// the above command suceeds regardless of
-	// whether anything was actually pushed or not
+	try {
+		await exec('git push');
+	} catch {
+		console.error('failed to push changes to remote');
+		process.exit(EXIT_ERROR_CODE);
+	}
 }());
```
privatenumber, line 36: I would recommend adding this to handle all errors:

```suggestion
}()).catch((error) => {
	console.error(error.message);
	process.exit(1);
});
```

privatenumber (approved):

PR #430, opened 2022-09-16
feat(choice): choice card

laurenhu:

We currently don't have an way to display Choice options as Cards

Adds ability to display Choice options as Cards.
Screen Shot 2022-09-16 at 11 23 46 AM

## Other information

https://square.github.io/maker/styleguide/choice-card/#/Choice


src/components/Choice/src/ChoiceCard.vue
```diff
@@ -0,0 +1,83 @@
...
+			$s.ChoiceCard,
+			{
+				[$s.selected]: isSelected,
+				[$s.disabled]: isDisabled,
+			},
+		]"
+		:disabled="isDisabled"
+		@click="selectSelf"
+	>
+		<slot />
```
tatems, line 13: Could we expose whether or not a card is selected to the slot? We use this pattern in checkout and it would make it easier to show/collapse the card content based on selection.

pretzelhammer (approved):

PR #433, opened 2022-09-21
feat: mvp Toast component

pretzelhammer:

no Toast component

adds Toast component

## Other information

- [TransitionStack util docs](https://square.github.io/maker/styleguide/toast-final/#/TransitionStack)
- [Toast, ToastLayer, & Bread component docs](https://square.github.io/maker/styleguide/toast-final/#/Toast)


src/components/Toast/src/Toast.vue
```diff
@@ -0,0 +1,316 @@
...
+			},
+		]"
+		:persistent="resolvedPersistent"
+		:dismiss-after="resolvedDismissAfter"
+		:style="style"
+		v-bind="$attrs"
+		v-on="$listeners"
+	>
+		<div :class="$s.ToastContent">
+			<m-icon
```
laurenhu, line 16: Can we make icon a prop as well as a slot? Similar to what we did with the Accordion header. So it's not necessary to add an icon to a theme if it's a toast has a one-off icon?

src/components/Toast/src/Toast.vue
```diff
@@ -0,0 +1,316 @@
...
+				v-if="!resolvedHideIcon"
```
laurenhu, line 17: Is `hideIcon` as a separate prop necessary? Could we conditionally render it or hide if depending on if an `icon` prop or slot content is present?

src/components/Toast/README.md
```diff
@@ -0,0 +1,663 @@
+# Toast
+
+Use MToast to notify users of things.
+
+```vue
+<template>
+	<div>
+		<button @click="openToast">
```
laurenhu, line 8: Can we add all the possible content to the default scenario? Including icon, and actions? So folks can copy/paste with all the options and just remove what they don't need.

laurenhu (commented):
This is fantastic! 🎉 🍞 
A few minor notes and questions and I think there are a few empty files in the lab directory.

src/components/Toast/src/Toast.vue
pretzelhammer: it's necessary. making `undefined`, `null`, or `'' (empty string)` valid values for `iconName` is not good for the following reasons:
- `undefined` is indistinguishable from actually NOT setting the value, and `resolveThemeableProps` will look further up the theming hierarchy to find a valid value
- `null`, while distinguishable from not setting the value, is currently caught by `resolveThemeableProps` similarly to `undefined` and changing this would be a breaking change, so i'd prefer not to make this change, furthermore, I don't like allowing `null` as a valid value for `iconName` since semantically `null` is not a valid icon name, also the `iconName` is used across several other components and not all of these components have optional icons, some do and some don't, so it'd be weird if some components supported `null` values to hide the icons and others don't
- `'' (empty string)` is not good for the same reason as `null`: it's not semantic and it wouldn't be consistent with other components which use the `iconName` prop

`hideIcon` is clear, semantic, keeps behavior & valid values consistent between components which have mandatory icon slots vs optional icon slots. furthermore, it allows us to both set a default icon and a default visibility status for the default icon in each pattern, so altho an icon might be hidden by default, if the user chooses they would like to show the icon then the correct icon will be shown, which is not possible if we allow null-ish/falsey values for `iconName`, e.g.:

this is better:
```js
{
  "iconName": "info",
  "hideIcon": false, // icon might be hidden by default BUT if they set this to true they will get the correct icon for the context/pattern
}
```

than this:
```js
{
  "iconName": "", // okay, icon is hidden but... if I want to show it what should I change this to!?!?
}
```

src/components/Toast/src/Toast.vue
pretzelhammer: hmmm, i kinda would prefer people to just use the `iconName` slot buuut since we have them as slots in Accordion and Notice i'll add it into Toast for consistency's sake

src/components/Toast/README.md
pretzelhammer: I think the first example in the docs of any component should be as simple and basic as possible to make it easy for devs new to the component to be able to grok it quickly. Then the examples get progressively more complicated as they scroll down.

I added 3 more examples to the docs. 2 examples specific to icons and 1 example combining several features show off in the previous one-off examples. I hope this is satisfactory.

src/components/Toast/src/Toast.vue
pretzelhammer: Actually, since the default value for `hideIcon` is `true` that means whenever they want to show icons they have to pass `<m-toast :hide-icon="false" />` where if we change the prop name to `shownIcon` and set the default value to `false` if they want to show the icon they only have to do `<m-toast show-icon />` so i've refactored the code and changed `hideIcon` to `showIcon`.

laurenhu (approved):

PR #434, opened 2022-09-21
fix(image): fix image:visible event for cached images

laurenhu:

The `image:visible` event was only firing for new images and not cached images

Emits the `image:visible` event for both new and cached images.


pretzelhammer (approved):

bashunaimiroy (approved):
this would indeed work in the context of ecom-website

PR #443, opened 2022-09-28
ci: increase deploy styleguide & lab job success rate

pretzelhammer:
so the deploy styleguide & lab job already had a very high success rate, the only times it would fail is if multiple instances of the job were run in parallel, in which case the first job to finish would complete successfully but all of the following jobs would fail because they would attempt to push their changes to remote but remote had changes that they didn't have locally so they would fail to push their changes and thus the job as a whole would fail

adds a `git pull --rebase` right before the `git push` to increase the success rate of the job, it'll still sometimes fail, but in general should fail much much less often as the pushes of independent jobs rarely conflict with each other and can be auto-merged

also, while poking around, i also made some CSS improvements to the Maker Deploys page, including styles for people who prefer dark mode (like myself)

## Other information
nope

### DIFF
--- a/.github/workflows/deploy-styleguide-lab.yml
+++ b/.github/workflows/deploy-styleguide-lab.yml
@@ -11,6 +11,7 @@ jobs:
   deploy-styleguide-lab:
     name: Deploy Styleguide & Lab
     runs-on: ubuntu-latest
+    concurrency: changes-deploys
 
     steps:
       - name: Checkout Maker master branch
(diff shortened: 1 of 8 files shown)


privatenumber:
> the only times it would fail is if multiple instances of the job were run in parallel

You might be able to use the `concurrency` property so that one one deploy can run at a time:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

privatenumber (approved):
🤞

PR #445, opened 2022-09-28
feat(button): add props for border, box-shadow and hover styles

laurenhu:
Buttons currently do not support custom border, box-shadow, shape and hover styles

Adds support to buttons for custom border, box-shadow, shape and hover styles

Sep-28-2022 12-18-30


src/components/Button/src/Button.vue
```diff
@@ -130,37 +144,110 @@ export default {
...
+			validator: (shape) => ['squared', 'rounded', 'pill'].includes(shape) || shape.match(/\d*\.?\d+(?:px|%)/i),
+		},
+		/**
+		 * Shape of button
+		 * @values 'Npx', 'N%'
+		 */
+		shapeHover: {
+			type: String,
+			default: undefined,
+			validator: (shape) => shape.match(/\d*\.?\d+(?:px|%)/i),
```
pretzelhammer, line 203: since this regex is used in a couple places can we refactor it out and declare it as a constant above? e.g.,
```js
// matches '{number}px' & '{number}%' strings
const PX_PERCENT_REGEX = /\d*\.?\d+(?:px|%)/i;
```

src/components/Button/src/Button.vue
```diff
@@ -130,37 +144,110 @@ export default {
...
 		},
 		/**
 		 * Shape of button
+		 * @values squared, rounded, pill, 'Npx', 'N%'
 		 */
 		shape: {
 			type: String,
 			default: undefined,
-			validator: (shape) => ['squared', 'rounded', 'pill'].includes(shape),
+			validator: (shape) => ['squared', 'rounded', 'pill'].includes(shape) || shape.match(/\d*\.?\d+(?:px|%)/i),
```
pretzelhammer, line 194: hmmm, did you consider making a `borderRadius` prop? e.g.
```js
borderRadius: {
  type: String,
  default: undefined,
  validator: cssValidator('border-radius'),
}
```
and if a `borderRadius` is passed we can ignore the `shape` prop. It would allow much greater customization, the current `shape` prop and its validator wouldn't allow people to customize the border-radius of the individual button corners independently, so things like "wavy" buttons wouldn't be possible, e.g. https://codepen.io/kfedorov91/pen/gOzemvN

I'm not sure if we want that level of customizability buuuut if we do it might just be worth it to add a `borderRadius` prop if it's possible and makes sense.

src/components/Button/src/Button.vue
```diff
@@ -130,37 +144,110 @@ export default {
...
+			type: String,
+			default: undefined,
+			validator: (shape) => shape.match(/\d*\.?\d+(?:px|%)/i),
+		},
+		/**
+		 * Border width of button (e.g. '3px')
+		 */
+		borderWidth: {
+			type: String,
+			default: undefined,
```
pretzelhammer, line 210: since this is a css property we can use `cssValidator('border-width')` here

src/components/Button/src/Button.vue
```diff
@@ -130,37 +144,110 @@ export default {
...
+		},
+		/**
+		 * Border hover width of button (e.g. '3px')
+		 */
+		borderWidthHover: {
+			type: String,
+			default: undefined,
```
pretzelhammer, line 217: `cssValidator('border-width')`

src/components/Button/src/Button.vue
```diff
@@ -130,37 +144,110 @@ export default {
...
+			validator: (color) => colord(color).isValid(),
+		},
+		/**
+		 * Box-shadow of button (e.g. 0px 0px 1px 3px #000000 or 0px 0px 1px 3px rgba(0, 0, 0, 0.5))
+		 */
+		boxShadow: {
+			type: String,
+			default: undefined,
+			validator: (boxShadow) => boxShadow.match(/(\d*\.?\d+px\s+){4}(#[\da-z]{6}|rgba\((\s*\d+\s*,){3}\s*[\d.]+\))/i),
```
pretzelhammer, line 242: `cssValidator('box-shadow')` or if the regex was explicitly chosen because it limits some undesirable values that `cssValidator('box-shadow')` would otherwise allow then we should at least refactor this regex into a constant above, e.g.
```
// validates CSS box-shadow values
const BOX_SHADOW_REGEX = /(\d*\.?\d+px\s+){4}(#[\da-z]{6}|rgba\((\s*\d+\s*,){3}\s*[\d.]+\))/i;
```

src/components/Button/src/Button.vue
```diff
@@ -130,37 +144,110 @@ export default {
...
+		},
+		/**
+		 * Box-shadow hover of button
+		 */
+		boxShadowHover: {
+			type: String,
+			default: undefined,
+			validator: (boxShadow) => boxShadow.match(/(\d*\.?\d+px\s+){4}(#[\da-z]{6}|rgba\((\s*\d+\s*,){3}\s*[\d.]+\))/i),
```
pretzelhammer, line 250: we should use `cssValidator('box-shadow')` or refactor into a const here

src/components/Button/src/Button.vue
```diff
@@ -241,17 +349,19 @@ export default {
...
+		var(--box-shadow, 0 0),
 		var(--focus-border, 0 0);
 	cursor: pointer;
 	transition:
 		color 0.2s ease-in,
 		background-color 0.2s ease-in,
 		filter 0.2s ease-in,
-		box-shadow 0.2s ease-in;
+		box-shadow 0.2s ease-in,
+		border-radius 0.2s ease-in;
```
pretzelhammer, line 364: since `border-width` and `border-color` can also be customized should we add those here as well?

pretzelhammer:
the [Button Customization](https://square.github.io/maker/lab/button-customization/#/ButtonCustomization) lab is great! thanks for making it. i played around with it and was impressed how by both the fill variant and the outline variant maintained their character and their accessibility despite my most devious efforts to destroy both. I'm guessing you didn't include the ghost variant because most of the customizations would not apply to it, but it'd still be interesting to see it in context if you don't mind adding it.

this pr was a little difficult to review because it has a lot of extraneous diffs but i did my best :) i look forward to giving it a follow-up review once my initial comments have been resolved and once `master` has been merged in & merge conflicts are resolved.

lab/experiments/ButtonCustomization.vue
```diff
@@ -0,0 +1,597 @@
...
+	</m-theme>
+</template>
+
+<script>
+import { MButton } from '@square/maker/components/Button';
+import { MTheme } from '@square/maker/components/Theme';
+import makerColors from '@square/maker/utils/maker-colors';
+
+const defaultButtonProps = {
+	textPattern: 'paragraph',
```
pretzelhammer, line 452: the default text pattern for interactive clickable things should be `label`

src/components/Button/src/Button.vue
```diff
@@ -120,7 +141,7 @@ export default {
 		size: {
 			type: String,
 			default: undefined,
-			validator: (size) => ['small', 'medium', 'large'].includes(size),
+			validator: (size) => ['small', 'medium', 'large', 'expandable'].includes(size),
```
pretzelhammer, line 144: can you please document in the Button's styleguide how the `expandable` size is different from the other sizes and when/why it would be used?

src/components/Button/src/Button.vue
```diff
@@ -272,8 +411,13 @@ export default {
 		min-width: max-content;
 	}
 
+	&.size_expandable {
+		min-height: 48px;
+		padding: 5px 20px;
```
pretzelhammer, line 416: i think all of our margins and paddings (and explicit heights and explicit widths) are multiples of 8 to maintain a consistent visual rhythm vertically & horizontally, with the only exception being 4px but even that one is kinda okay because it's half of 8. is it possible to make the `padding` here `4px 16px` or `8px 24px`? a `padding` of `5px 20px` is kinda odd.

src/components/Button/src/Button.vue
```diff
@@ -189,24 +290,60 @@ export default {
...
+			case 'medium':
+				return '14px';
+			case 'large':
+				return '16px';
+			default:
+				return undefined;
+			}
+		},
+		fontWeight() {
+			return this.resolvedSize === 'expandable' ? undefined : '500';
```
pretzelhammer, line 345: i don't understand the logic behind this... why would whether or not a button is `expandable` change it's font-weight? 🤔

src/components/Button/src/Button.vue
```diff
@@ -272,8 +411,13 @@ export default {
...
 	}
 
+	&.size_expandable {
+		min-height: 48px;
+		padding: 5px 20px;
+	}
+
 	&.size_small {
-		height: 32px;
+		height: 36px;
```
pretzelhammer, line 420: this is a weird choice, like i mentioned in another comment pretty much all of our margins, paddings, and explicit widths & heights are multiples of 8, and prior to this change the button heights were 32px (8*4) for small, 48px (8*6) for medium, and 64px (8*8) for large. Changing the small size from 32px to 36px breaks this convention so i think it warrants some justification, and i think we should prefer keeping the small size 32px if possible.

src/components/Button/src/Button.vue
```diff
@@ -130,21 +151,44 @@ export default {
 			default: undefined,
 		},
 		/**
-		 * Background color of button
+		 * MText pattern in button label. Pattern's size only applies on (`size: undefined`) buttons.
+		 */
+		textPattern: {
+			type: String,
+			default: undefined,
```
pretzelhammer, line 158: the default for `textPattern` within MButton is `undefined`, and when it is `undefined` it's the same as deferring the default value to the MText component, in which case it will always resolve to `paragraph`, which functionally means the default for MButton also becomes `paragraph`, however if we check MButton's CSS styles we'll see that the default text pattern for buttons, and for all interactive clickable components, is suppose to be `label`, so to maintain that behavior we need to switch this `undefined` to `'label'`. And if `textPattern` is eventually made into a themeable prop we'd have to add `textPattern: 'label'` to MButton's default theme.

src/components/Button/src/Button.vue
pretzelhammer: Okay, after looking through the CSS i see that `expandable` is basically the same as `medium` except that `medium` sets `height: 48px` where `expandable` sets `min-height: 48px` hence the height becomes expandable.

Rather than adding an `expandable` value, what do you think about adding `min-small`, `min-medium`, and `min-large` values? where those styles are the same as their `small`, `medium`, and `large` counterparts except that they set `min-height` instead of `height`?

pretzelhammer:
I looked through the [most recent commit](https://github.com/square/maker/pull/445/commits/d00774a7f2408afb8a2d451ae4fc99180e3a68d5) and to be honest I don't think Shimaa would have been a fan of `expandable` or `textPattern` props on buttons. Buttons having consistent sizes is important for site layout and site visual cohesiveness/consistency. By adding `expandable` and `textPattern` buttons can be made all kinds of weird and odd heights now. Furthermore, I think it was intended for Buttons to always have the `label` style with the label font family and label font weight... being able to change it to the `paragraph` style or even the `title` or `headline` styles is pretty weird, especially since the button's padding is not responsive to the text changes so any buttons using the `title` or `headline` styles look bad/cramped/claustrophobic.

At the end of the day if Website needs it then we'll ship it, but I'd just like to make sure that everyone involved is aware of the negative design implications of these changes.

src/components/Button/src/Button.vue
laurenhu: We don't actually set `border-color` or `border-width`, they're actually styled as box-shadows too.

PR #454, opened 2022-10-07
fix(modal): closing bugs

pretzelhammer:

The bugs:

❌ (1) a beforeClose hook set via options would never run

One way to set a beforeClose hook was via the options object in the 2nd param of `modalApi.open`, e.g.

```js
this.modalApi.open(
  // 1st param, render function
  () => <MyModal />,
  // 2nd param, modalApi options object
  {
    beforeCloseHook: () => console.log('I will never run 🤦'),
  },
);
```

This is bad obviously because the whole purpose of the beforeClose hook is to potentially block the modal from closing if the user forgot or failed to do something, but it would close regardless potentially leading to bad user experience or data in a corrupt or incomplete state. The beforeClose hook technically could also be used to perform final data clean up steps after a user was done interacting with a modal, and it failing to run could also lead to corrupt or incomplete data state.

Nobody noticed it was broken because... nobody was using it, which begs the question: why was it implemented in the first place 🤷? It was likely used for a time, then dropped or refactored, and then broken in later commit.

❌ (2) setting a beforeClose hook via the Modal's `beforeClose` prop would overwrite the beforeClose hook set via options

This is a very subtle bug that no one would have ever likely noticed because of bug (1) but even if that bug was fixed in an earlier or later version of Maker then this bug would pop up, e.g.

```js
this.modalApi.open(
  // hook set via Modal prop
  () => <MyModal
    beforeClose={() => console.log('this hook overwrites the one below')}
  />,
  // hook set via modalApi options
  {
    beforeCloseHook: () => console.log("I'll be overwritten by propHook"),
    // this hook will never run 🤦
  },
);
```

❌ (3) calling the close function returned by `modalApi.open` would force close an entire stack of modals _from the bottom_

Calling `modalApi.close` is suppose to only close the current modal, and if there is another modal stacked on top of the current modal then closing is blocked. This was correctly enforced for `modalApi.close` but not correctly enforced for the close function returned from `modalApi.open`, e.g.

```js
const closeFn = this.modalApi.open(() => <MyModal />);
closeFn(); // this could possibly force close multiple modals
```

Note: force closing a stack of modals _from the bottom_ is not good because: the bottommost modal usually isn't aware of the state of the modals above it, and if the user is performing some action in the upper modals like filling out a form then randomly closing it on them is horrible user experience. This is why we encourage and attempt to enforce closing modals _from the top only_ since the topmost modals have the most context and are the ones which are guiding the user experience.

❌ (4) calling the close function returned by `modalApi.open` would not run any beforeClose hooks on the closed modal

```js
const closeFn = this.modalApi.open(
  () => <MyModal beforeClose={hookOne} />,
  {
    beforeCloseHook: hookTwo,
  },
);
closeFn(); // this would skip running hookOne & hookTwo 🤦
```

❌ (5) calling `modalApi.state.parentModal.close` no longer works, but even when it "worked" it was broken in the same way as described by bug (4)

Originally `modalApi.state.parentModal.close` was implemented to close a stack of modals from the top. Since then it broke during some commit and stopped working, however even when it did "work" it was still broken in the same way as described by bug (4), i.e. none of the beforeClose hook would run.

❌ (6) calling `modalApi.state.parentModal.close` would only close a stack of exactly 2 modals, but would break for any other number, e.g. 1 or 3+

I think it's okay to want to close an entire stack of modals _from the top_. However, the way to do it should be reliable and independent regardless of how many modals there are in the stack, e.g. 1, 2, 3, 4 and so on. The original implementation of this feature only enabled users to close a stack of exactly 2 modals, and for any other number it was useless.

The fixes:

✅ (1) beforeClose hook set via options now runs

```js
this.modalApi.open(
  () => <MyModal />,
  {
    beforeCloseHook: () => console.log('Yay, I run now! 🙌'),
  },
);
```

✅ (2) beforeClose hook set via Modal's `beforeClose` prop no longer overwrites the beforeClose hook set via options, both are run

```js
this.modalApi.open(
  // hook set via Modal prop
  () => <MyModal
    beforeClose={() => console.log('I will run first! 🙌')}
  />,
  // hook set via modalApi options
  {
    beforeCloseHook: () => console.log('I will run second! 🙌'),
    // this hook will never run 🤦
  },
);
```

Both hooks have the opportunity to block closing. The hook set via props is run first and the hook set via options is run second. The hooks are run lazily, i.e. if the first hook blocks closing then the second hook isn't run.

✅  (3) calling the close function returned by `modalApi.open` will at most close the single specific opened modal, but will not close if there is a stack of multiple modals

It's not possible to accidentally or unintentionally close a stack of modals from the bottom anymore, as the close function returned from `modalApi.open` will now only close that specific single opened modal, and if there are modals stacked on top of it then closing will fail and the function will return `false`, e.g.

```js
const closeFn = this.modalApi.open(() => <MyModal />);
const result = closeFn();
// if result is true then MyModal was closed,
// if result is false then MyModal was not closed,
// which may mean there are other modals stacked
// on top of it which must be closed first
```

✅  (4) calling the close function returned by `modalApi.open` will run all beforeClose hooks registered on the opened modal

```js
const closeFn = this.modalApi.open(
  () => <MyModal beforeClose={hookOne} />,
  {
    beforeCloseHook: hookTwo,
  },
);
// this will now run hookOne & hookTwo before closing 🙌 
const result = await closeFn();
// if result is true then MyModal was closed,
// if result is false then MyModal was not closed,
// which may mean at least one of the beforeClose
// hooks blocked closing, i.e. returned false
// NOTE: if all of the hooks are sync then result will be a
// boolean, if any are async then result will be a Promise<boolean>
```

✅  (5) you can now call `modalApi.closeAll` to close an entire stack of modals from the top, and it will correctly run all of the beforeClose hooks on each modal

```js
// inside some part of the app
this.modalApi.open(
  () => <MyModal beforeClose={hookOne} />,
  {
    beforeCloseHook: hookTwo,
  },
);
// inside MyModal
this.modalApi.open(
  () => <MyNestedModal beforeClose={hookThree} />,
  {
    beforeCloseHook: hookFour,
  },
);
// inside MyNestedModal
const result = await this.modalApi.closeAll();
// hookThree & hookFour are run, if both return true MyNestedModal is closed
// hookOne & hookTwo are run, if both return true MyModal is closed
// if any of the hooks return false closing is stopped at that modal
// if all of the hooks are sync then result will be a boolean, if any of
// the hooks are async then result will be a Promise<boolean>
```

✅  (6) calling `modalApi.closeAll` works for closing a stack of modals of any size

`modalApi.closeAll` works for closing a single modal, or a couple modals, or 3, or 4, or so on.

## Other information
- beforeClose hooks can be sync or async
- any code using `modalApi.state.parentModal.close` should be refactored to use `modalApi.closeAll` after this PR is merged & released
- I renamed the internal modalApi state variable `currentLayer` to `parentModalApi` because I felt that name is more descriptive
- I added a bunch of comments to ModalLayer where the bulk of the confusing nested layer code exists
- I updated & improved the [Modal styleguide docs](https://square.github.io/maker/styleguide/close-all-modals/#/Modal)
- I created a [modal stacking & closing lab](https://square.github.io/maker/lab/close-all-modals/#/CloseAllModals) to debug and test all of my code works correctly

### DIFF
--- a/lab/App.vue
+++ b/lab/App.vue
@@ -19,6 +19,6 @@ export default {
 <style>
 body {
 	margin: 0;
-	background-color: #f8f8fb;
+	font-family: -apple-system, 'Helvetica Neue', sans-serif;
 }
 </style>
--- a/lab/components/MultiModal.vue
+++ b/lab/components/MultiModal.vue
@@ -80,17 +80,18 @@
 			async beforeClose hooks: mixed
 			<br>
 			<button @click="openNested({ propSync: false, propReturn: true, optSync: false, optReturn: false })">
-				async prop: true, sync option: false
+				async prop: true, async option: false
 			</button>
 			<br>
 			<button @click="openNested({ propSync: false, propReturn: false, optSync: false, optReturn: true })">
-				async prop: false, sync option: true
+				async prop: false, async option: true
 			</button>
 			<br>
 			mixed a/sync beforeClose hooks: mixed
 			<br>
-			<!-- todo implement this part -->
-
+			<button @click="openNested({ propSync: false, propReturn: true, optSync: false, optReturn: false })">
+				async prop: true, sync option: false
+			</button>
 			<br>
 			closing via popover
 			<br>
--- a/lab/components/MultiModal.vue
+++ b/lab/components/MultiModal.vue
@@ -89,7 +89,7 @@
 			<br>
 			mixed a/sync beforeClose hooks: mixed
 			<br>
-			<button @click="openNested({ propSync: false, propReturn: true, optSync: false, optReturn: false })">
+			<button @click="openNested({ propSync: false, propReturn: true, optSync: true, optReturn: false })">
 				async prop: true, sync option: false
 			</button>
 			<br>
--- a/lab/components/MultiModal.vue
+++ b/lab/components/MultiModal.vue
@@ -1,82 +1,232 @@
 <template>
-	<m-modal>
+	<!-- eslint-disable max-len -->
+	<m-modal :before-close="beforeClose">
 		<m-modal-content>
+			depth: {{ depth }}
+			<br>
+			no beforeClose hooks
+			<br>
 			<button @click="openNested">
-				open nested modal with no beforeClose hook
+				no beforeClose hooks
 			</button>
-			<button @click="openNestedHook({sync: true, returns: true})">
-				open nested modal with sync beforeClose hook that returns true
+			<br>
+			sync beforeClose hooks: true
+			<br>
+			<button @click="openNested({ propSync: true, propReturn: true })">
+				sync prop: true
 			</button>
-			<button @click="openNestedHook({sync: true, returns: false})">
-				open nested modal with sync beforeClose hook that returns false
+			<br>
+			<button @click="openNested({ optSync: true, optReturn: true })">
+				sync option: true
 			</button>
-			<button @click="openNestedHook({sync: false, returns: true})">
-				open nested modal with async beforeClose hook that returns true
+			<br>
+			<button @click="openNested({ propSync: true, propReturn: true, optSync: true, optReturn: true })">
+				sync prop & option: true
 			</button>
-			<button @click="openNestedHook({sync: false, returns: false})">
-				open nested modal with async beforeClose hook that returns false
+			<br>
+			sync beforeClose hooks: false
+			<br>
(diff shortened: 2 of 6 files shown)


laurenhu (approved):
Thank you for the comprehensive lab breaking this down!

PR #455, opened 2022-10-07
fix(pininput): support browser autofill behavior

### LINKED ISSUE #453: PinInput Triggers Chrome Incorrect Autofill for All but First Input
### Bug description

Chrome will try to guess autofill settings despite having `autocomplete="off"` for the second and beyond inputs in `PinInput` (see repro image). Instead, we should try to replicate what ShopPay has done, where they use a single input and fake the input boxes as users type.

Markup on shop pay: 
image

Shop pay UI:
shop pay pininput
 

### Reproduction

chrome pininput

### Environment

```shell
N/A
```

### Addressed by

_No response_

### Can you contribute a fix?

- [X] I’m interested in opening a pull request for this issue.

tatems:
Closes #453
This PR refactors the `PinInputControl` component to contain a single `input` rather than one for each 'cell'. This is to avoid autofill issues with Chrome, which will ignore `autocomplete="off"` if it decides to. With a single input with `autocomplete="one-time-code"`, we should not run into this anymore.

**Before**
image

**After**
image

## Other info
- current Pininput docs: https://square.github.io/maker/styleguide/latest-stable/#/PinInput
- PinInput docs with this PR's changes: https://square.github.io/maker/styleguide/tatems/pininput-single-input/#/PinInput


src/components/PinInput/src/PinInputControl.vue
```diff
@@ -76,131 +90,113 @@ export default {
...
-			if (this.attemptSplitPinIntoInputs(inputValue, index)) {
-				return;
-			}
-
-			this.$set(this.pin, index, inputValue);
+			this.handlePinChange(input);
+		},
 
-			const firstIncompleteCellIndex = this.findFirstIncompletePinCellIndex();
+		onPastePin(event) {
```
mattmewton, line 145: Is this still being used?

src/components/PinInput/src/PinInputControl.vue
```diff
@@ -76,131 +90,113 @@ export default {
...
+			this.caretPosition = this.isFocused
+				? this.$refs.input.selectionStart
+				: undefined;
 		},
 
-		resetPin() {
-			const newArray = (new Array(this.pinLength)).fill('');
-			this.$set(this, 'pin', newArray);
+		handleComplete(pin) {
+			this.completeTimeout = setTimeout(() => this.$emit('complete', pin), COMPLETE_TIMEOUT_MS);
```
mattmewton, line 186: We should be sure to clear this timeout on destroy as well

src/components/PinInput/src/PinInputControl.vue
```diff
@@ -76,131 +90,113 @@ export default {
...
 		 * shakes and clears pininput
 		 * @public
 		 */
 		shakeAndClearInputs() {
-			this.resetPin();
-			this.focusOnPinCell(0);
+			this.inputValue = '';
 			this.isShaking = true;
 			const TIMEOUT_LENGTH_MS = 1000;
 			setTimeout(() => {
```
mattmewton, line 197: This too, this timeout should probably be cast to a variable and cleared on destroy

src/components/PinInput/src/PinInputControl.vue
```diff
@@ -209,16 +205,33 @@ export default {
...
+	width: var(--pin-input-width);
+}
+
+.PinInput {
+	position: absolute;
+	top: 0;
+	right: 0;
+	bottom: 0;
+	left: 0;
+	z-index: 1;
```
mattmewton, line 218: Does Maker have any z-indexing variables? Would be handy to ensure this isn't layered above/below other components

pretzelhammer: Maker doesn't have z-indexing variables. They're not that useful anyway, as they only affect the relative z-ordering of items within a particular [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). Generally, where possible, we let the order of the items in the DOM dictate how things stack so that devs have control over this within their own apps, e.g.
```
<template>
  <!-- if an app is setup like this, dialogs will appear above modals -->
  <m-dialog-layer />
  <m-modal-layer />
  
  <!-- but if it's setup like this, modals will appear above dialogs -->
  <m-modal-layer />
  <m-dialog-layer />
</template>
```
and where it's not possible, or where we have a thing we know we want to just slightly be above other things even within a particular stacking context, we usually just slap a `z-index: 1` on it and call it a day, so I think it's fine here as well.

mattmewton: Makes sense to me. Just wanted to see if there was an internal convention here. I agree that the layering concerns should be owned by the applications using Maker, generally speaking.

pretzelhammer:
I played around with the [new version on the styleguide](https://square.github.io/maker/styleguide/tatems/pininput-single-input/#/PinInput) and it functions great, I think it's a solid improvement over the [old/current version](https://square.github.io/maker/styleguide/latest-stable/#/PinInput). I like:
1. the new version enforces that the code is input from left-to-right, one number at a time. The old version allowed inputting the code in arbitrary cells which was always weird.
2. the new version fixes a bug it seems like we overlooked in the old version, which is: pressing backspace while focused on the last cell would delete the last 2 cells of numbers at once, which is weird. The new version deletes only 1 number for every backspace press which is nice.
3. the new version actually functions more like a regular input in general, for example like when entering leading numbers, where the old version was like a bunch of random inputs loosely coordinated together.

However, there's a couple small visual regressions.

Old Version, the cursor is perfectly centered within an input cell:
image

New Version, the cursor is not centered by offset slightly to the left:
image

Old Version, cursor after the last number in the last cell appears immediately after the number:
image

New Version, cursor after the last number in the last cell appears offset waaay to the side, almost outside of the cell itself:
image

src/components/PinInput/src/PinInputControl.vue
```diff
@@ -1,51 +1,65 @@
...
 	props: {
 		/**
-		 * Lenth of pin
+		 * Length of pin
 		 */
 		pinLength: {
 			type: Number,
 			default: DEFAULT_INPUT_SIZE,
-			validator: (value) => value > 0,
+			validator: (value) => value > 0 && value <= DEFAULT_INPUT_SIZE,
```
pretzelhammer, line 62: oh, interesting, is there a reason why we're now limiting the max pin input size to 6? I'm fine either having a limit or not, just curious why you added this.

src/components/PinInput/src/PinInputControl.vue
pretzelhammer, line 186: Can you explain, or add a code comment, why this timeout is necessary? What goes wrong or what breaks if we just emit `complete` in the same frame that the last number is entered?

tatems: I was trying to match the old functionality that had it on paste, but I'll rework things a bit. I realize I didn't actually hook things up for pasting.

src/components/PinInput/src/PinInputControl.vue
tatems: Whoops, it should be. Will fix.

tatems: Actually we can drop it.

src/components/PinInput/src/PinInputControl.vue
tatems: Done

src/components/PinInput/src/PinInputControl.vue
tatems: Done

src/components/PinInput/src/PinInputControl.vue
tatems: Adjusted to use order of markup instead of explicit z-index

src/components/PinInput/src/PinInputControl.vue
tatems: Let me try with larger sizes and see what breaks.

tatems: Reverted

src/components/PinInput/src/PinInputControl.vue
tatems: Removed this. The only thing is if the validation is quick, users won't see the last input value. Not sure if that's what we want.

https://user-images.githubusercontent.com/3402466/194622690-107d19a8-61b5-4568-abbd-c7d6279bdf47.mov

pretzelhammer: i think it's okay because the invalid state / shake animation is managed by the user code not the library code, so if users want to introduce a small delay/pause before marking the pininput invalid they can do so, and the library remains unopinionated in this regard, so i think it's fine/good to keep it like this.

mattmewton (approved):

pretzelhammer (approved):
I found a small bug, but it seems pretty niche so it's not a merge blocker so I'm approving this PR.

The bug, if you have time to fix it, is that when a cell is selected and backspace is pressed it always deletes the number in the cell to the left of the selected cell, this has two consequences:

(1) in this state, it's not possible to delete the first number using backspace:

image

it's a little counterintuitive, but the 2nd cell has to be focused and then pressing backspace from that cell will delete the contents in cell 1, e.g.

image

(2) in general, to delete the number inside of any cell using the backspace key, the cell to the right of the cell needs to be focused, which is fine if the cell to the right is empty but is slightly confusing if the cell to the right contains a number

the behavior seems to only work correctly in the case of the last cell being focused and full, in which case backspace deletes the number IN the focused cell and moves to the left, instead of deleting the number in the cell to the left and moving left

i imagine getting this correct might be a little troublesome since the backspace behavior is different depending on whether the focused cell has or doesn't have a number in it, again, not a merge blocker if it would be too troublesome to fix and it's urgent to get this out, but if we could fix it eventually that would be great

pretzelhammer:
thanks for fixing the small niche bug in https://github.com/square/maker/pull/455/commits/952dba8b055f3f4f670390a11b202dc02bb40794 ! code looks good and behavior is perfect now!

PR #469, opened 2022-10-20
chore: upgrade to node 16

privatenumber:

Node v12 is no longer supported.

Upgrade Node to latest LTS


privatenumber:
Yeah, should be fine since the deploys succeeded.

PR #470, opened 2022-10-20
docs: adds color theming controls to styleguide

pretzelhammer:

no color theming in styleguide, it's common for devs to overlook certain theming related bugs since most devs still develop the components assuming the background color is light and never check the component against dark colored backgrounds

adds a subtle button in the center of the left side of the screen:
Pasted_Image_10_20_22__4_55_PM

Which, when clicked, opens up a blade that allows the user or dev to customize the styleguide's colors:
Pasted_Image_10_20_22__4_57_PM

And here's an example of changing the background to black and the primary color to hot pink:
Pasted_Image_10_20_22__4_58_PM

ALSO, this PR adds a new npm script called `styleguide-run` which can be run using `npm run styleguide-run` and it'll open & serve your current branch's styleguide code ASSUMING you have run `styleguide-build` since your last save. I haven't bothered adding an automatic code watcher & builder yet but will probably do so eventually.

## Other information

example: https://square.github.io/maker/styleguide/styleguide-theming-controls/#/Choice


laurenhu:
Could we do something like this? Where its fixed as a drawer over just the examples area with a descriptive label? I think that will make it more discoverable and accessible when looking over the styleguide.
Screen Shot 2022-10-21 at 2 38 36 PM

Screen Shot 2022-10-21 at 2 38 27 PM

pretzelhammer:
@laurenhu i've updated the PR after our discussion on slack, please take another look. styleguide here: https://square.github.io/maker/styleguide/styleguide-theming-controls/#/

laurenhu (approved):
🌈 🌈 🌈

PR #472, opened 2022-10-21
ci: upgrade semantic-release

privatenumber:

Outdated `semantic-release`.

Upgrade to latest `semantic-release` v19.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,24 +14,29 @@ jobs:
     steps:
       - name: Checkout Maker master branch
         uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


pretzelhammer (approved):
I was about to ask if you remembered why we pinned it to v17 in the first place but after a little digging I found [the PR where we pinned it](https://github.com/square/maker/pull/133) and it was because semantic-release v18+ required node v14+ but we had node pinned to v12, but since we just [upgraded node to v16](https://github.com/square/maker/pull/469) we should we safe to upgrade semantic-release to v19 now.

Thanks for the PR!

PR #473, opened 2022-10-21
build: build to `dist` directory

privatenumber:

Multiple distribution files were emitted to the root of the project and was disorganized. This made ignoring them more cumbersome because they had to be manually hard-coded.

- Emit all distribution files to `dist`, and move them to the root on [prepack](https://docs.npmjs.com/cli/v8/using-npm/scripts#:~:text=npm%20publish.-,prepack,-Runs%20BEFORE%20a).
- Ignore `dist` directory from git. (I'll be able to start ignoring `dist` in other tools too.)

### Context
I'm working on a ECOM/Weebly linting library and it makes it a lot easier if we can ignore one directory to ignore all distribution files. I also want to start linting the whole project and exclude specific folders, instead of linting specific folders.

### Follow up questions
Curious what impact this may have on local development.

- Is `npm link`commonly used?
  For this purpose, devs should be able to run `npm run prepack` to yield the same result as before.

- Do consuming projects support export maps?

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "sync-to-latest": "node ./build/sync-to-latest",
     "pull-deploys": "node ./build/deploys/pull.js",
     "push-deploys": "node ./build/deploys/push.js",
-    "prepublishOnly": "npm run build"
+    "prepack": "npm run build && mv dist/* ."
   },
   "husky": {
     "hooks": {
--- a/.gitignore
+++ b/.gitignore
@@ -25,8 +25,7 @@ node_modules/
 .env
 
 # NPM package distribution files
-/components/
-/utils/
+/dist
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "sync-to-latest": "node ./build/sync-to-latest",
     "pull-deploys": "node ./build/deploys/pull.js",
     "push-deploys": "node ./build/deploys/push.js",
-    "prepublishOnly": "npm run build"
+    "prepack": "npm run build && mv dist/* ."
```
privatenumber, line 30: Changed to `prepack` so it runs on publish and `npm pack` (for local testing).

privatenumber:
Looks like Package size report needs to be updated to run pre-publish hooks.

PR #475, opened 2022-10-21
feat(button): add anchor tag option

laurenhu:
Buttons don't support anchor tags styled like buttons, a very common web convention.

Adds `tag` prop to MButton to allow them to render anchor tags instead of buttons for links.

### DIFF
--- a/src/components/Button/src/Button.vue
+++ b/src/components/Button/src/Button.vue
@@ -1,5 +1,6 @@
 <template>
-	<button
+	<component
+		:is="tag"
 		:class="[
 			$s.Button,
 			$s[`variant_${resolvedVariant}`],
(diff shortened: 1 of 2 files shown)


src/components/Button/src/Button.vue
```diff
@@ -130,6 +133,14 @@ export default {
 			type: String,
 			default: undefined,
 		},
+		/**
+		 * Tag of button
+		 * @values button, a
+		 */
+		tag: {
+			type: String,
+			default: 'a',
```
pretzelhammer, line 142: the default should be `'button'` and we should add a validator that restricts people to passing only `'button'` or `'a'`

pretzelhammer (approved):

PR #477, opened 2022-10-24
fix(button): make textPattern themeable

laurenhu:
Currently `MButton`'s `textPattern` prop is not themeable.

Makes `MButton`'s `textPattern` prop themeable.


bashunaimiroy (approved):
looks like this would fix my issue where textPattern is not themeable

pretzelhammer (approved):

PR #482, opened 2022-10-27
feat(container): stop rendering header when there's no header content

laurenhu:
When there's no header content, `MContainer` still renders an empty header wrapper.

Render no header wrapper if there's no header content.

### DIFF
--- a/src/components/Container/src/Container.vue
+++ b/src/components/Container/src/Container.vue
@@ -8,7 +8,10 @@
 		:style="style"
 		v-on="$listeners"
 	>
-		<header :class="$s.Header">
+		<header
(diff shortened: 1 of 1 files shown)


src/components/Container/src/Container.vue
```diff
@@ -8,7 +8,10 @@
 		:style="style"
 		v-on="$listeners"
 	>
-		<header :class="$s.Header">
+		<header
+			v-if="hasLabel || hasSublabel || hasRequirementLabel"
```
tatems, line 12: Might make sense to put this into a computed prop with a descriptive name

pretzelhammer (approved):

PR #485, opened 2022-10-31
fix(theme): update contextual colors

laurenhu:
Our current contextual colors are not as optimized against colorful MTheme backgrounds as they are against black and white backgrounds.
Screen Shot 2022-10-31 at 1 14 11 PM
Screen Shot 2022-10-31 at 1 15 25 PM

Some design-suggested changes to the contextual color logic for colorful backgrounds.
Screen Shot 2022-10-31 at 1 07 32 PM
Screen Shot 2022-10-31 at 1 09 49 PM

### DIFF
--- a/src/utils/maker-colors.js
+++ b/src/utils/maker-colors.js
@@ -73,8 +73,8 @@ function isDark(hex) {
 }
 
 function enoughContrastForFill(hex1, hex2) {
-	const CONTRAST_FILL = 0.25;
-	return colord(hex1).delta(hex2) >= CONTRAST_FILL;
+	const MAKER_DELTA_FILL = 0.25;
+	return colord(hex1).delta(hex2) >= MAKER_DELTA_FILL;
 }
 
 function enoughContrastForText(hex1, hex2) {
--- a/src/utils/maker-colors.js
+++ b/src/utils/maker-colors.js
@@ -1,9 +1,9 @@
 import { colord, extend } from 'colord';
 import a11yPlugin from 'colord/plugins/a11y';
 import mixPlugin from 'colord/plugins/mix';
+import labPlugin from 'colord/plugins/lab';
 import { cloneDeep } from 'lodash';
 import {
-	WCAG_CONTRAST_TEXT,
 	WCAG_CONTRAST_TITLE,
 	DARK_COLOR_LUMINANCE_THRESHOLD,
 	getContrast,
@@ -12,7 +12,7 @@ import defaultColors from '../components/Theme/src/default-colors';
 
 const DEFAULT_COLORS = defaultColors();
 
-extend([a11yPlugin, mixPlugin]);
+extend([a11yPlugin, mixPlugin, labPlugin]);
 
 const NEUTRAL_RATIOS = {
 	light: {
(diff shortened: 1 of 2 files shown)


src/utils/maker-colors.js
```diff
@@ -73,6 +73,11 @@ function isDark(hex) {
 }
 
 function enoughContrastForFill(hex1, hex2) {
+	const CONTRAST_FILL = 0.25;
+	return colord(hex1).delta(hex2) >= CONTRAST_FILL;
```
pretzelhammer, line 77: I don't understand why we're using `colord(hex1).delta(hex2)` as the contrast check in the `enoughContrastForFill` function but `colord(hex1).contrast(hex2)` as the contrast check in the `enoughContrastForText` function. Can we standardize on one and use that for all contrast checks?

laurenhu: I made them different because we primarily use the fill for solid backgrounds that have to be differentiated but note necessarily legible whereas text needs to pass a standardized WCAG contrast threshold. A color could look fine as a fill but not pass WCAG standards for text:

Primary color doesn't pass contrast, but looks fine as a fill color:
Screen Shot 2022-10-31 at 2 05 52 PM
Screen Shot 2022-10-31 at 2 05 26 PM

pretzelhammer: Yup, I get that. My complaint has to do with using different functions for the same purpose. I would suggest renaming `CONTRAST_FILL` to `MAKER_CONTRAST_FILL` so it's clearly distinguished that this is an arbitrary threshold that we came up with and that didn't come from the WCAG and then we can do `colord(hex1).contrast(hex2) >= MAKER_CONTRAST_FILL` within `enoughContrastForFill` which would allow us to get rid of the `.delta()` method call and the `labPlugin` extension. Going from `.delta()` to `.contrast()` will probably require changing the constant value from `0.25` to something else, like `2` or `1.5` or something, but I think it'll probably work similarly to how it works now, right?

laurenhu: They're not exactly the same - delta passes more colors that might have less contrast due to luminance but are visually very different based on their hues:

1. contrast at 1.25:
Screen Shot 2022-10-31 at 2 55 43 PM
Screen Shot 2022-10-31 at 2 55 53 PM

2. contrast at 1.3:
Screen Shot 2022-10-31 at 2 48 01 PM
Screen Shot 2022-10-31 at 2 48 07 PM

3. contrast at 1.5:
Screen Shot 2022-10-31 at 3 01 56 PM
Screen Shot 2022-10-31 at 3 02 02 PM

4. delta at .25:
Screen Shot 2022-10-31 at 2 56 17 PM
Screen Shot 2022-10-31 at 2 56 25 PM


If we feel strongly about not including an additional colord extension, I'm fine with going with option 3, but costs us some colors that could still work against the backgrounds.

pretzelhammer: Thanks for explaining. If you think `delta` is worth it that's good enough for me, but can you please rename the constant `CONTRAST_FILL` to `MAKER_DELTA_FILL`? I feel like it's a lot more descriptive and follows the convention established by `WCAG_CONTRAST_TITLE` and `WCAG_CONTRAST_TEXT`, i.e. `{ORG}_{COLORD-METHOD}_{PURPOSE}`.

pretzelhammer (approved):
Thanks for explaining, I understand the reasoning behind the changes now. This looks good, but please rename `CONTRAST_FILL` constant to `MAKER_DELTA_FILL` prior to merging. Thanks!

PR #488, opened 2022-11-03
feat(ordercard): add OrderCard component

allensquare:

We plan to use this for order history cards and subscription cards in buyer management.
Following figma designs here: https://www.figma.com/file/kmUoROmFpcprMw7xW2YIT9/Subscriptions-in-Customer-Accounts-%E3%83%BC-Buyer-%E2%80%94-Erin-%E2%80%94-Q322?node-id=7699%3A29177

Added new OrderCard component to display an Order with title, label, thumbnails and action buttons. 

## Other information

https://square.github.io/maker/styleguide/order-card-component/#/OrderCard

Screen Shot 2022-11-02 at 4 55 23 PM

### DIFF
--- a/src/components/OrderCard/README.md
+++ b/src/components/OrderCard/README.md
@@ -1,0 +1,97 @@
+# OrderCard
+
+Use OrderCard to display order information
+
+
+## Examples
+```vue
+<template>
+	<div class="container">
+		<m-order-card
+			:label="'Order Label'"
+			:title="'Order Title'"
+			:thumbnails="thumbnails"
+			:primary-action-text="'Primary'"
+			:secondary-action-text="'Secondary'"
+		/>
+		<m-divider />
+		<m-order-card
+			:title="'Order with no label'"
+			:thumbnails="thumbnails"
+			:primary-action-text="'Primary'"
+			:secondary-action-text="'Secondary'"
+		/>
+		<m-divider />
+		<m-order-card
+			:label="'Order Label'"
+			:title="'Order with fewer thumbnails'"
+			:thumbnails="fewerThumbnails"
+			:primary-action-text="'Primary'"
+		/>
+		<m-divider />
+		<m-order-card
+			:label="'Order Label'"
+			:title="'Order with no thumbnails'"
+			:primary-action-text="'Primary'"
+			:secondary-action-text="'Secondary'"
+		/>
+		<m-divider />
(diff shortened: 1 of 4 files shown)


tatems (changes requested):
Can we add doc blocks for each of the `props` and `emit`s?

pretzelhammer (changes requested):
HI Allen, thanks for your PR!

I'm not sure about the name of this component because "Card" implies this is a container component which has a border or a different background color, although this component has neither of those features. This is more of a generic layout or container component.

Also, I think this component looks very, very similar to our existing [MContainer component](https://square.github.io/maker/styleguide/latest-stable/#/Container). The only differences I can see are:

1. MContainer puts the subtitle below the title instead of above it.
2. MContainer doesn't have an actions slot for buttons.

I think it's worth pushing back on Design on 1. Putting subtitles below titles is already an established pattern across multiple Maker components so we should stick with it unless there's a strong Design argument for this one particular exception for this one particular component otherwise. As for 2. we can just add an actions slot to MContainer.

Once 1. & 2. are resolved we can use MContainer for the designs and don't need a dedicated MOrderCard component.

I think MThumbnails is a good component and should exist as a standalone independent component separate from MOrderCard or MContainer. Also, I think MThumbnails should emit some events when certain thumbnails are clicked or when the overflow thumbnail is clicked in case the dev wants to react to these user actions.

laurenhu:
To follow up in @pretzelhammer feedback - I notice the padding on this layout is also different than what we offer on MContainer. In addition to action slots for the buttons on MContainer we could also add a `padding` prop.

allensquare:
@pretzelhammer @laurenhu

Great ideas, I've looped with UX and we think it's a good idea to leverage the existing MContainer and switch to keeping just a thumbnails component here as a new component. For the #2 mentioned, I was thinking of just keeping the container content part generic and provide the buttons + thumbnails in the content slot without needing to introduce new action slots to the container to keep it simple, thoughts?

pretzelhammer:
@allensquare yup, regarding 2. skipping the actions slot and providing the buttons as part of the default content slot sounds good to me too. In that case we'd only need to change this PR to create a new standalone MThumbnails component.

allensquare:
Created a separate PRs for the Thumbnails component and adding new Container padding prop - closing this one.

PR #490, opened 2022-11-03
feat(thumbnails): add Thumbnails component

allensquare:

We need a component to display list of images with overflow text displayed when hitting the max thumbnail count. See desgins - [Figma](https://www.figma.com/file/kmUoROmFpcprMw7xW2YIT9/Subscriptions-in-Customer-Accounts-%E3%83%BC-Buyer-%E2%80%94-Erin-%E2%80%94-Q322?node-id=7699%3A29177)

Added a thumbnails component

### DIFF
--- a/src/components/Thumbnails/src/Thumbnails.vue
+++ b/src/components/Thumbnails/src/Thumbnails.vue
@@ -1,0 +1,113 @@
+<template>
+	<div
+		:class="$s.Thumbnails"
+		:style="thumbnailStyles"
+	>
+		<template
+			v-for="(thumbnail, index) in thumbnails.slice(0, maxThumbnails)"
+		>
+			<m-image
+				:key="`image-${index}`"
+				:class="$s.ThumbnailImage"
+				:src="thumbnail"
+				@click="$emit('thumbnail:click', thumbnail)"
+			/>
+		</template>
+		<div
+			v-if="hasThumbnailOverflow"
+			@click="$emit('overflow:click')"
+		>
+			<m-text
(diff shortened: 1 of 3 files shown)


src/components/Thumbnails/src/Thumbnails.vue
```diff
@@ -0,0 +1,113 @@
...
+			type: Number,
+			default: DEFAULT_THUMBNAIL_COUNT,
+		},
+		/**
+		 * Overflow text size. Size of text as step in fluid scale.
+		 * @values 7, 6, 5, 4, 3, 2, 1, 0, -1, -2
+		 */
+		overflowTextSize: {
+			type: Number,
+			default: DEFAULT_OVERFLOW_TEXT_SIZE,
```
laurenhu, line 74: We could default this to `undefined`, similar to the `size` prop on `MText` so by default it inherits the size of the context.

src/components/Thumbnails/src/Thumbnails.vue
```diff
@@ -0,0 +1,113 @@
...
+			};
+		},
+	},
+};
+</script>
+
+<style module="$s">
+.Thumbnails {
+	display: flex;
+	align-items: center;
```
laurenhu, line 103: We could include gap: 8px here instead of the margin on ThumbnailImage. That way there's no trailing right margin on a set of thumbnails with no overflow.

laurenhu (approved):
This looks great! 👍 Two minor suggestions that you can apply or ignore at your discretion.

PR #491, opened 2022-11-03
feat(container): add padding prop to container component

allensquare:

Container padding cannot be customized

Added new padding prop to Container component. Updated docs.


src/components/Container/src/Container.vue
```diff
@@ -112,6 +112,20 @@ export default {
 			default: undefined,
 			validator: (color) => colord(color).isValid(),
 		},
+		/**
+		 * Padding of the container
+		 */
+		padding: {
+			type: String,
+			default: '16px 24px',
+			validator: (padding) => {
```
laurenhu, line 121: I believe we can import `import cssValidator from '@square/maker/utils/css-validator';` and just do `validator: cssValidator('padding'),` here.

laurenhu (approved):

PR #495, opened 2022-11-08
docs: separate advanced customization props into their own table

laurenhu:
Advanced customization options are less-used props which can clutter up the component docs.

Separates any component prop whose description marks it as `Advanced:` into a separate table.


build/generate-readme-api-tables/generate-api-tables.js
```diff
@@ -84,22 +84,37 @@ function getPropsTable({ displayName, props, tags }, tablePrefix) {
...
-				...mapIn(props, (propInfo) => [
+				...mapIn(standardProps, (propInfo) => [
 					getPropName(propInfo),
 					getPropType(propInfo),
 					getPropDefaultValue(propInfo),
 					getPropPossibleValues(propInfo),
 					getPropDescription(propInfo),
 				]),
 			]) : '',
+			advancedProps ? '### Advanced customization props' : '',
```
pretzelhammer, line 106: For the advancedProps title can we do this instead?:
```
`${tablePrefix ? `${tablePrefix} ` : ''}Props (Advanced)`,
```
I think it matches the title scheme a bit better, also will work for directories which export multiple components, e.g. ActionBar, which exports ActionBarButton and InlineActionBar both with their own sets of props

src/components/Button/src/Button.vue
```diff
@@ -194,55 +180,91 @@ export default {
...
+		},
+		/**
+		 * Toggles button loading state
+		 */
+		loading: {
+			type: Boolean,
+			default: false,
+		},
+		/**
+		 * Advanced: MText pattern in button label
```
pretzelhammer, line 220: Rather than prefixing the prop description with `Advanced:` can we just tag the JSDoc comment with `@advanced`? We already use tags in a few other places, notably `@public` for marking public methods or `@private` for marking private props & slots. So i think we should stick with that convention and use `@advanced` here.

pretzelhammer:
This looks really good, and you got it out super quick! I left a couple small comments. Also, I think it would be nice if we alphabetically sort props (and slots and events) before we generate the tables for them, as I think it would make it easier to find stuff, but that's optional, so we can pass on it for this PR and I'll just make a PR for it myself later when I have time.

pretzelhammer (approved):

PR #497, opened 2022-11-09
docs: alpha sort api docs

pretzelhammer:

props, slots, events, and methods tables in the api docs of components were always in a random order, now they will be generated in sorted alphabetical order

stated above, actual impl here: https://github.com/square/maker/pull/497/files#diff-68901b8f2c4e06cdd0c84154c2207e4892a475ff9531b038a79ebfe27b0aa719R11-R17

## Other information

nope


laurenhu:
I'm a little concerned that by alphabetizing the props, we lose some of the priority or meaningful groupings we get from the component prop order. For instance, on Button, we end up with `color` separated from `text-color` and `pattern` and `variant`, which are relatively more important props buried toward the middle/bottom of the list.
Current:
Screen Shot 2022-11-15 at 4 41 05 PM

Alphabetized:
Screen Shot 2022-11-15 at 4 41 16 PM

pretzelhammer:
@laurenhu okay, fair enough, i'm cool with closing this.

PR #499, opened 2022-11-10
feat(button): add hover text pattern support

laurenhu:
Button currently does not support hover text styles.

Adds a `textPatternHover` prop to Button, removes MText from inside button in favor of directly applying text pattern styles.

### DIFF
--- a/src/components/Button/src/Button.vue
+++ b/src/components/Button/src/Button.vue
@@ -48,6 +48,7 @@
 </template>
 
 <script>
+import { kebabCase } from 'lodash';
 import cssValidator from '@square/maker/utils/css-validator';
 import { colord } from 'colord';
 import { getContrast } from '@square/maker/utils/get-contrast';
@@ -106,19 +107,19 @@ const TEXT_STYLES = new Set([
 	'textDecoration',
 ]);
 
-function kebabCase(string) {
-	return string.replace(/([\da-z])([A-Z])/g, '$1-$2').toLowerCase();
+function isValidCss(style, value) {
+	if (global.CSS) {
+		return global.CSS.supports(style, value);
+	}
+	return true;
 }
 
 function formatCssStyles(theme, styles, suffix) {
 	const formattedStyles = {};
 	for (const [style, value] of Object.entries(styles)) {
-		if (TEXT_STYLES.has(style) && cssValidator(style, value)) {
-			if (value.startsWith('@')) {
-				formattedStyles[`--${kebabCase(style)}${suffix}`] = theme.resolve(value);
-			} else {
-				formattedStyles[`--${kebabCase(style)}${suffix}`] = value;
-			}
+		const cssStyle = kebabCase(style);
+		if (TEXT_STYLES.has(style) && isValidCss(cssStyle, value)) {
+			formattedStyles[`--${cssStyle}${suffix}`] = theme.resolve(value);
 		}
 	}
 	return formattedStyles;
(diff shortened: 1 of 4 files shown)


src/components/Button/src/Button.vue
```diff
@@ -104,6 +97,33 @@ function setColorVariables(tokens, variant) {
...
+
+function kebabCase(string) {
+	return string.replace(/([\da-z])([A-Z])/g, '$1-$2').toLowerCase();
+}
+
+function formatCssStyles(theme, styles, suffix) {
+	const formattedStyles = {};
+	for (const [style, value] of Object.entries(styles)) {
+		if (TEXT_STYLES.has(style) && cssValidator(style, value)) {
+			if (value.startsWith('@')) {
```
pretzelhammer, line 117: i don't think you have to do this `if` check, this line:
````js
formattedStyles[`--${kebabCase(style)}${suffix}`] = theme.resolve(value);
````
works regardless of whether `value` is a pointer or not

File: src/components/Button/src/Button.vue [hunk 2]
```diff
@@ -353,8 +381,19 @@ export default {
 				boxShadow: this.resolvedBoxShadow,
 				boxShadowHover: this.resolvedBoxShadowHover,
 			};
+
+			const { resolvedTextPattern, resolvedTextPatternHover, theme } = this;
+			const textPattern = resolvedTextPattern
```
pretzelhammer, line 386: i think this line can be simplified down to just:
```js
const textPattern = theme?.text?.patterns?.[resolvedTextPattern] || {};
```

src/components/Button/src/Button.vue
```diff
@@ -353,8 +381,19 @@ export default {
...
+				? theme?.text?.patterns?.[resolvedTextPattern] || {} : {};
+			const textPatternHover = resolvedTextPatternHover
```
pretzelhammer, line 388: i think this line can be simplified down to:
```js
const textPatternHover = theme.text.patterns?.[resolvedTextPatternHover] || {};
```

src/components/Button/src/Button.vue
```diff
@@ -104,6 +97,33 @@ function setColorVariables(tokens, variant) {
...
+]);
+
+function kebabCase(string) {
+	return string.replace(/([\da-z])([A-Z])/g, '$1-$2').toLowerCase();
+}
+
+function formatCssStyles(theme, styles, suffix) {
+	const formattedStyles = {};
+	for (const [style, value] of Object.entries(styles)) {
+		if (TEXT_STYLES.has(style) && cssValidator(style, value)) {
```
pretzelhammer, line 116: `cssValidator(style, value)` will always return a function which will always evaluate to true. It's meant for quickly setting cssValidator functions on Vue props. I think we need to add a new util function called `isValidCss(cssProperty, cssValue)` and use that here instead. Also, the current use is incorrect since we're passing the css properties in camelCase but for `CSS.supports` to work correctly we need to pass the css properties in kebab-case, example screenshot: 
image

src/components/Button/src/Button.vue
```diff
@@ -104,6 +97,33 @@ function setColorVariables(tokens, variant) {
...
+const TEXT_STYLES = new Set([
+	'fontFamily',
+	'fontWeight',
+	'fontSize',
+	'fontStyle',
+	'textTransform',
+	'textDecoration',
+]);
+
+function kebabCase(string) {
```
pretzelhammer, line 109: we don't need to implement this, it already exists in [lodash](https://lodash.com/docs/4.17.15#kebabCase), and their impl works for camelCase identifiers that are longer than two words. let's remove this and add
```js
import { kebabCase } from 'lodash';
```
to the top of the file.

src/components/Button/src/Button.vue
```diff
@@ -104,6 +98,33 @@ function setColorVariables(tokens, variant) {
...
+	'fontStyle',
+	'textTransform',
+	'textDecoration',
+]);
+
+function isValidCss(style, value) {
+	if (global.CSS) {
+		return global.CSS.supports(style, value);
+	}
+	return false;
```
pretzelhammer, line 114: `global.css` is undefined when server-side rendering, so this will always return `false` in that use-case, which might break things or just annoy people, lol. I think defaulting to `true` is the lesser of both evils, assuming you haven't already tested this with SSR and it works fine.

pretzelhammer (approved):
left a small comment you might wanna address before merging, but otherwise everything else looks good 👍

PR #508, opened 2022-11-29
fix(textbutton): add overflow wrap

kevinlee11:

Long words aren't properly truncated with the `truncate-text` prop
image

Add the overflow-wrap property
image

### DIFF
--- a/src/components/TextButton/src/TextButton.vue
+++ b/src/components/TextButton/src/TextButton.vue
@@ -229,6 +229,7 @@ export default {
 	white-space: initial;
 	text-align: inherit;
 	text-overflow: ellipsis;
+	overflow-wrap: anywhere;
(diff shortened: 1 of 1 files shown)


sohnj877 (approved):

pretzelhammer:
i’m not sure if `overflow-wrap: anywhere` is supported by safari… https://caniuse.com/?search=anywhere

image

can you test this on safari and verify that it works in that browser?

kevinlee11:
Oddly enough looks like caniuse is out of date on this one, works fine on Safari

image

pretzelhammer (approved):
confirmed with Kevin over Slack that this fix works in Safari, it seems like caniuse's browser compat data might be out of date for this particular feature

PR #512, opened 2022-12-13
refactor(image): improve Image loading

berwyn:

With the latest release version, image loading has a few snags. This is especially apparent in Safari, where the user agent frequently loads two or more of the images. In my example case, I have an `MImage` with `src`, `srcset`, and `sizes` all set, which causes Safari to load two images from the `srcset` in two resolutions. Additionally, the internal cache inside `MImage` creates an `<image>` tag that loads its `src`, meaning that every image on the page is loading three separate copies of the same image at different resolutions. This has a huge impact in page load performance, as the number of in-flight requests as well as the amount of data loaded can become quite significant on image-heavy pages.

You can observe this in Safari's network tab using the "group by resource" option. Here, you can see two of the variants are loaded as a result of the `` itself, while a 3rd is loaded from `MImage`'s hidden `<image>` tag.

image

This PR does a couple things to `MImage` that change implementation details without touching the public API
- the internal cache is removed. Maker cannot have any insight into whether an image has or has not been cached, so the `loaded` flag is now reactive to the `load` event fired on every image as its contents are loaded.
- the hidden `<image>` tag used to preload images has been removed. Instead, lazy loaded images will simply lack their `src`+`srcset` attributes. As the `IntersectionObserver` indicates that the image enters the viewport, those attributes will be filled in, allowing the user agent to fetch the image.

The end result is that user agents like Safari will now properly load exactly one candidate from the `src`+`srcset` tags.

Here, you can see Safari's network tab with the same settings shows exactly one version of the image loaded:
image

## Other information

An interesting aside, but the lazyloading process used the `Image` constructor, which maps to the long-deprecated [`<image>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/image) tag. Normally, if this element was attached to the DOM, the user agent would convert it to an ``, but in this case because it remains outside the DOM tree it uses the legacy `<image>` behaviour and would not parse `srcset`/`sizes` values!

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -18,10 +18,11 @@
 					[$s.thumbnail]: isThumbnail,
 				}"
 				:style="style"
-				:srcset="srcset"
+				:srcset="calculatedSrcSet"
 				:sizes="sizes"
-				:src="src"
+				:src="calculatedSrc"
 				v-bind="$attrs"
+				@load="onLoaded"
 				v-on="$listeners"
 			>
 		</m-transition-fade-in>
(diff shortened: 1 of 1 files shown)


src/components/Image/src/Image.vue
```diff
@@ -145,54 +160,29 @@ export default {
...
-			if (this.sizes) {
-				img.sizes = this.sizes;
-			}
-
-			img.addEventListener('load', () => {
-				imgCache.add(this.src + this.srcset);
-				this.loaded = true;
-				this.getImageDimensions();
-			});
+			this.shouldLoad = true;
```
mattmewton, line 185: Should this value be reset to `false` after the initial load?

berwyn: I don't think `shouldLoad` should ever change after it's been set to `true`. Changing it to `false` would cause the `src` and `srcset` values on the element to become unset, causing the image to unload.

pretzelhammer (approved):

mattmewton (approved):

PR #518, opened 2022-12-16
docs: group components in styleguide sidenav

pretzelhammer:

components in styleguide sidenav aren't grouped/categorized/organized in any way

adds groups/categories for components

## Other information

https://square.github.io/maker/styleguide/sidenav-groupings/#/


laurenhu (approved):
This looks great! I like that all ungrouped components are in general so nothing gets lost if we don't manually update the groups.

Only changes I would include:
- Theme feels more like a Container?
- Should we include Notice under Container as well as General? The inline Notice is a general component but the block Notice feels more like a Container
- Should we hide BlockFormControlLayout and InlineFormControlLayout for now? Since they're more internal components that are not documented?

PR #520, opened 2022-12-22
feat(choice): add hover style to choice card, fix active state

laurenhu:
Choice cards do not have hover states

Adds hover and a
Screen Shot 2022-12-22 at 3 28 12 PM
ctive states to ChoiceCard that match outline form components.

### DIFF
--- a/src/components/Choice/src/ChoiceCard.vue
+++ b/src/components/Choice/src/ChoiceCard.vue
@@ -63,23 +63,24 @@ export default {
 </script>
 
 <style module="$s">
-/*
-	Defined in Choice component:
-	--selected-color
-	--selected-contrast-color
-*/
 .ChoiceCard {
+	--color-border-active: $maker-color-neutral-80;
+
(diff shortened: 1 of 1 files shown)


pretzelhammer:
Is this PR outdated? I noticed on the latest-stable the selected color is the primary color:

image

But that on this branch it's just neutral-80:

image

Also, I'm not sure how I feel about:
1. The hover state on ChoiceCards being identical to the selected state
2. ChoiceCards having a hover state but Choices not having a hover state

If we could solve 1 & 2 then that would be awesome.

laurenhu:
@pretzelhammer The decision to change the selected color of a choice card to always be neutral instead of the primary color like the choice options was based on design feedback our of a desire for its styling to more closely match the form elements. I've:
- Updated the select state on choice cards to be more distinct from the hover state
- Added a hover state for choice options.

pretzelhammer (approved):
the updates look great, thanks!

PR #521, opened 2022-12-22
fix(button): don't change a custom button hover color

laurenhu:
Button hover color get's lightened or darkened slightly even it's a custom value.

No longer lighten/darken hover color if it's different than the main button fill color.
Screen Shot 2022-12-22 at 5 24 07 PM
Screen Shot 2022-12-22 at 5 23 44 PM

### DIFF
--- a/src/components/Button/src/Button.vue
+++ b/src/components/Button/src/Button.vue
@@ -75,8 +75,8 @@ function setColorVariables(tokens, variant) {
 	const hoverStateAdjust = 0.08;
 	const activeStateAdjust = 0.16;
 	const focusAlphaAdjust = 0.3;
-	const colorHover = colorMainHover === colorMain
-		? colorMainHover[stateAdjustment](hoverStateAdjust).toHex() : colorMainHover.toHex();
+	const colorHover = tokens.colorHover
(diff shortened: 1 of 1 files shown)


src/components/Button/src/Button.vue
```diff
@@ -75,7 +75,8 @@ function setColorVariables(tokens, variant) {
 	const hoverStateAdjust = 0.08;
 	const activeStateAdjust = 0.16;
 	const focusAlphaAdjust = 0.3;
-	const colorHover = colorMainHover[stateAdjustment](hoverStateAdjust).toHex();
+	const colorHover = colorMainHover === colorMain
```
bashunaimiroy, line 78: this works but I wonder if it might be better to refactor all the hover logic so that if we've already got `tokens.colorHover`, we immediately assign `colorHover` to that value, and we skip calculating the `stateAdjustment` and `hoverStateAdjust` variables.

laurenhu: Updated so that we pass `tokens.colorHover` for hover color but we still need to calculate some of the `stateAdjustment` values for the active and focus states.

pretzelhammer (approved):

PR #523, opened 2023-01-12
feat(image): add object-fit property to image component

mbove-square:

* `object-fit` style is originally set to `cover` which will clip the provided image to the specified image dimensions. This allows the image to be scaled to the given dimensions in addition

* Specify `object-fit` property on the image
* Adds `contain` and `scale-down` object fit properties. `cover` is treated as the default

https://user-images.githubusercontent.com/82115556/212120973-382c6c62-4613-49bf-b6ec-febae3f3dfec.mov

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -111,7 +111,7 @@ export default {
 		},
 		objectFit: {
 			type: String,
-			validator: (fit) => ['cover', 'contain'].includes(fit),
+			validator: (fit) => ['cover', 'contain', 'scale-down'].includes(fit),
 			default: 'cover',
 		},
 	},
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -146,7 +146,7 @@ export default {
 		style() {
 			return {
 				'--image-height': `${this.height}px`,
-				'object-fit': this.objectFit,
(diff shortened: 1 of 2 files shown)


src/components/Image/src/Image.vue
```diff
@@ -141,6 +146,7 @@ export default {
 		style() {
 			return {
 				'--image-height': `${this.height}px`,
+				'object-fit': this.objectFit,
```
ashjtan, line 149: using the `.Image &.shape_arch` style as reference, should this be plugged in the same way as `--image-height` so that you have
```css
.Image {
   object-fit: var(--object-fit);
```

mbove-square: Yeah that makes sense, just applied that change

src/components/Image/src/Image.vue
```diff
@@ -109,6 +109,11 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		objectFit: {
+			type: String,
+			validator: (fit) => ['cover', 'contain', 'scale-down'].includes(fit),
```
laurenhu, line 114: We can probably import cssValidator here to allow all valid values for `'object-fit'.
https://github.com/square/maker/blob/master/src/components/Text/src/Text.vue#L48

laurenhu (approved):
LGTM! One small note about the validator.

PR #525, opened 2023-01-25
feat(cards): add glass variant for cards, rename default to outline

laurenhu:
No cards with glass backdrop.

Adds the `variant` prop to Cards, and styles for the `glass` variant. Renames current default card style to `outline` and makes it the theme default. Adds examples for usage to style guides.

Screenshot 2023-01-25 at 11 56 18 AM

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -220,6 +220,7 @@ export default {
 }
 
 .Image {
+	display: block;
 	width: 100%;
 	height: 100%;
 	object-fit: var(--image-object-fit);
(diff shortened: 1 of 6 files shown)


src/components/Image/src/Image.vue
```diff
@@ -220,6 +220,7 @@ export default {
 }
 
 .Image {
+	display: block;
```
pretzelhammer, line 223: can you comment explaining this change? i don't understand how it's relevant to the rest of the PR

pretzelhammer: cool, thanks for explaining

pretzelhammer (approved):

PR #527, opened 2023-02-01
feat: add `shape` property to `MPill`

berwyn:

We have a project where we'd like to use `MPill` to present informational items, however we'd like to be able to customise the shape of the pill.

This PR adds the `shape` property, accepting either a single shape or a tuple of two shapes to use on each endcap.

## Other information

I've implemented this with the `::before` and `::after` pseudo-elements but I'm not overly fond of this implementation. That being said, I couldn't think of a better way to do it without manually `clip-path`ing the entire capsule uniquely for every pair of shapes.

There's also an ancillary change to let Webpack process `style` tags with the `lang` explicitly set to `postcss`. This doesn't have any actual changes on the build, but fixes syntax highlighting and context-aware tools in editors like VSCode using Volar.

### DIFF
--- a/build/webpack-base-config.js
+++ b/build/webpack-base-config.js
@@ -45,7 +45,7 @@ const webpackBaseConfig = {
 				loader: 'babel-loader',
 			},
 			{
-				test: /\.(post)?css$/,
+				test: /\.css$/,
 				oneOf: [
 					// Matches `<style module>`
 					{
--- a/build/webpack-base-config.js
+++ b/build/webpack-base-config.js
@@ -45,7 +45,7 @@ const webpackBaseConfig = {
 				loader: 'babel-loader',
 			},
 			{
-				test: /\.css$/,
+				test: /\.(post)?css$/,
 				oneOf: [
 					// Matches `<style module>`
 					{
--- a/src/components/Pill/src/Pill.vue
+++ b/src/components/Pill/src/Pill.vue
@@ -71,7 +71,7 @@ export default {
 		shape: {
 			type: [String, Array],
 			default: 'pill',
-			validator: function validatePillShape(input) {
+			validator(input) {
 				if (Array.isArray(input)) {
 					return input.length === MAXIMUM_ALLOWED_SHAPES
 						&& input.every((value) => ALLOWED_SHAPES.has(value));
--- a/src/components/Pill/src/Pill.vue
+++ b/src/components/Pill/src/Pill.vue
@@ -1,6 +1,6 @@
 <template>
 	<span
-		:class="$s.Pill"
+		:class="[$s.Pill, shapeClass]"
 		:style="style"
 		v-bind="$attrs"
 		v-on="$listeners"
@@ -13,6 +13,15 @@
 <script>
 import cssValidator from '@square/maker/utils/css-validator';
 import { MThemeKey, defaultTheme, resolveThemeableProps } from '@square/maker/components/Theme';
+
+const MAXIMUM_ALLOWED_SHAPES = 2;
+const ALLOWED_SHAPES = new Set([
+	'sharp',
+	'rounded',
+	'pill',
+	'point',
+	'ribbon',
+]);
 
 /**
  * @inheritAttrs span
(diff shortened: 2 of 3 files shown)


build/webpack-base-config.js
```diff
@@ -45,7 +45,7 @@ const webpackBaseConfig = {
 				loader: 'babel-loader',
 			},
 			{
-				test: /\.css$/,
+				test: /\.(post)?css$/,
```
privatenumber, line 48: I appreciate the IDE compatibility improvement but would prefer to move this out and address it holistically in a separate Issue/PR for further discussion.

----

Based on https://github.com/johnsoncodehk/volar/issues/765, it seems the recommendation is to use https://marketplace.visualstudio.com/items?itemName=cpylua.language-postcss with the following configuration to make it work with `.css` extensions:
```
  "files.associations": {
    "*.css": "postcss"
  }
```

src/components/Pill/src/Pill.vue
```diff
@@ -14,6 +14,15 @@
 import cssValidator from '@square/maker/utils/css-validator';
 import { MThemeKey, defaultTheme, resolveThemeableProps } from '@square/maker/components/Theme';
 
+const MAXIMUM_ALLOWED_SHAPES = 2;
+const ALLOWED_SHAPES = new Set([
+	'sharp',
```
privatenumber, line 19: The name `sharp` is a little unintuitive to me (also not a shape). Could `rectangle` be a better name?

Screenshot 2023-02-02 at 12 10 54 PM

src/components/Pill/src/Pill.vue
```diff
@@ -52,6 +63,23 @@ export default {
...
 		},
+
+		/**
+		 * The shape the pill should take, or a tuple of shapes for each endcap.
+		 * @values pill, sharp, rounded, point, ribbon
+		 */
+		shape: {
+			type: [String, Array],
+			default: 'pill',
+			validator: function validatePillShape(input) {
```
privatenumber, line 74:
```suggestion
			validator(input) {
```

src/components/Pill/src/Pill.vue
pretzelhammer: The convention in Maker across all other components that have a `shape` prop is to use the names: `pill`, `rounded`, and `squared`. Please rename `sharp` to `squared` so it's consistent with the rest of the library. I think the custom fancy shape names being `point` and `ribbon` are fine.

pretzelhammer (approved):

laurenhu (approved):
I think it might be down to how browsers round the percentages, but I sometimes see a sliver of a gap between the pill and the pseudoelement:

Screenshot 2023-02-02 at 4 41 51 PM

privatenumber:
Just realized the Pill component will no longer be a pill 😫

berwyn:
There's a couple quirks like the gaps that I'm trying to iron out, the bottom of the right endcap on the ribbons with an outline pattern has all sorts of weirdness because of how rounding + `clip-path` interact and it really bugs me :/

I was also torn on making the `MPill` not a pill, but we're trying to surface different shapes of small informational items that are placed inline with their content and `MPill` seemed to fit the bill most. I did debate introducing an `MChip` that sort of matches the role of [Material Design chips](https://m3.material.io/components/chips/overview) but felt like its role would overlap a bit too much with `MPill` :/

berwyn:
Going to put out a call for help here. As Lauren noted, there is sometimes a sliver of a gap between the core element and the psudo-element. I have tried everything I can think of to ensure this never occurs, up to rewriting the component to use a flow wrapper and making the two endcaps discrete elements in the flex row, ensuring the `gap` and other positioning properties are set to `0`.

As far as I can tell, the gap shows up when the elements are laid out such that the border between the two elements falls onto a fractional pixel, leading the layout engine to align the start of the endcap with the next full pixel, and I have no idea how I can fix this short of rebuilding the whole thing to use a custom `clip-path` on the entire capsule for each potential shape combination, or some absolute nonsense with an embedded SVG with various `<g>` elements that can be set as the `background` property on the pseudo-elements :/

PR #528, opened 2023-02-20
refactor(icon)!: render fns instead of vue components

pretzelhammer:

described in migration notes in this pr:
> Before to theme icons you'd have to pass Vue components to the MTheme config object. This had two big limitations: icon components must be compiled either ahead of time or on the fly, and it wasn't possible to customize the prop values on the icon component, it would always be rendered with the defaults. Furthermore, doing this just for icon components was a little odd since we already established the pattern of using render functions for all other components which conditionally render nested content, e.g. MModals, MDialogs, MBlades, MToasts. Now to theme icons you use render functions, which solves all of the problems noted above.

also, while not mentioned in the migration notes (since it's not a breaking change) this PR also adds the ability to set/theme the `fill` property of MIcons separately from the `color` property. this is to enable setting values which are valid fill values but not valid color values, such as `url(#idToSvgDefinitionSuchAsALinearGradient)`.

- theme icons now using render fns instead of vue components
- can theme `fill` property

## Other information

check out the interactive icons lab here: https://square.github.io/maker/lab/icon-render-fns/#/CustomIcons

### DIFF
--- a/lab/components/AsyncInlineSvg.vue
+++ b/lab/components/AsyncInlineSvg.vue
@@ -1,0 +1,135 @@
+<template>
+	<!--
+		hack I temporarily used to fix an issue
+		that seems to have magically fixed itself
+		but I'm keeping this code here in case
+		I wanna experiment more with it later
+	-->
+	<html-fragment
+		v-if="fragment"
+		:class="classes"
+		:html="svgText"
+	/>
+	<!-- eslint-disable vue/no-v-html -->
+	<i
+		v-else
+		:class="classes"
+		v-html="svgText"
+	/>
+</template>
+
+<script>
+import HtmlFragment from './HtmlFragment.vue';
+
+/**
+ * trims anything before the opening svg tag
+ * and anything after the closing svg tag
+ * @param {String} svgText
+ * @return {String}
+ */
+function trimEnds(svgText) {
+	const start = '<svg';
+	const end = '</svg>';
+	const endLength = end.length;
+	const startIndex = svgText.indexOf(start);
+	const endIndex = svgText.indexOf(end) + endLength;
+	return svgText.slice(startIndex, endIndex);
+}
+
+/**
+ * removes width & height attributes from svg
+ * so it doesn't break out of container
+ * @param {String} svgText
+ * @return {String}
+ */
+function removeDims(svgText) {
+	return svgText.replace(/(width|height)=".*"/gim, '');
+}
+
+/**
+ * cleans svg text
+ * @param {String} svgText
+ * @return {String}
+ */
+function cleanText(svgText) {
+	return removeDims(trimEnds(svgText));
+}
+
+/**
+ * Vue component that asynchronously fetches an SVG
+ * from a URL and renders it inline. Rendering the SVG
+ * inline is important because then it can be targeted
+ * by CSS for styling.
+ */
+export default {
+	components: {
+		HtmlFragment,
+	},
+	props: {
+		src: {
+			type: String,
+			required: true,
+		},
+		fragment: {
+			type: Boolean,
+			default: false,
+		},
+		size: {
+			type: String,
+			default: 'medium',
+			validator: (size) => ['medium', 'large'].includes(size),
+		},
+	},
+	data() {
+		return {
+			svgText: '',
+		};
+	},
+	computed: {
+		classes() {
+			return [
+				this.$s.Icon,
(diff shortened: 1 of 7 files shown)


lab/components/HtmlFragment.vue
```diff
@@ -0,0 +1,30 @@
...
+		const nodes = new Vue({
+			beforeCreate() {
+				this.$createElement = h;
+			},
+			template: `<div>${context.props.html}</div>`,
+		}).$mount()._vnode.children;
+		return nodes;
+	},
+};
+</script>
```
privatenumber, line 30: Would rather have you use [vue-frag](https://github.com/privatenumber/vue-frag) here since it's a battle-tested solution.

pretzelhammer: I don't know what I'm doing wrong but I tried using vue-frag and couldn't make it work... I tried:
```
<fragment v-html="svgText" />
```
and
```
<div v-frag v-html="svgText" />
```
and neither worked. No errors during build. No errors in the console at run-time. Just doesn't render. No clue why.

lab/components/AsyncInlineSvg.vue
```diff
@@ -0,0 +1,135 @@
+<template>
+	<!--
+		hack I temporarily used to fix an issue
+		that seems to have magically fixed itself
+		but I'm keeping this code here in case
+		I wanna experiment more with it later
+	-->
+	<html-fragment
+		v-if="fragment"
```
linnea, line 9: Curious why give the option to use fragment instead of just using out of the box for consistency? What use case is there to not use this as a fragment? 

Seems like it's the more favored option, but you have to opt into it.

lab/components/AsyncInlineSvg.vue
```diff
@@ -0,0 +1,135 @@
...
+		:class="classes"
+		:html="svgText"
+	/>
+	<!-- eslint-disable vue/no-v-html -->
+	<i
```
linnea, line 14: Why the italic tag over, say, a span?

linnea (approved):
👍

lab/components/AsyncInlineSvg.vue
pretzelhammer: i thought the `i` in `<i>` stood for `icon`, oops, 😆 .

i think i was mislead to have this belief by Font Awesome. All of their docs recommend using `<i>` but i honestly have no clue why. I'm down to change this to a `<span>` 👍

lab/components/AsyncInlineSvg.vue
pretzelhammer: Yup, you're exactly right, fragment is better and should be the default, the issue is that I was having some problems with getting it to work in my Lab demo which is why I have a toggle to turn it off and on. Ideally, I'd like to ship the fragment-version of AsyncInlineSvg to Website BUT if it's not do-able I think we can live with just having a `span` wrapper around the SVGs. It depends on if @privatenumber has the time to help me debug the issue with `vue-frag` I mentioned in this comment: https://github.com/square/maker/pull/528#discussion_r1116023900

pretzelhammer: Note: this code is part of a Lab demo and is not shipped as part of Maker, it's just a toy thing I was experimenting with. I intend to implement a cleaned up version of AsyncInlineSvg in Website, i don't think the component itself needs to exist as part of Maker.

linnea: 👍 Just wanted to throw some quick notes in while I was reviewing, not a blocker at all

laurenhu (approved):
Very cool! 🤩

So far we're only theming error, warning, success, and info icons on Website so there's not much to fix for the breaking change.

linnea:
@laurenhu I'm only seeing one MIcon usage of 'info', can you point me to the other usages you see? Will be doing the upgrade here in a bit!

lab/components/HtmlFragment.vue
privatenumber: Sorry for the late response.

So, you can't set `v-html` on a Fragment because it doesn't exist in the DOM (same way a real DOM Fragment doesn't exist in the DOM and [doesn't support `innerHTML`](https://stackoverflow.com/questions/8202195/using-document-createdocumentfragment-and-innerhtml-to-manipulate-a-dom)).

I noticed `svgText` just returns a single SVG element though. Is a fragment necessary?

PR #530, opened 2023-03-06
fix(touchcapture): stop generating noisy console error

athammer:
Some events are not cancelable such as scrolling. We try to cancel theses events even though browser's wont' let us leading to an error in the console which is picked up by error monitoring.

To replicate go to: 
https://www.justonemoreplantco.com/events-and-workshops
Click "Philadelphia Flower Show Bus Trip" to bring up the modal scroll down a bit then try to drag all the way down

https://user-images.githubusercontent.com/16027039/223269826-6500fc2b-e74f-4dd8-a660-e7a5c59e1b56.mov

No UI changes only we should check if an event is cancelable before we actually try to cancel it.

### DIFF
--- a/src/components/TouchCapture/src/TouchCapture.vue
+++ b/src/components/TouchCapture/src/TouchCapture.vue
@@ -111,7 +111,7 @@ export default {
 
 	methods: {
 		handleTouchEvent(event) {
-			if (this.preventDefault) {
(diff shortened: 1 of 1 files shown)


pretzelhammer (approved):

athammer:
seems like the GA failed but isn't related to the code?

PR #533, opened 2023-03-14
docs: adds comment to disabled broken pkg size check for now

pretzelhammer:

it's not working and failing on every pr, let's remove it for now until it's fixed and we can add it back

disables/removes pkg size check for now

## Other information

nope


pretzelhammer:
@privatenumber thanks for the suggestion! i decided to leave the file where it is and disable it from the UI, i added a comment to the file which mentions it's been manually disabled from the UI and we can re-enable it after it is fixed

PR #534, opened 2023-03-14
feat: themable loading

pretzelhammer:

Loading component wasn't themable, would look the same across all sites/themes.

Makes Loading component themable. It now pulls its spinner icon from the `spinner` key in the `icons` object within the theme config object. this allows people to customize the loading spinner via the Theme component.

## Other information

not yet

### DIFF
--- a/src/components/Loading/src/Loading.vue
+++ b/src/components/Loading/src/Loading.vue
@@ -1,34 +1,27 @@
 <template>
 	<div :class="$s.Loading">
-		<svg
+		<m-icon
 			:class="[
 				$s.Spinner,
 				$s[`size_${size}`],
 			]"
-			viewBox="0 0 16 16"
-			xmlns="http://www.w3.org/2000/svg"
-		>
-			<circle
-				:class="[
-					$s.Circle,
-					$s[`size_${size}`],
-				]"
-				fill="none"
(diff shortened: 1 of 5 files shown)


src/components/Loading/src/Loading.vue
```diff
@@ -1,34 +1,27 @@
 <template>
 	<div :class="$s.Loading">
-		<svg
+		<m-icon
 			:class="[
 				$s.Spinner,
 				$s[`size_${size}`],
```
linnea, line 6: Does the size prop work properly with the new MIcon component? Looks like it was just setting height/width before, but you may need to pass it in as size prop where 'large' = 'xxlarge' icon size

pretzelhammer: setting the size via a CSS class bypasses the prop, so i think this is okay, in the future i'd like to make a breaking change where i make the Loading sizes match the Icon sizes but we don't need to do that right now

pretzelhammer: deployed styleguide with changes, large loading size is 64px as expected: https://square.github.io/maker/styleguide/themable-loading/#/Loading

linnea (approved):

PR #537, opened 2023-03-17
fix(icon): fail gracefully if icon not defined in theme

pretzelhammer:

an error being thrown in this edge case is a bit harsh as it can break/crash an app when it's not that big of a deal

emit a warning instead, and fallback to rendering `info` icon instead

## Other information

nope


linnea (approved):
ya

PR #538, opened 2023-03-17
refactor!: standardize sizes between MIcon & MLoading

pretzelhammer:

to quote the updated migration docs in this pr:
> The size prop values for MLoading did not match those of MIcon, which is a problem because MLoading was recently refactored to use MIcon internally. They now match. Also the size prop for either MLoading or MIcon can now also take any arbitrary valid CSS value for width or height. The default sizes for both components remains unchanged.

to quote the updated migration docs in this pr:
> The size prop values for MLoading did not match those of MIcon, which is a problem because MLoading was recently refactored to use MIcon internally. They now match. Also the size prop for either MLoading or MIcon can now also take any arbitrary valid CSS value for width or height. The default sizes for both components remains unchanged.

## Other information

- [rendered migration docs](https://github.com/square/maker/blob/normalize-icon-loading-sizes/.github/MIGRATION.md#17x---18x-pr)
- [icon docs](https://square.github.io/maker/styleguide/normalize-icon-loading-sizes/#/Icon)
- [loading docs](https://square.github.io/maker/styleguide/normalize-icon-loading-sizes/#/Loading)

### DIFF
--- a/src/components/Icon/src/Icon.vue
+++ b/src/components/Icon/src/Icon.vue
@@ -22,7 +22,8 @@ const ICON_SIZES = {
 	xxlarge: '48px',
 };
 
-const widthValidator = cssValidator('width');
+// width & height css props accept same set of values
+const sizeValidator = cssValidator('width');
 
 /**
  * @inheritAttrs svg
@@ -64,7 +65,7 @@ export default {
 		size: {
 			type: String,
 			default: 'small',
-			validator: (size) => ICON_SIZES[size] || widthValidator(size),
+			validator: (size) => ICON_SIZES[size] || sizeValidator(size),
 		},
 		/**
 		 * color of icon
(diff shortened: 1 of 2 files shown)


src/components/Icon/src/Icon.vue
```diff
@@ -15,12 +15,15 @@ import { MThemeKey, defaultTheme, resolveThemeableProps } from '@square/maker/co
...
-	xlarge: '48px',
-	xxlarge: '64px',
+	small: '16px',
+	medium: '24px',
+	large: '32px',
+	xlarge: '40px',
+	xxlarge: '48px',
 };
 
+const widthValidator = cssValidator('width');
```
linnea, line 25: Guessing there's no difference between `cssValidator('width');` and `cssValidator('height');` so this is probably fine?

linnea (approved):

src/components/Icon/src/Icon.vue
pretzelhammer: i thought of doing this at first but after checking MDN it seems like `width` and `height` properties accept the same set of values. i'll add a code comment so it's more clear.

PR #541, opened 2023-04-06
docs: provide unique keys for side nav

kevinlee11:
Some components are under multiple groups, which when run locally causes these vue warns to occur
image

Prepend group to the key name to make them unique


pretzelhammer (approved):
great catch, thanks

PR #549, opened 2023-05-01
fix(popover): fix destroy logic

kevinlee11:
Doing some memory leak testing and looks like popover poppers weren't being properly destroyed resulting in detached HTML elements.

Digging in there were 2 core issues:
1) `currentPopper` was being accessed in `close` in the actionApi scope instead of the vm scope, so `.destroy()` was never being called on the popper instance.
2) Even if 1) was working correctly, the `close` logic in the actionApi is only triggered if the user clicks the action element (e.g. a button). However, if the user clicks outside to close the popover, the `close` logic was never called and thus the `destroy` logic never called (as `popperToDestroy` is never set)

I don't think there was any need for the logic in `close` for destroying. We can just rely on `m-transition-fade-in @after-leave="destroyPopper"`, because leaving the transition indicates the popover is being closed, whether it's a click on the action element, or an outside click. So get rid of the `currentPopper` variable and just use `popperToDestroy`, setting it as before and using it to destroy when the popover fades away.

### DIFF
--- a/src/components/Popover/src/Popover.vue
+++ b/src/components/Popover/src/Popover.vue
@@ -123,7 +123,9 @@ export default {
 			popperToDestroy: undefined,
 			actionAPI: {
 				open(...ignoreElements) {
-					if (vm.isOpen) {
+					// Even though popover may be closed, need to wait for
+					// the popper to be destroyed before re-opening
+					if (vm.isOpen || Boolean(vm.popperToDestroy)) {
 						return;
 					}
 
--- a/src/components/Popover/src/Popover.vue
+++ b/src/components/Popover/src/Popover.vue
@@ -16,7 +16,7 @@
 						v-if="isOpen"
 						:action-el="actionEl"
 						:popper-config="popperConfig"
-						@popover-instance:new-popper="setPopper"
+						@popover-instance:new-popper="setPopperToDestroy"
 					>
 						<!-- @slot Content that will appear in the floating popover -->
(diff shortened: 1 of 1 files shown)


pretzelhammer:
hey, this introduces a bug, if you quickly click on the show/hide popover button [in this PR's styleguide deploy](https://square.github.io/maker/styleguide/popover-destroy/#/Popover) it will wrongly render the popover content inline sometimes:

fast-clicking-mpopover

this doesn't happen if you do it on the [latest-stable styleguide](https://square.github.io/maker/styleguide/latest-stable/#/Popover)

kevinlee11:
Ah thanks, good catch!

> hey, this introduces a bug, if you quickly click on the show/hide popover button [in this PR's styleguide deploy](https://square.github.io/maker/styleguide/popover-destroy/#/Popover) it will wrongly render the popover content inline sometimes:
> 
> fast-clicking-mpopover [ fast-clicking-mpopover ](https://user-images.githubusercontent.com/7769424/235947892-f8055f62-c112-4990-9a76-5b6607f52b7d.gif) [ ](https://user-images.githubusercontent.com/7769424/235947892-f8055f62-c112-4990-9a76-5b6607f52b7d.gif)
> 
> this doesn't happen if you do it on the [latest-stable styleguide](https://square.github.io/maker/styleguide/latest-stable/#/Popover)

laurenhu (approved):

PR #561, opened 2023-08-17
feat(text): balance letter-spacing when centered

privatenumber:

When the text is centered with letter spacing, it isn't center aligned:

Screenshot 2023-08-18 at 12 34 13 AM

Add padding left to balance it out:

Screenshot 2023-08-18 at 12 34 02 AM

### DIFF
--- a/src/components/Text/src/Text.vue
+++ b/src/components/Text/src/Text.vue
@@ -121,6 +121,12 @@ export default {
 		},
 	},
 
+	data() {
+		return {
+			isCentered: false,
+		};
+	},
+
 	computed: {
 		...resolveThemeableProps('text', [
 			'pattern',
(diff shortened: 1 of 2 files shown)


src/components/Text/src/Text.vue
```diff
@@ -198,10 +204,36 @@ export default {
...
+		/**
+		 * Letter spacing is applied to the right-side of the letter
+		 * so when the text is centered, it becomes misaligned
+		 *
+		 * Detect if the text is center aligned and add left padding
+		 * to balance out the letter spacing
+		 */
+		detectAlignCenter() {
+			const computedStyle = window.getComputedStyle(this.$el);
+			const textAlign = computedStyle.getPropertyValue('text-align');
```
pretzelhammer, line 232: the Text component already has a `textAlign` prop, can we just use that here instead of the `getComputedStyle` call?

privatenumber: No, because it's not always passed in. For example, the container (or the section layout) could determine how the text is aligned.

pretzelhammer: can we fix this in the website codebase (always make sure textAlign is passed when set when using MText) or would that be too time-consuming?

privatenumber: I guess it's possible via theming, but since it would touch every section layout I think it would be a pretty big task.

I'm thinking even if we do that refactor, we'd want this in in the interim as it would catch `text-area: center` in case we miss any.

1. Would it be worth emitting a warning if centering is detected but not passed in via `textAlign`?
2. Does Maker have a logging system that's only enabled within certain environments (dev mode)?

pretzelhammer: i think 1 would be worth it. i don't believe we have 2. if 2 is necessary for 1 and 2 would be too time-consuming to implement then i guess we can skip 1.

yeah, i think this is fine for the interm so i'll approve this pr, but would be nice if we could just rely on the prop and didn't have to detect.

pretzelhammer (approved):

PR #564, opened 2023-10-12
feat(image): add hexagon as an image shape

mbove-square:
Add hexagon as an image `shape` option

Screenshot 2023-10-12 at 12 44 41 PM

* Implemented using css `clip-path` with an inline svg

## Other information

* Initially tried implementing this using a shared svg file, but the `clip-path` css property only works with a specific id on `clipPath` which made referencing the file difficult. [Documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path) suggests that this can be done with `clip-path: url(hexagon.svg#hexPath)`, but I was not able to get this to work in the same way we would reference `background-image: url(./hexagon.svg)`. I could only get this to work if the referenced svg was visible in the DOM without being conditionally rendered. Referencing svg via file path would also have to add additional webpack loaders.
I also considered appending an svg to the DOM if one is not present, but that didn't quite seem best practice. 
Open to suggestions here

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -29,6 +29,28 @@
 		<pseudo-window
 			@resize="throttledResizeHandler"
 		/>
+		<div
+			:class="$s.ImageClipPaths"
+		>
+			<svg
+				viewBox="0 0 456 456"
+				fill="none"
+				xmlns="http://www.w3.org/2000/svg"
+			>
+				<defs>
+					<clipPath
+						id="mImageHexPath"
+						clipPathUnits="objectBoundingBox"
+						:transform="svgScale"
+					>
+						<path
+							:d="SVG_PATH"
+							fill="white"
+						/>
+					</clipPath>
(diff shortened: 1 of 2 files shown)


src/components/Image/src/Image.vue
```diff
@@ -252,5 +284,15 @@ export default {
...
+
+	&.shape_hexagon {
+		-webkit-clip-path: url(#mImageHexPath);
+		clip-path: url(#mImageHexPath);
+	}
+}
+
+.ImageClipPaths {
+	width: 0;
+	height: 0;
```
laurenhu, line 296: Could we add an `overflow: hidden` here just in case the svg inherits anything that would impact it's dimensions or visibility from context?

laurenhu (approved):
Minor comment otherwise lgtm! 👍

src/components/Image/src/Image.vue
mbove-square: Just added

PR #569, opened 2023-11-27
fix(modal,dialog): prevent unintended close for modals/dialogs when selecting text

tatems:
When clicking and dragging to select text, if the user releases their cursor outside the modal/dialog window, the modal/dialog will close (specifically on Chrome). This is because the `click` event uses the hovered target on `mouseup`, which trips the `closeOnClickOutside` logic. On other browsers like Firefox, the target is the element hovered on `mousedown`.

- Update `@click.capture` to `@mousedown.capture` for the modal/dialog layer components to trigger the `closeOnClickOutside` logic.

### DIFF
--- a/src/components/Dialog/README.md
+++ b/src/components/Dialog/README.md
@@ -35,7 +35,7 @@ export default {
 
 	methods: {
 		openDialog() {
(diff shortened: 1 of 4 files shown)


tatems:
Still need to test on mobile

PR #570, opened 2023-11-30
refactor: adjust sizing of MButton info slot

berwyn:

The `information` slot in `MButton` currently uses a `width` property inside a flex container, which causes the text to be cut off in situations where the main slot runs long.

image

This PR changes the slot to be sized with `flex` properties, specifically setting the `flex-basis` so that contents aren't elided.

image

### DIFF
--- a/src/components/Button/src/Button.vue
+++ b/src/components/Button/src/Button.vue
@@ -681,7 +681,7 @@ export default {
 }
 
 .InformationText {
-	width: min-content;
+	flex: 0 0 fit-content;
 	padding-left: var(--letter-spacing);
 	opacity: 0.6;
 }


ashjtan:
jw, what is the behavior when the main label (left side) is really really long? does the font size accommodate, so that the text continues to wrap? does it get cut off after two lines?

i believe the `#info` slot (right side) is generally "secondary" to the main label (left side). if we are going to ensure that it always shows via `fit-content`, we should ensure that the main label always shows as well.

src/components/Button/src/Button.vue
```diff
@@ -681,7 +681,7 @@ export default {
 }
 
 .InformationText {
-	width: min-content;
+	flex: 0 0 fit-content;
```
danlasker, line 684: Any reason for using `fit-contnent` over `auto`. Auto is the default and I think slightly better supported.

ashjtan:
> @berwyn Both slots are [supposed to ellipsize](https://github.com/square/maker/blob/master/src/components/Button/src/Button.vue#L701-L703C26) but I'm not sure the logic is correctly applying

Do you mind verifying this behavior? I think it's even more important that this works correctly if we're going to force the info slot to always `fit-content`. It could force a lot of existing MButtons to overflow that weren't before.

pretzelhammer:
note: if your pr title begins with "refactor" i believe merging it won't produce a maker release, if you'd like these changes to be released immediately after merging consider changing the pr title to begin with "fix" if it fixes an issue

PR #571, opened 2023-12-04
fix(skeleton): use opacity instead of background-color

kevinlee11:
Lighthouse analysis led to recommend using opacity for the skeleton loaders for better performance (https://web.dev/articles/stick-to-compositor-only-properties-and-manage-layer-count).

Anecdotally, testing locally, I did notice an improvement in the animation with less jag using opacity when simulating on our larger sites with a mid-tier mobile device.

Switch to using opacity for the skeleton loaders.

https://github.com/square/maker/assets/18740390/7d2048a7-2456-420a-93dc-08ad308fd8e8

https://github.com/square/maker/assets/18740390/a6178240-7316-4b97-828a-3afd6ade53e9

### DIFF
--- a/src/components/Skeleton/src/SkeletonBlock.vue
+++ b/src/components/Skeleton/src/SkeletonBlock.vue
@@ -42,6 +42,7 @@ export default {
 .SkeletonBlock {
 	width: 100%;
 	height: 100%;
+	background-color: $maker-color-neutral-20;
 
 	&.loading {
 		animation: pulsing 0.5s ease-in-out infinite alternate;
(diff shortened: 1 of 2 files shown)


privatenumber:
Great article! I'm fine with this change. Thanks for providing the videos for demonstration.

However, we should be aware that there might be unexpected behavior due to the element animating transparency. For instance, if it's on a blue background, the color of the skeleton could differ because the blue background might be visible through the skeleton.

pretzelhammer (approved):
ditto what Hiroki said. I'm fine with this, it may subtly change the appearance of the skeleton on some sites but it should be pretty close to what it was like before since the maker neutral colors are derived from the site background color anyway.

kevinlee11:
Yup, good callout! It is pretty close, here's a vid with a red card background

https://github.com/square/maker/assets/18740390/ec832a4a-8ae5-4926-bd8d-11803257449f

laurenhu (approved):

Ryguy11 (approved):

danlasker (approved):

PR #572, opened 2023-12-08
perf(image): scrolling large number of images in Safari

kevinlee11:
Safari seems to really struggle with getImageDimensions, specifically calling offsetHeight/offsetWidth which triggers a style recalculation. If you're scrolling a list that has images (e.g. OO on mobile view), it just causes the main thread to get frozen.

Screenshot 2023-12-08 at 3 14 12 AM
Screenshot 2023-12-08 at 3 14 45 AM

Try to preload the image dimensions in the mounted event instead of waiting of the onLoaded event which can cause the main thread to freeze for a significant amount of time.

Also added logic to just skip getting the image dimensions if the shape is square.

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -177,11 +177,23 @@ export default {
 		isThumbnail() {
 			return this.width < THUMBNAIL_MAX_WIDTH;
 		},
+
+		shouldGetImageDimensions() {
+			return this.shape !== 'square';
+		},
 	},
 
 	watch: {
 		src: 'load',
 		srcset: 'load',
+		shape: {
+			immediate: true,
+			handler() {
+				if (this.shouldGetImageDimensions && (!this.height || !this.width)) {
+					this.$nextTick(() => this.getImageDimensions());
+				}
+			},
+		},
 	},
 
 	mounted() {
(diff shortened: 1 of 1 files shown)


pretzelhammer:
it kinda looks like you've created an ad-hoc implementation of lodash's throttle/debounce, is it possible to just use those functions directly instead?

kevinlee11:
> it kinda looks like you've created an ad-hoc implementation of lodash's throttle/debounce, is it possible to just use those functions directly instead?

I think the issue with throttle/debounce is it would only trigger once, and I'm not as confident to use an arbitrary number value on mounted and hope after the throttle/debounce that the offesetHeight/offsetWidth is available. Because those aren't conditional that I can see?

It would just be the equivalent of doing
```
		const getImageDimensionsFn = () => {
			setTimeout(() => {
			    if (!this.height || !this.width) {
				this.getImageDimensionsTimeout = setTimeout(getImageDimensionsFn, timeoutValue);
			    }
			, 1000}
		};
		if (this.shouldGetImageDimensions) {
			this.$nextTick(getImageDimensionsFn);
		}
```

kevinlee11:
For the short term I'd also be fine just forging ahead with the `shouldGetImageDimensions` check, it'd be enough to test out the performance relief on several sites as they all use the square shape.

pretzelhammer (approved):

PR #573, opened 2023-12-11
perf(image): disable transition for safari + unwatch observer after it triggers

kevinlee11:
Trying to find additional improvements for Safari perf.

 Removing the transition as a prop, plus unwatch the observer after it triggers.


src/utils/devices.js
```diff
@@ -0,0 +1,17 @@
...
+function isBrowser() {
+	return (typeof window !== 'undefined');
+}
+
+/**
+ * Is it apple safari browser
+ * @return {boolean}
+ */
+export function isSafari() {
+	return isBrowser() && window.navigator.userAgent.includes('Safari')
```
danlasker, line 15: Should we toLowerCase these and then do `safari`?

danlasker (approved):

src/components/Image/src/Image.vue
```diff
@@ -179,7 +186,14 @@ export default {
...
 		shouldGetImageDimensions() {
-			return this.shape !== 'square';
+			return this.shape !== 'square' || this.shape !== 'original';
+		},
+
+		transitionComponent() {
+			if (!this.disableTransition) {
+				return 'm-transition-fade-in';
+			}
+			return 'span';
```
mattmewton, line 196: Might want to make this a `<div>` since `` is the direct child to this component.

src/components/Image/src/Image.vue
```diff
@@ -129,6 +132,10 @@ export default {
 			validator: cssValidator('object-position'),
 			default: 'center',
 		},
+		disableTransition: {
```
mattmewton, line 135: Can we call this `shouldDisableTransition`?

mattmewton (approved):

src/components/Image/src/Image.vue
kevinlee11: I tried div and it messed up the image shape
image
image

src/utils/devices.js
kevinlee11: Ended up making it a prop so we can control the logic in website and keep Maker browser agnostic

pretzelhammer (approved):

PR #575, opened 2023-12-12
perf(image): update css for perf on safari

kevinlee11:
Noticed on our OO sites in Safari removing the images vastly improved the scroll experience. From there it was just removing attributes and CSS until I narrowed it down to the `height: 100%` seemingly causing the issue. Will see if the deployed version provides the same perf gain.
 
Use `inherit` instead of `100%` and remove the parent `span` as that causes perf issues. `img` needs to be a direct child of the `div`.

### DIFF
--- a/src/components/Image/src/Image.vue
+++ b/src/components/Image/src/Image.vue
@@ -10,8 +10,27 @@
 				$s[`shape_${resolvedShape}`],
 			]"
 		/>
-		<component
-			:is="transitionComponent"
+		<!-- Need to use a template. Using an actual element (e.g span)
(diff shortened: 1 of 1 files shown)


rgershon:
Have you tested in other browsers to make sure these changes don't cause any issues?

kevinlee11:
> Have you tested in other browsers to make sure these changes don't cause any issues?

Yup, Chrome, Safari, and FF.

wkashdan (approved):

pretzelhammer (approved):

PR #576, opened 2023-12-13
perf(image): use static size styles for safari

kevinlee11:
Still trying to narrow down a fix for the Safari scroll issues.

Using overrides in Safari, setting a static height/width style on the images after they render seems to solve the perf issue. Tested on a photo layout site to ensure the images resize properly switching between desktop/mobile. Will only enable this prop for Safari.


mattmewton (approved):
This looks promising! 🤞

PR #578, opened 2024-03-26
fix(text): prevent always running detect text alignment

wkashdan:

This pull request is meant to improve browser performance of the Text component. The current implementation causes the browser to recompute actual styles each time a text component is mounted and/or updated. This occurs only to fix an issue with letter spacing.

This PR ensure the logic to detect center styling only runs if custom letter spacing is passed in to the component, otherwise there is no need to run this. 

## Other information
This issue frequently causes Safari to do a force layout calculation which is inefficient.

### DIFF
--- a/src/components/Text/src/Text.vue
+++ b/src/components/Text/src/Text.vue
@@ -123,7 +123,7 @@ export default {
 
 	data() {
 		return {
-			isCentered: false,
+			isCenteredAndSpaced: false,
 		};
 	},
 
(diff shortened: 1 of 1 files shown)


src/components/Text/src/Text.vue
```diff
@@ -227,10 +227,14 @@ export default {
 		 * Detect if the text is center aligned and add left padding
 		 * to balance out the letter spacing
 		 */
-		detectAlignCenter() {
+		detectAlignCenterAndLetterSpacing() {
+			if (!this.letterSpacing && !this.resolvedLetterSpacing) {
```
pretzelhammer, line 231: i don't think you need the `!this.letterSpacing` here, I think just `!this.resolvedLetterSpacing` is enough (since it checks `this.letterSpacing` under the hood).

pretzelhammer (approved):
looks fine, just left minor comment

wkashdan:
Closing this in favor of https://github.com/square/maker/pull/579

PR #580, opened 2024-03-26
fix(text): removes always computing text element alignment

wkashdan:
The text element calls `getComputedStyle` which is very inefficient sometimes causing many milliseconds of lag and layout recalculations. 

This PR removes calculating the computed text alignment. This means the padding fix will only be set when text alignment is directly set on the element and has letter spacing (whether inherited or not).

### DIFF
--- a/src/components/Text/src/Text.vue
+++ b/src/components/Text/src/Text.vue
@@ -232,9 +232,8 @@ export default {
 				return;
 			}
 
-			const computedStyle = window.getComputedStyle(this.$el);
-			const textAlign = computedStyle.getPropertyValue('text-align');
-			this.isCenteredAndSpaced = textAlign === 'center';
+			// this does not account for inheritted centered text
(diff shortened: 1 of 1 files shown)


privatenumber:
Thanks @wkashdan 
Do you mind adding the screenshot and the context you provided me over DM?
Specifically, I think you mentioned this happens on scroll, which is unexpected.

---

The fix you're removing (https://github.com/square/maker/pull/561) corrects text alignment in the Editor and Published sites so I'm not sure if we can remove it.

I think the fix is actually fine because it only runs on `mounted` and `updated`. I think this begs the question: why is the Vue subtree getting "updated" on scroll? I think even if we remove the `getComputedStyle` here, unless we prevent the "updates" happening on scroll, we'll always have unexpected perf issues.

wkashdan:
Closing, as @privatenumber recommended not going with this approach

PR #584, opened 2025-06-18
feat(a11y): Updating Maker components to be accessible 

athammer:

No functional UI changes, just updating docs and allowing consumers to add accessibility props to the components.

### DIFF
--- a/src/components/Theme/README.md
+++ b/src/components/Theme/README.md
@@ -412,7 +412,7 @@ export default {
 					primary: '#000000',
 					'neutral-0': '#ffffff',
 					'neutral-10': '#f1f1f1',
-					'neutral-20': '#d3d3d3',
+					'neutral-20': '#949494',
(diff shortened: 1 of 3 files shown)


laurenhu (approved):

.github/workflows/lint-pr-title.yml
```diff
@@ -1,13 +1,18 @@
...
+  pull-requests: read
 
 jobs:
   lint-pr-title:
     name: Lint PR title
     runs-on: ubuntu-latest
     steps:
       - name: Lint PR title
-        uses: amannn/action-semantic-pull-request@v2.1.0
+        uses: amannn/action-semantic-pull-request@v5.5.3
```
athammer, line 16: read through the breaking changes for version updates. nothing i saw that we'd care about. just upgrading node versions or changing flags around which we take care of

